PR #55, opened 2023-03-31
Feature: Add destroy services and Update services

deepsidhu85:
Added destroy services for:
- Groups
- Project
- Samples
- Members

Added members update service
Added permissions to services
Added tests to cover these services

**Note: Looks like a really big PR but mostly testing**


app/controllers/projects/samples_controller.rb
```diff
@@ -29,13 +29,12 @@ def edit; end
...
-          flash[:success] = t('.success')
-          format.html { redirect_to namespace_project_sample_path(id: @sample.id) }
-        else
-          format.html { render :new, status: :unprocessable_entity }
-        end
+      if @sample.persisted?
+        flash[:success] = t('.success')
+        redirect_to namespace_project_sample_path(id: @sample.id)
+      else
+        flash[:error] = @sample.errors.full_messages.first
```
ericenns, line 36: This has the same behaviour as the comment above, so please remove the flash here.

app/controllers/projects_controller.rb
```diff
@@ -32,7 +32,7 @@ def create
         project_path(@project)
       )
     else
-
+      flash[:error] = @project.errors.full_messages.first
```
ericenns, line 35: This has the same behaviour as the comment above, so please remove the flash here.

app/models/member.rb
```diff
@@ -45,6 +47,15 @@ def validate_namespace
     errors.add(namespace.type, 'namespace cannot have members')
   end
 
+  # Method to ensure we don't leave a group or project without an owner
+  def last_namespace_owner_member
```
ericenns, line 51: This only needs to apply to GroupMember, and it only has to be check last owner in the hierarchy. If you want this could be addressed in a separate PR.

app/services/groups/create_service.rb
```diff
@@ -3,20 +3,31 @@
...
+    def execute # rubocop:disable Metrics/AbcSize
+      unless allowed_to_modify_group?(group)
+        raise GroupCreateError, I18n.t('services.groups.create.no_permission',
+                                       namespace_type: group.type.downcase)
+      end
+
+      group.save
 
-      Members::CreateService.new(current_user, @group, {
+      Members::CreateService.new(current_user, group, {
```
ericenns, line 22: We only need to create a Membership if the `group.parent` is nil, as they would already have to be an Owner in the hierarchy if they could create a subgroup. This could also be done in a separate PR.

app/controllers/concerns/membership_actions.rb
```diff
@@ -33,22 +33,19 @@ def create
       flash[:success] = t('.success')
       redirect_to members_path
     else
-      flash[:error] = t('.error')
+      flash[:error] = @new_member.errors.full_messages.first
```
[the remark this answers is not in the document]
deepsidhu85: Sure thing! Thanks for the explanation and link! Updated in 2365691109fea5b67fd2bc7040892d5ad22edff2

app/controllers/projects/samples_controller.rb
deepsidhu85: Updated in 2365691109fea5b67fd2bc7040892d5ad22edff2

app/controllers/projects_controller.rb
deepsidhu85: Updated in 2365691109fea5b67fd2bc7040892d5ad22edff2

app/services/members/destroy_service.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def initialize(member, namespace, user = nil, params = {})
+      super(user, params)
+      @member = member
+      @namespace = namespace
+    end
+
+    def execute # rubocop:disable Metrics/AbcSize
+      unless current_user != member.user
+        raise MemberDestroyError, I18n.t('services.members.destroy.cannot_remove_self',
```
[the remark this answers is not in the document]
deepsidhu85: Updated in 74af98fa1df229298ecbe2832509b919e9f85bdc

app/services/groups/create_service.rb
deepsidhu85: Updated in 81beea8257c5b723281c071ae46caeb8cbe85653

app/services/projects/create_service.rb
```diff
@@ -3,26 +3,38 @@
...
+    rescue Projects::CreateService::ProjectCreateError => e
+      project.errors.add(:base, e.message)
+      project
+    end
 
-      Members::CreateService.new(current_user, @project.namespace, {
+    def create_associations(project)
+      project.build_namespace(namespace_params.merge(owner: current_user))
+      project.save
+      Members::CreateService.new(current_user, project.namespace, {
```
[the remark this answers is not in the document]
deepsidhu85: Updated in 81beea8257c5b723281c071ae46caeb8cbe85653

app/models/members/group_member.rb
```diff
@@ -4,9 +4,20 @@ module Members
...
     belongs_to :group, foreign_key: :namespace_id # rubocop:disable Rails/InverseOf
+    before_destroy :last_namespace_owner_member
 
     def self.sti_name
       'GroupMember'
     end
+
+    # Method to ensure we don't leave a group or project without an owner
+    def last_namespace_owner_member
+      return unless !destroyed_by_association && (namespace.owners.count == 1 && namespace.owners.last == user)
```
ericenns, line 15: Instead of checking `namespace.owners.count` you want to get all the `Owners` in the Group hierarchy, since if there is at least a `Owner` in the `self_and_ancestors` this one can be safely removed, as  long as it is not the only one.

config/locales/en.yml
```diff
@@ -36,11 +36,15 @@ en:
         name: Group name
         path: Group slug
         parent_id: Group URL
+        one: Group
```
ericenns, line 39: These need to be under `models` and not `attributes`, please re-refer to comment above.

app/services/projects/create_service.rb
```diff
@@ -3,26 +3,41 @@
...
+      project
+    end
 
-      @project.save
+    def create_associations(project)
+      project.build_namespace(namespace_params.merge(owner: current_user))
+      project.save
 
-      Members::CreateService.new(current_user, @project.namespace, {
+      return unless user_has_namespace_maintainer_access?
```
ericenns, line 35: I think this might result in the creation of a Membership when the following is true:

User creates a Project in a Subgroup, where they have Maintainer role in the parent group, but Owner role in the Subgroup. So really you want to check that they don't have Owner role and they do have Maintainer role.

app/services/members/destroy_service.rb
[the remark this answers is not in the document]
ericenns: The translations look to have been added in the wrong location, I left a comment on the en.yml file

ericenns (changes requested):
Looking better, still a couple issues to resolve

app/models/members/group_member.rb
deepsidhu85: Whoops thanks for catching that. Updated in 20e119965680ad3aa0f927bffc00468e33c0fff7

config/locales/en.yml
deepsidhu85: Updated in 20e119965680ad3aa0f927bffc00468e33c0fff7

app/services/projects/create_service.rb
deepsidhu85: Yeah looks like it would do that. Updated in 20e119965680ad3aa0f927bffc00468e33c0fff7

app/models/members/group_member.rb
```diff
@@ -4,9 +4,22 @@ module Members
...
   class GroupMember < Member
     belongs_to :group, foreign_key: :namespace_id # rubocop:disable Rails/InverseOf
+    before_destroy :last_namespace_owner_member
 
     def self.sti_name
       'GroupMember'
     end
+
+    # Method to ensure we don't leave a group or project without an owner
+    def last_namespace_owner_member
```
ericenns, line 14: This still needs a bit more work.

I think you should first check that the member being removed is an owner and then you can use the statement where you check for many

ericenns (changes requested):
Found a couple more issues

app/controllers/concerns/membership_actions.rb
```diff
@@ -33,28 +33,24 @@ def create
...
+      flash[:error] = @member.errors.full_messages.first
     end
+    redirect_to members_path
   end
 
   private
 
   def access_levels
-    member = Member.find_by(user: current_user, namespace: @namespace, type: @member_type)
+    member = Member.find_by(user: current_user, namespace: @namespace.self_and_ancestors)
```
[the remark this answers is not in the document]
deepsidhu85: Updated to get the highest member role in a11c9c768378fa5e50e557466b32a1b73e43342f

app/models/members/group_member.rb
deepsidhu85: Updated to return if destroyed by association, if the member isn't an owner, and only run the query to check for many if the first 2 statements are not true in a11c9c768378fa5e50e557466b32a1b73e43342f

ericenns (approved):
Looks good to me!

PR #61, opened 2023-04-05
Viral components custom library

joshsadam:
Custom UI Components strongly based of off [Polaris Components](https://github.com/baoagency/polaris_view_components).  Currently moved over:
1. Breadcrumbs
2. Flash alerts
3. Icons
4. Page header
5. Dropdown

Fully integrated with LookBook previews.


app/components/viral/base_component.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Viral
+  # Base component to be inherited from
+  class BaseComponent < Viral::Component
+    def initialize(tag:, classes: nil, **system_arguments)
```
deepsidhu85, line 6: This comment applies to all the components in which you access the variable from the initialize method but will just leave the comment in this one place as to not clutter up the PR

When needing access to an instance variable we should be using `attr_accessor` to provide access to it from the rest of the class.

So here above the initialize method you would have:

`attr_accessor :tag, :content_tag_args`

Then when needing access to them in the class you would just use `tag` and `content_tag_args`. When updating this, the initialize methods do not need to be changed at all, and if the class doesn't have anything other than an initialize method then we don't need to add the attr_accessor

There's a few reasons for doing it this way. First it helps with some self documenting of the code, provides better error messages if a typo happens in the class with the variable name, and last but not least just for consistency across the application. We've been using attr_accessor in our services rather than accessing the instance variable directly from the initialize method

deepsidhu85 (changes requested):
This is looking good just have a couple of comments below

app/components/viral/dropdown_component.rb
```diff
@@ -0,0 +1,48 @@
...
+      @system_arguments[:tag] = :button
+
+      @system_arguments = system_arguments_for_button if @label.present?
+
+      return if @icon_name.blank?
+
+      @system_arguments = system_arguments_for_icon
+    end
+
+    def system_arguments_for_button
```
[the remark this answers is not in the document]
joshsadam: Cleaned up in 9ea555d

app/components/viral/base_component.rb
joshsadam: Cleaned up in 9ea555d

deepsidhu85 (approved):
This looks good to me. Thanks for making the changes!

deepsidhu85 (changes requested):
Sorry just missed one thing. Left a comment below

deepsidhu85 (approved):

ericenns (approved):
Looks good to me!

PR #63, opened 2023-04-10
Enhancement: Added paging to the projects list page

joshsadam:
- Used `pagy` for `next` and `prev` pagination on project list 
  - Left pagy defaults for page size (20).
  - No sorting yet, set to modified date descending
- Created PaginationComponent for consistent pagination 
- Add "time ago" to project list


app/views/projects/index.html.erb
```diff
@@ -12,25 +12,35 @@
...
-          <% @projects.each do |project| %>
-          <tr
-            class="text-sm font-normal text-gray-500  hover:bg-gray-50 dark:hover:bg-gray-700 dark:text-gray-400"
-          >
-            <td class="p-4 whitespace-nowrap">
-              <%= link_to project.human_name,
-                project_path(project),
-                class: "text-grey-900 dark:text-grey-100 font-semibold hover:underline" %>
-                <div class="text-sm font-normal text-gray-500 dark:text-gray-400"><%= project.description %></div>
+          <% 15.times do %>
```
deepsidhu85, line 26: What is going on here exactly? Is this to display something like the skeleton loading in ant design?

app/views/projects/edit.html.erb
```diff
@@ -1,22 +1,43 @@
...
+          font-semibold
+          leading-tight
+          text-gray-900
+          dark:text-gray-100
+        "
+      >
+        <%= t(:".general.title") %>
       </h2>
       <%= form_with(model: @project, url: namespace_project_path, method: :patch) do |form| %>
-        <%= turbo_frame_tag "project_form" do %>
```
deepsidhu85, line 9: Do we no longer require the `turbo_frame_tag` here?

deepsidhu85 (changes requested):
This looks good just have a couple of questions below

app/views/projects/edit.html.erb
joshsadam: Yeah, this form never actually used a turbo frame, I just was copying code.  We need to break this one up anyways because of the namespace stuff so this will change again shortly.

app/views/projects/index.html.erb
deepsidhu85: Sounds good!

deepsidhu85 (approved):
Looks good to me!

ericenns (approved):
Looks good to me!

PR #65, opened 2023-04-13
WIP: Consolidated member, group member, and project member models

deepsidhu85:
Consolidated member, group member, and project member into a single member model. Added migration to remove STI type column from member table and updated tests


ericenns (approved):
Looks good to me!

PR #66, opened 2023-04-14
Feature: Linear ancestor traversal

ericenns:
Replace recursive traversal of ancestors with linear traversal using route paths.

Fix bug in groups create service, that was passing the group to be created to permission evaluator instead of the parent.


deepsidhu85 (approved):
This looks good to me!

PR #69, opened 2023-04-18
Internal: Github Pages Docs Site

ericenns:
Setup auto deployment of docs-site to Github pages on push to main.


.github/workflows/docs-site.yml
```diff
@@ -0,0 +1,59 @@
+name: Docs Site Deploy
+
+on:
+  push:
+    branches: [ main ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
deepsidhu85, line 8: Is anything else required under this section or that's all thats required if we want to run the workflow manually?

deepsidhu85 (changes requested):
Just one comment below otherwise looks good

.github/workflows/docs-site.yml
ericenns: Just having this section present, should allow manual invocation

PR #81, opened 2023-04-26
Update: Migrate old dropdowns to Viral::DropdownComponent

ericenns:
Groups index and Project Samples index, where not migrated over to Viral::DropdownComponent. This PR fixes that.

### DIFF
--- a/app/views/groups/index.html.erb
+++ b/app/views/groups/index.html.erb
@@ -21,16 +21,7 @@
             "
           >
             <% @groups.each do |group| %>
-              <tr
-                class="
-                  text-sm
-                  font-normal
-                  text-gray-500
-                  hover:bg-gray-50
-                  dark:hover:bg-gray-700
-                  dark:text-gray-400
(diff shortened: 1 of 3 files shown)


app/views/groups/index.html.erb
```diff
@@ -20,22 +20,22 @@
...
-            <tr
-              class="
+            <% @groups.each do |group| %>
+              <tr
+                class="
                   text-sm
                   font-normal
                   text-gray-500
                   hover:bg-gray-50
                   dark:hover:bg-gray-700
```
joshsadam, line 30: Can you get rid of these hovers here so that it matches how you wanted the projects table

ericenns: Addressed in 313ea06

joshsadam: Thanks :)

joshsadam (approved):
Works as advertised.  Thanks for fixing this.

PR #83, opened 2023-04-27
Internal: Namespace Traversal Updates

ericenns:
Updated Namespace instance lineage methods (e.g. `ancestors`, `descendants`, etc) to only return instances of the current class (e.g. `Group` when calling on a `Group` instance, etc).

Added in Namespace class lineage methods, so that we can get all the Namespaces of a specific type that a User has access to.

Updated Groups Index and Projects Index to only show the corresponding models that the current_user has access to through `membership`.


deepsidhu85 (approved):
Looks good to me! Thanks for getting this updated!

PR #85, opened 2023-04-28
Update: Viral Card Component

joshsadam:
Create a viral card component so that all cards are uniform.  You can see all the variants in LookBook:
- Card with header actions
- Card with header
- Card with title and basic test
- Card with multiple sections
- Card with section titles
- Card with section actions


app/components/viral/card/header_component.rb
```diff
@@ -0,0 +1,25 @@
...
+      def initialize(title: '', **system_arguments)
+        @title = title
+        @system_arguments = system_arguments
+        @system_arguments[:classes] = class_names(
+          @system_arguments[:classes],
+          'p-4'
+        )
+      end
+
+      def simple?
```
deepsidhu85, line 20: Let's rename this to something a bit more clear of what it's for

deepsidhu85 (changes requested):
This looks great just have one comment below.

app/components/viral/card/header_component.rb
joshsadam: Updated in 4eeca1a97e13f85aa9e1ce703fbb47dd4a348197

deepsidhu85: Thanks!

deepsidhu85 (dismissed):

app/views/projects/edit.html.erb
```diff
@@ -2,28 +2,11 @@
...
-        border
-        border-gray-200
-        rounded-lg
-        shadow-sm
-        2xl:col-span-2
-        dark:border-gray-700
-        sm:p-6
-        dark:bg-gray-800
-      "
+      class="p-4 bg-white border border-gray-200 rounded-lg shadow-sm  2xl:col-span-2 dark:border-gray-700 sm:p-6 dark:bg-gray-800"
```
ericenns, line 5: Can you undo these changes, it looks like you edited the file outside of vscode and our formatter so it changed the formatting. If I save in VSCode it puts it back to the before state.

test/components/previews/card_component_preview.rb
```diff
@@ -0,0 +1,58 @@
...
+
+class CardComponentPreview < ViewComponent::Preview
+  def default
+    render Viral::CardComponent.new(title: 'Simple card with Title') do
+      'This is a card'
+    end
+  end
+
+  def card_with_header_actions
+    render Viral::CardComponent.new(title: 'Card with header actions') do |card|
```
ericenns, line 11: For some reason this preview uses a different font size for the header text, can you see why this is happening.

ericenns (changes requested):
Looking good, one issue with the "Card With Header Actions" example where it uses a different font size.

test/components/previews/card_component_preview.rb
joshsadam: Thanks, updated in bd31499

app/views/projects/edit.html.erb
joshsadam: Ok fixed

ericenns (approved):
Looks good to me!

PR #87, opened 2023-05-02
Bugfix: Access_level nil error when adding a project member under user namespace

deepsidhu85:
Fixed bug:
- which prevented project owner under their own user_namespace from adding a project member due to a nil error.
- added test to ensure the owner could add a project member

Fixed translation key


ericenns (approved):
Looks good to me!

PR #89, opened 2023-05-03
Internal: Fix Migrations

ericenns:
Previously we we're adding foreign keys with `t.integer blah_id` instead of `t.references :blah`. Using the `references` syntax results in the type of `blah_id` to be `:bigint` which matches the type of the id column in the database.

Note: This change will require you to run the following command:

`bin/rails db:drop db:create db:migrate db:seed`

### DIFF
--- a/test/controllers/groups_controller_test.rb
+++ b/test/controllers/groups_controller_test.rb
@@ -76,7 +76,7 @@ class GroupsControllerTest < ActionDispatch::IntegrationTest
   test 'should delete a group' do
     sign_in users(:john_doe)
 
-    group = groups(:group_one)
+    group = groups(:group_two)
     assert_difference('Group.count', -1) do
       delete group_path(group)
     end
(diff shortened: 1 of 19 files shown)


test/controllers/groups_controller_test.rb
```diff
@@ -76,7 +76,7 @@ class GroupsControllerTest < ActionDispatch::IntegrationTest
   test 'should delete a group' do
     sign_in users(:john_doe)
 
-    group = groups(:group_one)
+    group = groups(:group_two)
```
deepsidhu85, line 79: Can you also add a test here and the test files below to use `group_one` as well to ensure it deletes the group and all it's subgroups

deepsidhu85 (changes requested):
This looks great! Just one comment on the tests below otherwise code is good to go. Just have to give it a thorough manual test

test/controllers/groups_controller_test.rb
deepsidhu85: Awesome great idea. Thanks!

deepsidhu85 (approved):

PR #91, opened 2023-05-04
Update: Projects Index Action

ericenns:
This PR updates the Projects index action to only load a page of projects when the user has access to Projects, and only when processing the turbo_stream request.

Prior to this PR a page of projects was loaded when processing the html request, but the index.html.erb action was not utilizing the page as we defer loading of the table via turbo_stream request.

This PR also updates the db/seeds.rb to seed a user with email `user9@email.com` who does not have access to any projects. This user can be used to test accessing the Projects Index without any projects to see that it does not attempt to load a page of projects.


joshsadam (approved):
This looks good to me.  I think at some point we should create some sort of empty state, but it looks good for now.

deepsidhu85 (approved):
This looks good to me! Working great

PR #93, opened 2023-05-08
Feature: Alert component for login flash notices

joshsadam:
Created an alert component to specifically handle login notices where the flash component is not appropriate.  Has two states:

Notice:

Alert:

image


ericenns (approved):
Looks good to me!

PR #98, opened 2023-05-13
Feature: Member Access Level Update

deepsidhu85:

Added in functionality to allow a member's access level to be updated from the group and project members listing page using turbo_stream. Using turbo_stream allows us to update just the access level select box for the particular member that is being updated rather than having to do a full page reload. 

A few other UI related things were added in such as an 'It's you' tag next to any member listing which matches the current user, as well as, an updated_at column to the projects and group members tables/ 

**Error when attempting to set a lower access level for user than they have in a parent group:**
image

**Success message when an update is successful:**
image

**It's you tag next to the member which matches the current user and the updated_at column:**
image

To test this out:

1) Create a group and add a few members with different access levels
2) Create a project under the above group and add members with the same roles
3) Try to update a project member's access level to any other role other than owner when they have membership in the group as an owner
4) You should be presented with an error such as in the first image under the screenshots section
5) Try updating other members access levels which don't have a higher access role in a parent group. They should update without error. To ensure that the role is actually being updated you can keep an eye on the server logs in the terminal. You should see a logging statement such as `UPDATE "members" SET "access_level" = $1, "updated_at" = $2 WHERE "members"."id" = $3`. You should also be presented with a success message as in the second image under the screenshot section

For further information on how permissions work in IRIDA Next, view the docs for  [Member Permissions](https://phac-nml.github.io/irida-next/docs/user/organization/members/member-permissions)

**Things to verify:**

1) A maintainer can update other member's access levels up to the `Maintainer` access level. Can sign in using `user6@email.com` which is only a maintainer on groups and projects to test this out.
2) A maintainer or owner cannot update their own access levels
3) A maintainer cannot update a member that has the `owner` access level
4) Setting an invalid value for a select option (using chrome -> inspect element) for one of the access levels and then updating the user to that access level should result in an error
5) A Member's access level in a project which is a part of a group can only be set to a minimum of this inherited access level
6) If a member's access level is the same in both a parent group and project, then if the member's access level in the group is updated to a higher access level, then the member's access level should also be updated to match their access level in the parent group

### DIFF
--- a/app/views/groups/members/_access_level.html.erb
+++ b/app/views/groups/members/_access_level.html.erb
@@ -8,7 +8,7 @@
     access_levels,
     {},
     {
-      id: "member-#{member.id}-access-level",
+      id: "member-#{member.id}-access-level-select",
       value: member.access_level,
       onchange: "this.form.submit();",
       "aria-label": "Access Level"
--- a/app/views/groups/members/_access_level.html.erb
+++ b/app/views/groups/members/_access_level.html.erb
@@ -1,18 +1,17 @@
-<%= form_with(model: member, url: group_member_path(id: member.id), method: :patch,
-                        data: {
-                          remote: true
-                        }
-                      ) do |form| %>
-  <%= form.select(
-    :access_level,
-    access_levels,
-    {},
-    {
-      id: "member-#{member.id}-access-level-select",
-      value: member.access_level,
-      onchange: "this.form.submit();",
-      "aria-label": "Access Level"
-    }
-  ) %>
-  <input type="hidden" name="format" value="turbo_stream"/>
-<% end %>
+<% invalid_access_level = member.errors.include?(:access_level) %>
+<div class="form-field <%= 'invalid' if invalid_access_level %>">
+  <%= form_with(model: member, url: group_member_path(id: member.id), method: :patch) do |form| %>
+    <%= form.select(
+      :access_level,
+      access_levels,
+      {},
+      {
+        id: "member-#{member.id}-access-level-select",
+        value: member.access_level,
+        onchange: "this.form.requestSubmit();",
+        "aria-label": "Access Level"
+      }
+    ) %>
+    <input type="hidden" name="format" value="turbo_stream"/>
+  <% end %>
+</div>
--- a/app/views/groups/members/_access_level.html.erb
+++ b/app/views/groups/members/_access_level.html.erb
@@ -1,0 +1,18 @@
+<%= form_with(model: member, url: group_member_path(id: member.id), method: :patch,
+                        data: {
+                          remote: true
+                        }
+                      ) do |form| %>
+  <%= form.select(
+    :access_level,
+    access_levels,
+    {},
+    {
+      id: "member-#{member.id}-access-level",
+      value: member.access_level,
+      onchange: "this.form.submit();",
+      "aria-label": "Access Level"
+    }
+  ) %>
+  <input type="hidden" name="format" value="turbo_stream"/>
+<% end %>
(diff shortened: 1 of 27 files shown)


ksierks (commented):
I think I may have found an error. I was logged in as a maintainer and I was trying to change another member on the project from analyst to guest. I got a 500 back and a "Content missing" for the drop down. Please see below for the screenshot.

image

ksierks (commented):
This is more of a question. It states that "A Member's access level in a project which is a part of a group can only be set to a minimum of this inherited access level". My question is, if a members access changes in the group level to something less than what was already assigned in the project level, should the project level change with the group level access?

ksierks:
> I think I may have found an error. I was logged in as a maintainer and I was trying to change another member on the project from analyst to guest. I got a 500 back and a "Content missing" for the drop down. Please see below for the screenshot.
> 
> image

Disregard. I pulled the latest changes and I'm not getting this error anymore.

ksierks (dismissed):
This is great!

deepsidhu85:
> This is more of a question. It states that "A Member's access level in a project which is a part of a group can only be set to a minimum of this inherited access level". My question is, if a members access changes in the group level to something less than what was already assigned in the project level, should the project level change with the group level access?

As we discussed if a user is a member of a parent group and a member of a project under this group, they can have a lower access level in the group but can have an equal to the lower level and higher access level in the project so they can have full access to the project but we want to limit their actions within the group

deepsidhu85:
> This is great!

Thanks! :)

app/views/groups/members/index.html.erb
```diff
@@ -24,6 +24,7 @@
               dark:divide-gray-700
             "
           >
+            <%= turbo_frame_tag("member-update-alert") %>
```
joshsadam, line 27: I'm curious as to how this is working properly (and it seems to be looking at the generated html.  You have this set within the table tag, and it is not a table element.  Rails seems to be adding the alert before the table.  Interesting.

app/views/groups/members/_access_level.html.erb
```diff
@@ -0,0 +1,17 @@
+<% invalid_access_level = member.errors.include?(:access_level) %>
+<div class="form-field <%= 'invalid' if invalid_access_level %>">
```
joshsadam, line 2: Same comment I sent you in Teams, it would be great if we could disable access levels that the user cannot be set to, it is better UX then allowing them to attempt to change and show and error.

app/views/groups/members/index.html.erb
deepsidhu85: Oh wow good catch it's supposed to be on the outside of the table. Very interesting indeed

app/views/groups/members/_access_level.html.erb
deepsidhu85: Yup I can take a look at how I'm going to do that. I will probably want to do that in a separate pr

app/views/groups/members/index.html.erb
deepsidhu85: Updated this in 67b1261203e2417028fdc8854c4454b402ece0ef

joshsadam: Thanks

joshsadam (approved):
This works really well.  We need to update the alert component to not have a margin on the bottom since it makes the top of the table move down by 4px, but there is a couple other things in there that need updating so I will take care of that in a separate PR (like I would like to see if closable).  Thanks for all the work you put into this PR.

ksierks (approved):
Tested #6 from "Things to verify" and it works as advertised.

PR #90, opened 2023-05-03
Feature: Transfer samples

ksierks:
Fixes #84.

## What does this PR do and why?
This PR only covers transfers samples to another existing project, not a new project. 
This feature is restricted to owners.

Note: CSS not included.

## Screenshots or screen recordings
image
image

## How to set up and validate locally
1. Navigate to a project samples page that the logged in user is an owner of.
2. Check the samples you wish to transfer.
3. Hit the "Transfer samples" button.
4. A modal will appear with a project drop-down. Select the project you wish to transfer the samples to.
5. Verify the samples transferred to the target project.

## PR acceptance checklist
This checklist encourages us to confirm any changes have been analyzed to reduce risks in quality, performance, reliability, security, and maintainability.

- [x] I have evaluated the PR acceptance checklist for this PR.

PR #106, opened 2023-05-24
Modal Component

ksierks:

This is a simple modal component that was pulled from PR #90. 

image

May be viewed via Lookbook.


joshsadam (commented):
Please add a viral helper in `view_helpers.rb`, doing this enables the use of the modal ueing `viral_modal`

app/components/viral/modal_component.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Viral
+  # A component for displaying a modal.
+  class ModalComponent < Component
+    attr_reader :title
```
joshsadam, line 6: **chore (blocking):** Currently, by looking at the preview, there is only one size modal and it takes up most of the screen.  The one feature I would like you to add is sizes.  Can you implement it the same way that the `dropdown_component.rb` handles `TRIGGER_MAPPINGS`.  You can use the styles from [Flowbite modal sizes](https://flowbite.com/docs/components/modal/#sizes).  Make sure you set a default size.

app/components/viral/modal_component.html.erb
```diff
@@ -0,0 +1,73 @@
...
+                dark:hover:text-white
+              "
+              data-action="click->viral--modal-component#close"
+            >
+              <%= viral_icon(name: "x_mark", classes: "w-5 h-5") %>
+              <span class="sr-only">Close modal</span>
+            </button>
+          </div>
+          <!-- Modal body -->
+          <div id="Viral-Modal-Body" class="p-6 space-y-6">
```
joshsadam, line 66: **issue (blocking):** Text is not optimized for dark mode.  To help you can see the dark mode in LookBook preview, there is a toggle in the menu bar.

joshsadam (changes requested):
A couple more things I picked up on.

app/components/viral/modal_component.rb
ksierks: Please see 8c9b83bf72a3527e4bbb5c2b451f3501f0485d39.

app/components/viral/modal_component.html.erb
ksierks: Please see adf30595bd5fa82d8e5b6830338b0017ead26a97.

app/components/viral/modal_component.html.erb
```diff
@@ -0,0 +1,73 @@
...
+                ml-auto
+                inline-flex
+                items-center
+                dark:hover:bg-gray-600
+                dark:hover:text-white
+              "
+              data-action="click->viral--modal-component#close"
+            >
+              <%= viral_icon(name: "x_mark", classes: "w-5 h-5") %>
+              <span class="sr-only">Close modal</span>
```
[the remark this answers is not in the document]
ksierks: Please see 148f1aa6b41f3038f10c5d327d84eab14e46ce3f.

test/components/previews/modal_component_preview.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class ModalComponentPreview < ViewComponent::Preview
+  def default
+    render Viral::ModalComponent.new(title: 'This is a modal title') do |modal|
+      modal.body do
```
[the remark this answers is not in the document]
ksierks: UPDATE: I've reached out about this request. I purposely did not include a button/link in the component. because I wanted the modal to function like antd. There may be times where we might not need a link/button. For example, announcements within IRIDA.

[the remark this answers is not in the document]
ksierks: Please see latest changes.

app/components/viral/modal_component.html.erb
```diff
@@ -0,0 +1,99 @@
+<div data-controller="viral--modal-component">
+  <!-- Modal toggle -->
+  <button
```
joshsadam, line 3: **issue (blocking):** This current implementation will leave an button on the page with no text.  What could possibly work better, would be to accept two new attribute `trigger_text` and `trigger_tag` with the `trigger_tag` defaulting to `button`.  If the `trigger_text` is `nil` do not render the trigger (e.g. could be when triggering a turbo_frame render.  Use the `Viral::BaseComponent` to render the `trigger_tag` adding any `system_arguments` passed along, don't forget you will need to add the data attributes required to trigger the modal.

app/components/viral/modal_component.html.erb
```diff
@@ -0,0 +1,99 @@
...
+      focus:ring-blue-300
+      font-medium
+      rounded-lg
+      text-sm
+      px-5
+      py-2.5
+      text-center
+      dark:bg-blue-600
+      dark:hover:bg-blue-700
+      dark:focus:ring-blue-800
```
joshsadam, line 23: **nitpick (blocking):** For now just used the default `btn btn-default` classes if you want to make the default button or update all the `*-blue* to be *-primary* so it grabs the primary colour from the configuration file.

app/components/viral/modal_component.html.erb
```diff
@@ -0,0 +1,99 @@
...
+      h-[calc(100%-1rem)]
+      max-h-full
+    "
+  >
+    <div class="relative w-full <%= size %> max-h-full">
+      <!-- Modal content -->
+      <div class="relative bg-white rounded-lg shadow dark:bg-gray-700">
+        <!-- Modal header -->
+        <div
+          id="Viral-Modal-Header"
```
joshsadam, line 53: **nitpick (non-blocking):** Try to avoid setting element id's unless there is no other option.  The only time we have been using the `Viral-` class is if a style was created in `application.tailwind.css` which you can freely do.  What we are trying to keep to is the syntax `Viral-`**Component**`--`**element**. E.g. `Viral-Icon--colorSuccess`.

joshsadam (changes requested):
I need to run to get the kids, but these are some early things I have picked up on.

joshsadam (changes requested):
I need to run to get the kids, but these are some early things I have picked up on.

joshsadam (changes requested):
I need to run to get the kids, but these are some early things I have picked up on.

joshsadam (changes requested):
I need to run to get the kids, but these are some early things I have picked up on.

app/components/viral/modal_component.html.erb
joshsadam: It does leave an empty button on the page.  Please see the attached image.

ksierks: Yes, that is hidden, not shown. As does clipboard.

app/components/viral/modal_component.html.erb
ksierks: Please let me know what you want the ids to be to make ui testing easier. In the last IRIDA project the standard was that it should be started with "t-".

app/components/viral/modal_component.html.erb
joshsadam: If you do not like `system_arguments` then you can do the same thing buy passing another argument, perhaps `trigger_classes` and pass the style that way.

joshsadam: > Yes, that is hidden, not shown. As does clipboard.

The difference is that the clipboard component will throw an error if the label is not set (I just tested it), which is because it needs the button to actually work.  Here we do not need to have a button, so that it can work work turbo\_streams easily.  A button with no text is not "hidden" is is still in the document flow, which means that it could possibly be access via tabbing.

ksierks: I don't feel like it's necessary to accept 2 more attributes, when we know it should be triggered via a button if a label is passed. Maybe we can talk tomorrow.

ericenns: `**system_arguments` is a common pattern employed by both [Primer View Components](https://github.com/primer/view_components/blob/main/app/components/primer/base_component.rb) and [Polaris View Components](https://github.com/baoagency/polaris_view_components/blob/main/app/components/polaris/base_component.rb), which were both used as inspiration for our Viral View Components. It is basically a way of accepting unknown keyed arguments. If you take a look at `app/components/viral/card_component.rb` it accepts `**system_arguments` and merges the default classes with any provided classes. Then if you look at `app/components/viral/card_component.html.erb` it renders the `Viral::BaseComponent` passing the system arguments, in this case this results in a `section` tag with the specified `classes`.

With that all being said, for this PR I would like the button to be removed and instead just have the modal render open by default, which suits the purpose of your other PR. We can later revisit the best way to have a button launch the modal.

app/components/viral/modal_component.html.erb
ksierks: I used developer tools to inspect the modal body, then copied the xpath and selector options. This is what it returned:

xpath
/div/div/div/div[2]

selector
div > div > div > div.p-6.space-y-6.text-gray-500.dark\:text-gray-400

I liked xpath, but assert_selector uses css selectors. I changed it to div > div > div > div:nth-child(2).

Please see f549f635c1b8a63fdf655ffd8f2fe375284ce34c.

app/components/viral/modal_component.html.erb
ksierks: Both potential solutions of a partial and button within component reverted.

test/components/previews/modal_component_preview.rb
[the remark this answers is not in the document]
joshsadam: Decided to remove this functionality for now.

test/components/viral/system/modal_component_test.rb
```diff
@@ -0,0 +1,25 @@
...
+      within('.Viral-Preview > [data-controller-connected="true"]') do
+        assert_text title
+      end
+    end
+
+    test 'with title and body' do
+      body = 'This is a modal body'
+      visit('rails/view_components/modal_component/default')
+      within('.Viral-Preview > [data-controller-connected="true"]') do
+        assert_selector(:xpath, './div/div/div') do
```
joshsadam, line 19: **chore (blocking):** Remove the assert_selector as we are already within the component, and you are checking that the text is there

test/components/viral/system/modal_component_test.rb
```diff
@@ -0,0 +1,25 @@
...
+
+require 'application_system_test_case'
+
+module System
+  class ModalComponentTest < ApplicationSystemTestCase
+    test 'with title' do
+      title = 'This is a modal title'
+      visit('rails/view_components/modal_component/default')
+      within('.Viral-Preview > [data-controller-connected="true"]') do
+        assert_text title
```
joshsadam, line 11: **chore (blocking):** Add `assert_accessible` call within the `within` blocks on both these tests.  Accessibility is currently failing on the modals.

joshsadam (changes requested):

app/components/viral/modal_component.html.erb
ericenns: Instead of relying on the dom structure, I think it can be simplified to just the `assert_text` line. We are already limiting the asserts to the content of the modal with the `within` statement. In general we only need to assert that the specific content is visible to the user, as they do not necessarily care that it is in a div just that it is visible.

test/components/viral/system/modal_component_test.rb
ksierks: In the [capybara](https://rubydoc.info/github/teamcapybara/capybara#beware-the-xpath--trap) documentation that was provided, it says that the .// expression means "any descendant of the current node". Yes, I'm in the component, but the component is made up of a title and body and I was trying to be more specific and look in the body for the text. 
Is this closer to what you're thinking? Please see f07000c01022cbf43440954ed68fe3bf02078196.

test/components/viral/system/modal_component_test.rb
ksierks: Please see 680529043f3d0e344215004ae09a7632a2c3fb6c.

app/components/viral/modal_component.html.erb
ksierks: Yup, please see the making test less specific commit, f07000c01022cbf43440954ed68fe3bf02078196.

test/components/viral/system/modal_component_test.rb
joshsadam: Thanks for adding the `assert_accessible`

app/components/viral/modal_component.html.erb
```diff
@@ -0,0 +1,71 @@
+<div data-controller="viral--modal-component">
+  <!-- Main modal -->
+  <div
+    data-viral--modal-component-target="modal"
+    tabindex="-1"
+    aria-hidden="true"
+    aria-label="Modal"
```
joshsadam, line 7: **chore (blocking):** Sorry, I thought I wrote this here, but it has disappeared so if it repeats twice sorry.  By setting `aria-label` to be "Modal" it describes all modals that way instead of something explicit like "Confirm deletion of project?".  A better way to handle this can be found by looking at [MDN's `aria-modal`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-modal) write up and example:

```html
<div id="backdrop" class="no-scroll">
  <div
    role="alertdialog"
    aria-modal="true"
    aria-labelledby="dialog_label"
    aria-describedby="dialog_desc">
    <h2 id="dialog_label">Confirmation</h2>
    <div id="dialog_desc">
      <p>Are you sure you want to delete this file?</p>
    </div>
    <button type="button" onclick="closeDialog(this)">
      No. Close this popup.
    </button>
    <button type="button" onclick="deleteFile(this)">
      Yes. Delete the file.
    </button>
  </div>
</div>
```

- `role="alertdialog"` : From [MDN ARIA: alertdialog role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alertdialog_role) - "The alertdialog role is to be used on modal alert dialogs that interrupt a user's workflow to communicate an important message and require a response."
- `aria-labelledby`: They use the id for the modal title, which is probably a great way to set it.

app/components/viral/modal_component.html.erb
```diff
@@ -0,0 +1,77 @@
...
+              dark:hover:text-white
+            "
+            data-action="click->viral--modal-component#close"
+          >
+            <%= viral_icon(name: "x_mark", classes: "w-5 h-5") %>
+            <span class="sr-only"><%= t("components.modal.close") %></span>
+          </button>
+        </div>
+        <!-- Modal body -->
+        <div id="dialog_desc" class="p-6 space-y-6 text-gray-500 dark:text-gray-400">
```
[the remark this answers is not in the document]
ksierks: Please see 3574940b44260fddd8379c56997e15d5ef99a854.

PR #110, opened 2023-05-26
Refactor: Update component previews

joshsadam:
The purpose of this PR is to clean up the way previews are generated for our view components.  Currently previews are generated using rails markup which leads to an incorrect representation of how to use the component in the LookBook source panel.  This update changes to using erb files to generate the preview, which is the same syntax that is used to write the component in regular view files allowing for easier adoption of the components.

* If the component is prefixed with `viral_` it is part of the _Viral Component Library_

**Source Before:**
image

**Source After:**
image

All updated components previews can be viewed within our [LookBook](http://localhost:3000/rails/lookbook/) while the development server is running.

## PR acceptance checklist


app/components/viral/dropdown_component.rb
```diff
@@ -20,6 +20,7 @@ def initialize(label: nil, icon: nil, caret: false, trigger: TRIGGER_DEFAULT,
       @trigger = TRIGGER_MAPPINGS[trigger]
 
       @system_arguments = system_arguments
+      @system_arguments[:id] ||= "dd-#{SecureRandom.hex(4)}"
```
deepsidhu85, line 23: Could this possibly lead to an issue of duplicate ids?

deepsidhu85 (changes requested):
This looks good to me just the one question below. Great work on the component library so far!

app/components/viral/dropdown_component.rb
joshsadam: I increased it to 10 so I think that there should any duplicates. Thanks

joshsadam: I increased it to 10 so I think that there should any duplicates. Thanks

deepsidhu85 (approved):

PR #113, opened 2023-06-02
Feature: Soft Delete Records

### LINKED ISSUE #108: Soft Deleting Records
### Problem

Currently, IRIDA Next does not have functionality to restore deleted records. To add this functionality, we will want to use a gem `paranoia` which allows soft deleting of records and the record can be restored as needed. See docs for [paranoia](https://www.ruby-toolbox.com/projects/paranoia) for more information

### Implementation
1) Add gems `paranoia` to the `Gemfile` and run the command `bundle`
2) Follow the instructions for [paranoia](https://www.ruby-toolbox.com/projects/paranoia) to setup soft-deleting of records
3) Update backend code following the docs in step 2 to allow soft-deleting of records
4) Add testing (should include tests to delete and restore records)

### Additional Details
This issue should be implemented after setting up auditing(#107)

deepsidhu85:

This PR adds in functionality to soft delete records instead of permanently deleting them. In conjunction with the logidze (auditing) when we destroy a record we won't lose the log of changes for a record

Resolves #108 

1. Login as admin@email.com
2. View the group `Bacillus` (in the UI and rails console) and make note of all members, subgroups, projects, samples, project members, and all routes
3. In the rails console `rails c` within your terminal find the group `Bacillus` and call destroy on the group
4. Verify with `Model.only_deleted` where Model is Group, Project, Namespaces::ProjectNamespace, Sample, Member; that the records from step 2 are found in the only_deleted scopes. No other records should be found within
5. Verify in the UI that the Bacillus group, its subgroups, members, projects, samples etc cannot be accessed. Also, verify this in the rails console by just calling Model.where, find etc
6. To restore you can either restore individual records, or you can restore a record and its dependently destroyed association records. See [paranoia docs](https://github.com/rubysherpas/paranoia) for more information.
7. Verify that you can delete a record and then create a record with the same parameters. You should be able to create them but if you try to restore the record that was deleted you should receive a uniqueconstraint exception in the rails console. Try this out for the models (Group, ProjectNamespace, Sample, Member, User)

## PR acceptance checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,8 @@ gem 'logidze'
 
 # Database functions
 gem 'fx'
+# Soft delete records
+gem 'paranoia'
 
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -227,6 +227,8 @@ GEM
     orm_adapter (0.5.0)
     pagy (6.0.3)
     parallel (1.23.0)
+    paranoia (2.6.1)
+      activerecord (>= 5.1, < 7.1)
     parser (3.2.2.0)
       ast (~> 2.4.1)
     pg (1.4.6)
@@ -389,6 +391,7 @@ DEPENDENCIES
   omniauth-rails_csrf_protection
   omniauth-saml
   pagy (~> 6.0)
+  paranoia
   pg
   puma (~> 5.0)
   rails (~> 7.0.4, >= 7.0.4.2)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,7 +10,7 @@ class ApplicationController < ActionController::Base
   around_action :use_logidze_responsible, only: %i[create destroy update] # rubocop:disable Rails/LexicallyScopedActionFilter
 
   def use_logidze_responsible(&)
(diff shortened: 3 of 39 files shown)


ericenns (changes requested):
This is working as expected. I can remove a Sample from a Project and add a new one with the name of the old one and then if I try to restore the sample it raises a Postgres unique constraint violation which is expected.

One edge case I found though is that if a user deletes their account, they cannot register a new one with the email address of the old account. An example of this could be triggered by having a local account and then deleting it to be able to log in with an External account that uses the same email address. So likely we need to update the unique constraint for email address on User.

deepsidhu85:
> This is working as expected. I can remove a Sample from a Project and add a new one with the name of the old one and then if I try to restore the sample it raises a Postgres unique constraint violation which is expected.
> 
> One edge case I found though is that if a user deletes their account, they cannot register a new one with the email address of the old account. An example of this could be triggered by having a local account and then deleting it to be able to log in with an External account that uses the same email address. So likely we need to update the unique constraint for email address on User.

Thanks for reviewing it and catching the edge case. I've updated the indexes on the users table and now should we working as expected

ericenns (approved):
This looks good to me. We may come across some edge cases later that I have not yet identified but I think this is worth merging in now.

JeffreyThiessen (commented):
Looks good!

Issue #96, opened 2023-05-11
Add ablity for a member to leave group or project


deepsidhu85:
**Problem**
Currently in IRIDA Next members are only able to be removed from either a group or project by either `owners` and `maintainers`. We want to add the ability for a member to be able to leave by themselves without asking for permission

**Implementation**
- Authorization will need to be updated to allow a member to leave the group or project. Currently `owners` and `maintainers` of a group or project can remove members (owners can remove all access levels, and maintainers can remove members with access levels up to maintainer). In `members/destroy_service.rb` the `unless current_user != member.user` block can be removed, and everything else in the execute method minus the `member.destroy` can be wrapped in an `if current_user != member.user` block
- The translation `services.members.destroy.cannot_remove_self` can be removed
- The views will need to be updated (`app/views/groups/members/index.html.erb` and `app/views/projects/members/index.html.erb`) to allow the `current_user == member.user` to view the `Leave group` or `Leave project` links in the dropdown. Specifically an `or` will need to be added to the `if allowed_to?` method call
- Tests will need to be updated

**Additional Details**

This issue depends on issue #67 and #95 so they will need to be done before this is added

PR #114, opened 2023-06-02
Feature: Leave Group or Project

### LINKED ISSUE #95: Update text on members page for member removal
### Problem

On the members page for both groups and projects, currently the text to remove a member from the group or project just has the text `Remove`. We want to update this to `Remove member`, if the member is not the currently logged in user, and to either read `Leave group` or `Leave project`, if the member is the currently logged in user. The confirmation message would also need to be updated to be in-line with this new text for the case of current user is the member. 

### Implementation

- Take a look at the relevant views `app/views/groups/members/index.html.erb` and `app/views/projects/members/index.html.erb`
- Within the views, look at the viral_dropdown section which has the links for actions (dropdown_item).
- Make a note of the i18n translation keys for `label` and the `turbo_confirm`
- In the translation file `en.yml`, update the value to be in-line with the new text `Remove member`
- Also, add in new translation key/values for `Leave group`, `Leave project`, and confirmation message to match these new values
- Once the `en.yml` file has been updated you will need to conditionally display the correct translation key if the current user is the member or not
- You will also need to update the system tests to ensure the correct text is displayed

### Additional Details

Currently this is what the members pages look like. Members pages are accessible from within a group and within a project

Group members page: (Change to `Leave Group` as the currently logged in user is `admin@email.com` and the user being removed is this user)

image

Project members page: (Change to `Remove member` as the currently logged in user is `admin@email.com` and the user being removed is not this user)

image

- To view the current confirmation message you can click on the `Remove` for a member

deepsidhu85:

This PR adds functionality for a member to be able to leave a group or project on their own without needing an owner or maintainer to remove them.

Resolves #95 and #96 

**Logged in as user that is removing themselves from the group:**

image

remove-confirmation

**Removing a member that is not the current logged in user**

image

remove-confirmation-remove-member

1. Login as `admin@example.com`
3. Click the ellipsis for the `admin@example.com` user and you should see a link `Leave group`
4. In the confirmation you should see `Are you sure you want to remove yourself from the group?`
5. Try step 3 for another user in the listing and you should see a link `Remove member`
6. In the confirmation you should see `Are you sure you want to remove this member from the group?`
7. Try out the same steps but for a project. Projects from the seed data only have user6 through user8 as members so you will need to log in as one of those users.

## PR acceptance checklist

### DIFF
--- a/app/javascript/controllers/selection_controller.js
+++ b/app/javascript/controllers/selection_controller.js
@@ -1,0 +1,52 @@
+import { Controller } from "@hotwired/stimulus";
+
+export default class extends Controller {
+  static targets = ["rowSelection"];
+  static values = {
+    storageKey: {
+      type: String,
+      default: location.protocol + "//" + location.host + location.pathname,
+    },
+  };
+
+  connect() {
+    this.element.setAttribute("data-controller-connected", "true");
+
+    const storageValue = JSON.parse(
+      sessionStorage.getItem(this.storageKeyValue)
+    );
+
+    if (storageValue) {
+      this.rowSelectionTargets.map((row) => {
+        if (storageValue.indexOf(row.value) > -1) {
+          row.checked = true;
+        }
+      });
+    } else {
+      this.save([]);
+    }
+  }
+
+  toggle(event) {
+    const newStorageValue = JSON.parse(
+      sessionStorage.getItem(this.storageKeyValue)
+    );
+
+    if (event.target.checked) {
+      newStorageValue.push(event.target.value);
+    } else {
+      const index = newStorageValue.indexOf(event.target.value);
+      if (index > -1) {
+        newStorageValue.splice(index, 1);
+      }
+    }
+    this.save(newStorageValue);
+  }
+
+  save(storageValue) {
+    sessionStorage.setItem(
+      this.storageKeyValue,
+      JSON.stringify([...storageValue])
+    );
+  }
+}
--- a/test/components/previews/selection_preview.rb
+++ b/test/components/previews/selection_preview.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class SelectionPreview < ViewComponent::Preview
+  # The selection controller stores the values of checkboxes in session storage. The default storage key is the
+  # url. When a user toggles a checkbox, the value is updated in session storage.
+  def default; end
+
+  # The selection controller stores the values of checkboxes in session storage. The default storage key is the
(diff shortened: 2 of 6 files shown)


joshsadam:
I think this is great, thanks Deep.  I personally like the browser alert confirmation message, but not sure if @ericenns would prefer a modal.  I think we should come to a consensus on this, and implement going forward.  The other thing I have been noticing is sometimes we are using and flash alert (top right of screen) and sometimes we just use an on page alert banner, again, let's just come to a consensus  on where and when to use those and we can add that to the UI docs to keep things standardized.

app/views/projects/members/index.html.erb
```diff
@@ -85,15 +85,42 @@
...
+                <% if current_user == member.user || allowed_to?(:destroy_member?, @namespace) %>
                   <td class="px-4 py-3 text-right">
                     <%= viral_dropdown(icon: "ellipsis_vertical", classes: "member-settings-ellipsis", aria: { label: t(:'projects.members.index.actions.member_dropdown_aria_label', member: member.user.email) }) do |dropdown| %>
                       <%= dropdown.item(
-                        label: t(:"projects.members.index.remove"),
+                        label:
+                          (
+                            if current_user == member.user
+                              t(:"projects.members.index.leave_project")
+                            else
```
joshsadam, line 96: **nitpick (blocking):** I think since you have multiple `current_user == member.user` you should create a partial for each.  This would require you to only need to have one clause.

deepsidhu85: Definitely was a good idea..thanks! Updated in c77f10c5b6582c4430ebf914ac27cffbb6ce464b

joshsadam:
So I logged in as `user1` and removed myself from a group I owned.   This works, but leave the page in a confusing state:



I think it would be better if the user remove itself from a group / project that they are redirected to the dashboard page.  Thoughts?

app/views/groups/members/_current_user_is_member.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= viral_dropdown(icon: "ellipsis_vertical",
```
joshsadam, line 1: **question (non-blocking):** Gitlab does not use dropdowns for these, I am assuming because it is a single action and is partially hidden in the dropdown.  Should we move these out to buttons on the table row?  Thoughts @ericenns

joshsadam (commented):
Just one random question the rest looks really good from my point of view.  @ericenns can you double check the `*.rb` files.

PR #117, opened 2023-06-06
Transfer samples service

ksierks:
The sample transfer service moves samples between projects. This functionality is restricted to owners.
Pulled from https://github.com/phac-nml/irida-next/pull/90.

N/A

1.  Call the sample transfer service within the rails console:
`Samples::TransferService.new(project, user).execute(sampleTransfer)`
For examples, see test/services/samples/transfer_service_test.rb.
3. Verify only owners can call the service.
4. Verify you can't move samples to the same project.

## PR acceptance checklist


app/models/sample_transfer.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# entity class for SampleTransfer
+class SampleTransfer
```
deepsidhu85, line 4: Since the samples transfer service only takes 2 parameters we don't really need a special class for them. We can just pass the parameters directly to the `execute` method of the transfer service. So please remove this class and update accordingly

app/services/samples/transfer_service.rb
```diff
@@ -0,0 +1,50 @@
...
+      return unless @project.id == @sample_transfer.project_id.to_i
+
+      raise TransferError,
+            I18n.t('services.samples.transfer.duplicate_project')
+    end
+
+    def transfer(sample_transfer)
+      @sample_transfer = sample_transfer
+      ActiveRecord::Base.transaction do
+        JSON.parse(@sample_transfer.sample_ids.first).each do |sample_id|
```
deepsidhu85, line 43: Is there a specific reason for using JSON.parse here? We should just be able to pass an array of ids from the controller and then iterate over it without having to use JSON.parse.

Also, can be add a check to make sure the destination project doesn't already contain a sample with the same name as the sample being transferred.

We probably also want to keep track of sample transfers that have errored and once the loop is done to raise a `TransferError`

config/locales/en.yml
```diff
@@ -477,6 +479,9 @@ en:
       update:
         cannot_update_self: "You cannot update your own %{namespace_type} role"
         role_not_allowed: "A maintainer can only update a member's role up to the Maintainer role"
+    samples:
+      transfer:
+        duplicate_project: Please select a different project.
```
deepsidhu85, line 484: Update the text here to be clear of why the user should be selecting a different project

deepsidhu85 (changes requested):
Generally looks good, just have some comments below. The comments for the testing section will be separate

app/policies/project_policy.rb
```diff
@@ -98,9 +98,37 @@ def update_sample?
...
+    return true if record.namespace.parent.user_namespace? && record.namespace.parent.owner == user
+    return true if Member.can_transfer?(user, record.namespace) == true
+
+    details[:name] = record.name
+    false
+  end
+
+  def transfer_sample_into_project?
+    return true if record.namespace.parent.user_namespace? && record.namespace.parent.owner == user
+    return true if Member.can_transfer?(user, record.namespace) == true
```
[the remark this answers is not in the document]
ksierks: Ok, I think that logic is already defined in namespace_owners_include_user? & can_transfer_into_namespace?. I hope it's okay I reused those. Please see e9b52958fdb6ced25f2dd7cb508ed24ac6cbf354 and let me know if that's not right.

app/policies/project_policy.rb
```diff
@@ -98,9 +98,37 @@ def update_sample?
...
+    false
+  end
+
   scope_for :relation do |relation|
     relation.where(namespace: { parent: user.groups.self_and_descendant_ids })
             .include_route.order(updated_at: :desc).or(relation.where(namespace: { parent: user.namespace })
             .include_route.order(updated_at: :desc))
   end
+
+  scope_for :relation, :transferable do |relation|
```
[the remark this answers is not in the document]
ksierks: Please see 1c68e9870cf52381e354459c388f187d2a219660.

1. True, but I'm terrible at naming things. I don't think `sample_transferable` is any better. I know you've used `manageable` for namespaces. I was worried that might be confusing, but this is suppose to list projects which the user is an owner or maintainer of. 
2. Fixed.

app/policies/project_policy.rb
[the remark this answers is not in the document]
ksierks: Please see e9b52958fdb6ced25f2dd7cb508ed24ac6cbf354.

I think I can reuse namespace_owners_include_user? and can_transfer_into_namespace?. Please let me know if that's wrong.

app/models/sample_transfer.rb
ksierks: I can modify the transfer service to only accept 2 params, but I created SampleTransfer to help with the form. SampleTransfer is an active model so I can call ` form_with(model: @sample_transfer)`. I liked being able to use a form like we do for our active records. Would you have me remove SampleTransfer entirely and not use an active model in the form at all?

deepsidhu85: Yeah let's go with removing the `SampleTransfer` entirely. I don't think we should be using forms with models unless the model has a table attached to it. So yeah let's update the transfer service to accept 2 params

app/policies/project_policy.rb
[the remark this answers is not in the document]
deepsidhu85: Oh don't worry I'm horrible with naming stuff too! I think `manageable` would make sense if we were getting all the projects that the user is an owner and maintainer of. Since we are getting n-1 I think we need another name. I'm going to think on a name and we can discuss it tomorrow!

app/policies/project_policy.rb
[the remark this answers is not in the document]
deepsidhu85: Yup it's perfectly fine to use those methods within the `member.rb` from other methods within it!

app/services/samples/transfer_service.rb
ksierks: This has a little story behind it, lol.
In my first implementation I had wrapped the sample table in a form to send the selected samples to the next step. This didn't work, unless I commented out the action buttons that are available for each sample row. I don't know the exact cause, but stimulus broke the form. This is why the selection controller was implemented. I am populating a hidden field in the form that calls the transfer service. The form is sending that data in the following format:
image
I don't know how else to populate an array in a hidden field with JavaScript.

app/models/sample_transfer.rb
ksierks: Quick question. I've started removing `SampleTransfer`, I just don't know what I should be attaching the errors to be displayed on the page.

app/policies/project_policy.rb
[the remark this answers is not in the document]
deepsidhu85: I think I got a name for the scope, `receivable`. Let me know what you think or if you have any other potential names

app/models/sample_transfer.rb
deepsidhu85: I believe we can just attach them to the @new_project variable and then have the execute method return the @new_project to the controller

app/services/samples/transfer_service.rb
deepsidhu85: Oh I see. I will take a look at that. Which file is it in?

ksierks: I populate the hidden field with the form_controller.js & app/views/projects/samples_transfer/new.html from within the transfer_samples branch.

app/policies/project_policy.rb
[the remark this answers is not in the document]
ksierks: Ok, will do.

app/models/sample_transfer.rb
ksierks: I might not have a defined @new_project at the time the error occurs. I might need to reach out to you about this one... I'm sorry.

ksierks: As discussed, I am to append the errors to the current project. Please see 018e40ae38dddb1ed820bb0e72bc6ba699544a67.

config/locales/en.yml
ksierks: I think I updated this in the last commit. Please see 018e40ae38dddb1ed820bb0e72bc6ba699544a67.

app/policies/project_policy.rb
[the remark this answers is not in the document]
deepsidhu85: Hey Katherine, sorry let's change this scope to `manageable` since I missed that we weren't actually removing the current project from the listing. From wherever the scope is being called you can add a `where.not(id: ID_OF_NAMESPACE)` which should remove it from the listing of projects that the sample(s) can be transferred to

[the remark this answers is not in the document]
ksierks: Please see 928caa1d31798becc289afd1f333959cb00d3621. This branch doesn't have where the scope is being called, but I did update it in the transfer-samples branch.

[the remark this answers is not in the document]
deepsidhu85: Thanks!

ericenns (approved):
Looks good to me! Thanks for making all the requested changes from other reviewers!

deepsidhu85 (approved):
Looks good to me! Thanks for getting this all updated.

PR #118, opened 2023-06-08
Selection controller

ksierks:
The selection controller is used to store a list of checkbox values in a hash within local storage. The url being the local storage key.
Pulled from https://github.com/phac-nml/irida-next/pull/90.

image

1. Navigate to http://localhost:3000/rails/lookbook.
2. Click on "Selection" previews.
3. Ensure the checkbox values are stored in local storage with the key being the url.
4. Ensure the latest value is populated on page refresh.

## PR acceptance checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -83,6 +83,7 @@ gem 'action_policy-graphql'
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem 'debug', platforms: %i[mri mingw x64_mingw]
+  gem 'lookbook', '>= 2.0.0.rc.1'
 end
 
 group :development do
@@ -103,7 +104,6 @@ group :development do
   # LookBook
   gem 'actioncable'
   gem 'listen'
-  gem 'lookbook', '>= 2.0.0.rc.1'
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -83,7 +83,6 @@ gem 'action_policy-graphql'
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem 'debug', platforms: %i[mri mingw x64_mingw]
-  gem 'lookbook', '>= 2.0.0.rc.1'
 end
 
 group :development do
@@ -104,6 +103,7 @@ group :development do
   # LookBook
   gem 'actioncable'
   gem 'listen'
+  gem 'lookbook', '>= 2.0.0.rc.1'
 end
 
 group :test do
--- a/app/javascript/controllers/selection_controller.js
+++ b/app/javascript/controllers/selection_controller.js
@@ -10,33 +10,36 @@ export default class extends Controller {
   };
 
   connect() {
-    this.localStorageValue = new Map(
-      JSON.parse(localStorage.getItem(this.storageKeyValue))
-    );
-
-    if (this.localStorageValue.size === 0) {
+    if (localStorage.getItem(this.storageKeyValue)) {
       this.rowSelectionTargets.map((row) => {
-        this.localStorageValue.set(row.value, row.checked);
-      });
-      this.save();
-    } else {
-      this.rowSelectionTargets.map((row) => {
-        if (this.localStorageValue.get(row.value)) {
+        if (
+          new Map(JSON.parse(localStorage.getItem(this.storageKeyValue))).get(
+            row.value
+          )
+        ) {
           row.checked = true;
         }
       });
+    } else {
+      let newLocalStorageValue = new Map();
+      this.rowSelectionTargets.map((row) => {
+        newLocalStorageValue.set(row.value, row.checked);
+      });
+      this.save(newLocalStorageValue);
     }
   }
 
   toggle(event) {
-    this.localStorageValue.set(event.target.value, event.target.checked);
-    this.save();
+    let newLocalStorageValue = new Map(
+      JSON.parse(localStorage.getItem(this.storageKeyValue))
+    ).set(event.target.value, event.target.checked);
+    this.save(newLocalStorageValue);
   }
 
-  save() {
+  save(localStorageValue) {
     localStorage.setItem(
       this.storageKeyValue,
-      JSON.stringify([...this.localStorageValue])
+      JSON.stringify([...localStorageValue])
     );
   }
 }
--- a/app/javascript/controllers/selection_controller.js
+++ b/app/javascript/controllers/selection_controller.js
@@ -10,10 +10,10 @@ export default class extends Controller {
   };
 
   connect() {
-    if (localStorage.getItem(this.storageKeyValue)) {
+    if (sessionStorage.getItem(this.storageKeyValue)) {
       this.rowSelectionTargets.map((row) => {
         if (
-          new Map(JSON.parse(localStorage.getItem(this.storageKeyValue))).get(
+          new Map(JSON.parse(sessionStorage.getItem(this.storageKeyValue))).get(
             row.value
           )
         ) {
@@ -21,25 +21,25 @@ export default class extends Controller {
         }
       });
     } else {
-      let newLocalStorageValue = new Map();
+      let newStorageValue = new Map();
       this.rowSelectionTargets.map((row) => {
-        newLocalStorageValue.set(row.value, row.checked);
+        newStorageValue.set(row.value, row.checked);
       });
-      this.save(newLocalStorageValue);
+      this.save(newStorageValue);
     }
   }
 
   toggle(event) {
-    let newLocalStorageValue = new Map(
-      JSON.parse(localStorage.getItem(this.storageKeyValue))
+    let newStorageValue = new Map(
+      JSON.parse(sessionStorage.getItem(this.storageKeyValue))
     ).set(event.target.value, event.target.checked);
-    this.save(newLocalStorageValue);
+    this.save(newStorageValue);
   }
 
-  save(localStorageValue) {
-    localStorage.setItem(
+  save(storageValue) {
+    sessionStorage.setItem(
       this.storageKeyValue,
-      JSON.stringify([...localStorageValue])
+      JSON.stringify([...storageValue])
     );
   }
 }
--- a/app/javascript/controllers/selection_controller.js
+++ b/app/javascript/controllers/selection_controller.js
@@ -12,29 +12,40 @@ export default class extends Controller {
   connect() {
     this.element.setAttribute("data-controller-connected", "true");
 
-    if (sessionStorage.getItem(this.storageKeyValue)) {
+    const storageValue = JSON.parse(
+      sessionStorage.getItem(this.storageKeyValue)
+    );
+
+    if (storageValue) {
       this.rowSelectionTargets.map((row) => {
-        if (
-          new Map(JSON.parse(sessionStorage.getItem(this.storageKeyValue))).get(
-            row.value
-          )
-        ) {
+        if (storageValue.indexOf(row.value) > -1) {
           row.checked = true;
         }
       });
     } else {
-      let newStorageValue = new Map();
+      const newStorageValue = [];
       this.rowSelectionTargets.map((row) => {
-        newStorageValue.set(row.value, row.checked);
+        if (row.checked) {
+          newStorageValue.push(row.value);
+        }
       });
       this.save(newStorageValue);
     }
   }
 
   toggle(event) {
-    let newStorageValue = new Map(
-      JSON.parse(sessionStorage.getItem(this.storageKeyValue))
-    ).set(event.target.value, event.target.checked);
+    const newStorageValue = JSON.parse(
+      sessionStorage.getItem(this.storageKeyValue)
+    );
+
+    if (event.target.checked) {
+      newStorageValue.push(event.target.value);
+    } else {
+      const index = newStorageValue.indexOf(event.target.value);
+      if (index > -1) {
+        newStorageValue.splice(index, 1);
+      }
+    }
     this.save(newStorageValue);
   }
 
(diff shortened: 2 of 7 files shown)


test/components/previews/selection_preview/default.html.erb
```diff
@@ -0,0 +1,11 @@
+<div data-controller="selection">
```
deepsidhu85, line 1: For this and the preview file below can I get you to add a few more items with checkboxes

deepsidhu85 (changes requested):
The code here looks good to me! Just the one comment below about adding a few more items to the views

test/components/previews/selection_preview/default.html.erb
ksierks: Please see c182d2c3bbcf38c96a7dcd8f3e6ff1880d3c5124.

Gemfile
```diff
@@ -83,6 +83,7 @@ gem 'action_policy-graphql'
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem 'debug', platforms: %i[mri mingw x64_mingw]
+  gem 'lookbook', '>= 2.0.0.rc.1'
```
deepsidhu85, line 86: Was there a specific reason for moving the lookbook gem from the development group into this group?

deepsidhu85 (changes requested):
Just a question below

Gemfile
ksierks: I moved it, because I got these errors...

image

test/components/previews/selection_preview.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class SelectionPreview < Lookbook::Preview
```
joshsadam, line 3: **chore (blocking):** This should be a `ViewComponent::Preview` not `Lookbook::Preview`

Gemfile
joshsadam: This branch is the only one that has tests failing without this modification to the Gemfile.  I did some looking around and it appears to be caused in the file `selection_preview.rb`.  Try reverting the `Gemfile` and updating line 3 of `selection_preview.rb` to be:
```
class SelectionPreview < ViewComponent::Preview
```
This is how the other previews are set up.

test/components/previews/selection_preview.rb
```diff
@@ -0,0 +1,12 @@
...
+  # The selection controller stores the values of checkboxes in local storage. The local storage key is the url.
+  # When a user toggles a checkbox, the value is updated in local storage.
+  def default; end
+
+  # The selection controller stores the values of checkboxes in local storage. The local storage key is the url.
+  # This example uses the selection controller within a table to select/deselect rows.
+  # When a user toggles a checkbox, the value is updated in local storage.
+  def within_a_table; end
+end
```
joshsadam, line 12: **chore (blocking):** Please add a tests for these.  You will probably want to include a page refresh of some sort to ensure that the checkboxes are still selected on reload.

test/components/previews/selection_preview/default.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <% (0...3).each do |index| %>
+    <div>
+      <input
```
joshsadam, line 4: **nitpick (blocking):** I know we do not have a component for this, but can you please keep it in line with the Flowbite style: https://flowbite.com/docs/components/forms/#checkbox

test/components/previews/selection_preview/within_a_table.html.erb
```diff
@@ -0,0 +1,19 @@
+<table data-controller="selection">
+  <tbody>
```
joshsadam, line 2: **nitpick (blocking):** Same comment here, we do not have a table component, but when I looked at the preview I though it was still a list, can you make it look more like a table by using some of Flowbite styles: https://flowbite.com/docs/components/tables/#default-table

joshsadam (changes requested):
Thanks for working on this, can you please update the following

test/components/previews/selection_preview.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class SelectionPreview < Lookbook::Preview
+  # The selection controller stores the values of checkboxes in local storage. The local storage key is the url.
+  # When a user toggles a checkbox, the value is updated in local storage.
+  def default; end
+
+  # The selection controller stores the values of checkboxes in local storage. The local storage key is the url.
```
joshsadam, line 8: **praise (non-blocking):** I like the way these add comments into the note section of the Lookbook preview.  Nice!

Gemfile
ksierks: I didn't think I was implementing a ViewComponent, just a stimulus controller. I think that's why I thought I should use Lookbook::Preview instead of ViewComponent::Preview. My implementation isn't a subclass of ViewComponent::Base. But it fixes the error. Please see f6d382edd6ade340811620b7511566a80dd0e8de.

app/javascript/controllers/selection_controller.js
```diff
@@ -0,0 +1,38 @@
+import { Controller } from "@hotwired/stimulus";
+
+export default class extends Controller {
+  static targets = ["rowSelection"];
+
+  connect() {
+    this.localStorageKey =
+      location.protocol + "//" + location.host + location.pathname;
```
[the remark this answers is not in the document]
ksierks: I thought I was asked to use the url as the key. This opens it up to anything. What if we pass a value, but make the url the default? Please see 4efdfbe9d81208f524dc614f679420a15379bffb.

test/components/previews/selection_preview.rb
ksierks: This is a big miss. Please see 2ed1488b97b1d7024d1c38fa0b3ed54b874d14de.

test/components/previews/selection_preview/within_a_table.html.erb
ksierks: I thought the selection controller should be the :star: of the preview. Please see e019021985ff32e2b8378ecd2a0fa724db063c2c.

test/components/previews/selection_preview/default.html.erb
ksierks: I thought the selection controller should be the :star: of the preview. Please see e019021985ff32e2b8378ecd2a0fa724db063c2c.

app/javascript/controllers/selection_controller.js
```diff
@@ -0,0 +1,38 @@
...
+    this.localStorageValue = new Map(
+      JSON.parse(localStorage.getItem(this.localStorageKey))
+    );
+
```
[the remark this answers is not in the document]
ksierks: Please see ff7af67af5aeec5034d4f78d0523bff63d289eb7.

app/javascript/controllers/selection_controller.js
```diff
@@ -0,0 +1,38 @@
+import { Controller } from "@hotwired/stimulus";
+
+export default class extends Controller {
+  static targets = ["rowSelection"];
+
+  connect() {
+    this.localStorageKey =
+      location.protocol + "//" + location.host + location.pathname;
+    this.localStorageValue = new Map(
+      JSON.parse(localStorage.getItem(this.localStorageKey))
```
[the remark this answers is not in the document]
ksierks: Hopefully, I understood this correctly. Please see ddcaa13a97ff027fe68c7235798586ed88b0985a for removing the in-memory storage value.

test/components/previews/selection_preview.rb
ksierks: Thanks.

app/javascript/controllers/selection_controller.js
```diff
@@ -0,0 +1,38 @@
+import { Controller } from "@hotwired/stimulus";
+
+export default class extends Controller {
+  static targets = ["rowSelection"];
```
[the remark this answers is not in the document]
ksierks: Disregarding for now.

app/javascript/controllers/selection_controller.js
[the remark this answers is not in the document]
joshsadam: Having a default value is a fantastic idea :) The only one thing I would ask is that you make a LookBook preview showing how to use the non-default value.  I know visually it won't change anything in the preview, but it could help someone using this controller know what is going on.

joshsadam (changes requested):
Just a few more things to consider with the selection_controller.  Thanks for the updates you made easlier.

app/javascript/controllers/selection_controller.js
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  connect() {
+    this.element.setAttribute("data-controller-connected", "true");
+
+    if (sessionStorage.getItem(this.storageKeyValue)) {
+      this.rowSelectionTargets.map((row) => {
+        if (
+          new Map(JSON.parse(sessionStorage.getItem(this.storageKeyValue))).get(
+            row.value
```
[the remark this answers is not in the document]
ericenns: I think the bigger reasoning to use an `Array` instead of the `Map` is that we are only using the map to store a boolean state (i.e. checked or not checked).

My original idea of using a Map was not to maintain unchecked items in it. I.e. if the checkbox was unchecked we delete the key from the map and if it is checked we add the key to the map if it doesn't exist in it. Using the Map to store the checked state as a boolean instead makes it a little hard to use local storage later on. I.e. you have to check the value to decide to include it or not.

Thinking a bit more about this, an `Array` is probably the nicer solution. Yes it makes the js code a little less simple in that we have to decide to `splice` the array to remove items, after using `indexOf` to determine the location of the item. But then in other methods that need to use the selection we simply have an `Array` that can likely be sent to the server without any modification. Hopefully that all makes sense.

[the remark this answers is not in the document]
ericenns: In this case presence in the `Array` would mean that the row is checked and absence would mean that the row is unchecked.

app/javascript/controllers/selection_controller.js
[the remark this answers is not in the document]
ksierks: I tried. Please see 62ea06896306b4be73affc9b5cfcb161c667ddc7.

app/javascript/controllers/selection_controller.js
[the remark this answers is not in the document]
ksierks: Please see 02bb7249f8df9fc1fbbb4a15977bcbbcb43a4c45.

app/javascript/controllers/selection_controller.js
```diff
@@ -0,0 +1,47 @@
...
+    },
+  };
+
+  connect() {
+    this.element.setAttribute("data-controller-connected", "true");
+
+    if (sessionStorage.getItem(this.storageKeyValue)) {
+      this.rowSelectionTargets.map((row) => {
+        if (
+          new Map(JSON.parse(sessionStorage.getItem(this.storageKeyValue))).get(
```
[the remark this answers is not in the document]
ksierks: Well, you did mention earlier that JSON.parse was speedy... I dunno.
Should be fixed in the array commit. Please see 02bb7249f8df9fc1fbbb4a15977bcbbcb43a4c45.

app/javascript/controllers/selection_controller.js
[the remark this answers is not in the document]
joshsadam: Does what I needed it to do, thank you.

test/components/previews/selection_preview/default.html.erb
joshsadam: Thanks, much better

test/components/previews/selection_preview/within_a_table.html.erb
joshsadam: Thanks, much better

app/javascript/controllers/selection_controller.js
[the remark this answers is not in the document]
joshsadam: Thanks for updating this.  Works the way I would like.

test/components/previews/selection_preview.rb
joshsadam: I do not know why you say this is a big mess, I think this is coming really well.  Thanks for making the preview go back to the page.

app/javascript/controllers/selection_controller.js
```diff
@@ -0,0 +1,58 @@
...
+      sessionStorage.getItem(this.storageKeyValue)
+    );
+
+    if (storageValue) {
+      this.rowSelectionTargets.map((row) => {
+        if (storageValue.indexOf(row.value) > -1) {
+          row.checked = true;
+        }
+      });
+    } else {
```
joshsadam, line 25: **nitpick (blocking):** There is actually no need to loop through the row targets and see if they are selected since the `connect` method only get run when the stimulus controller connects to the page.  You will need to set an empty array into storage though if one does not exist.

joshsadam (changes requested):
Sorry just two last things then this is good to go from my end. Thanks for making the updates

app/javascript/controllers/selection_controller.js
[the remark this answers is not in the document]
joshsadam: Thanks

app/javascript/controllers/selection_controller.js
ksierks: Fair enough. Please see d0afd614ed19b17be12e69e03229978532a8ae5c.

joshsadam: Thanks

joshsadam (approved):
Thanks for all the updates, I think this is good to go from my point of view.

joshsadam (approved):
Thanks for making those testing changes.

deepsidhu85 (approved):
Looks good to me!

PR #119, opened 2023-06-08
Update: Membership source

### LINKED ISSUE #111: Update members page to show inherited members
## Problem

On the members page for both groups and projects, currently you can only see members that are directly added to the group or project, and not the inherited ones (from parent groups).  We want to update the column that currently states who added the user to the project, to by the membership type.

### Membership Types
Users can become members of a group or project in different ways, which define their membership type. (this is from [GitLab's Members of a project page](https://docs.gitlab.com/ee/user/project/members/))

membership

|Membership type|Membership process|
|--------------------------|------------------------------|
|Direct|The user is added directly to the current group or project|
|Inherited|The user is a member of an ancestor group or project that is added to the current group or project|
|Direct shared|The user is a member of a group or project that is shared into the current group or project|
|Inherited shared|The user is a member of an ancestor of a group or project that is shared into the current group or project.|

*   Views in both `app/views/groups/members/index.html.erb` and `app/views/projects/members/index.html.erb` will need to be updated.
*   Update the column displaying the "Added by" details, to reflect GitLab 's `source` column on their members table.
    *   For members directly added to the group or project set them as "Direct Member by ..."
    *   For members added through inheritance, add the parent group (including link to group) that gives them access.

image

deepsidhu85:

Updated members tables for groups and projects to include a `source` column which displays if the member is a `Direct Member` or displays the group name with a link to it, if the membership is inherited from a parent group.

Resolves #111

image

image

image

1. Login as `admin@email.com`
3. You should see a listing of the members and in the source column you should see `Direct Member`
4. Now go to the `Bacillus cerus` group and view it's members
5. You should see the same members listed from step 3 but now under the source column there should be a link to the `Bacillus` group in the source column for each member
6. Go to the `Outbreak 2021` project under the `Bacillus cerus` group
7. You should see user6@email.com, user7@email.com, and user8@email.com with `Direct Member` listed under the source column. These members are also a part of the group but they are also added as direct members to the project. `admin@email.com` and `user9@email.com` have the group link from which they are inherited. You will need to add `user9@email.com` to the `Bacillus` group to see them in the listing for the project.
8. Try this out with different scenarios such as where the user is a member of a parent group, and is a direct member of a subgroup within the parent group, and is a direct member of a project within the subgroup. Try removing the member as a direct member from the project. Under the project members table you should see the user is still in the listing but inherited from the subgroup with a link to it. Now try removing them from the subgroup. View the members for the subgroup and project. You should see the member is inherited from the parent group with a link to it.

**Note: An inherited member's access level can only be updated via the group that they are inherited from. Only a direct member's access level can be updated for both groups and projects**

## PR acceptance checklist


app/views/groups/members/index.html.erb
```diff
@@ -16,6 +16,25 @@
       <div class="shadow">
         <%= turbo_frame_tag("member-update-alert") %>
         <table class="min-w-full table-fixed dark:divide-gray-600">
+          <thead>
```
joshsadam, line 19: **suggestion (blocking):** Not a fan of the table header design.  I have been mostly following what Flowbit does for their stuff.   This is their table header:

image

@ericenns made a modification of that for his "Active Personal Access Tokens" table that you could take a look at.  Some point we will need to make a component for that.

app/views/projects/members/_updated_at.html.erb
```diff
@@ -0,0 +1,4 @@
+<span class="text-sm">
```
joshsadam, line 1: **suggestion (if-minor):** This is exactly the same as `projects/updated_at.html`. Could you make a component that take two parameters like `description` and `updated` and use it in both places (and anywhere else that would need it?

app/views/projects/members/index.html.erb
```diff
@@ -16,6 +16,25 @@
       <div class="shadow">
         <%= turbo_frame_tag("member-update-alert") %>
         <table class="min-w-full table-fixed dark:divide-gray-600">
+          <thead>
+            <tr
```
joshsadam, line 20: **chore (blocking):** Same comment here about the table header.  Would probably be really easy to make a re-usable component for this.  Let me know if you want me do to that and I will make it and get it in to main so you can use it

joshsadam (dismissed):
Just a quick look through, and I really like the way this is looking.  Just a few suggestions to kick things off.

app/views/groups/members/index.html.erb
deepsidhu85: I updated the table header design to follow how it looks on the personal access tokens page for now. Once we create a component for the table header we can update these pages. Updated in d212883497cdc5da02bc5ca354c0cea05d7f9afd

app/views/projects/members/_updated_at.html.erb
deepsidhu85: Created a component for displaying the updated at time ago in d212883497cdc5da02bc5ca354c0cea05d7f9afd

app/views/projects/members/index.html.erb
deepsidhu85: As per the comment above we will update this in a different pr to create a table header component

app/views/groups/members/index.html.erb
joshsadam: Looks great, thanks.

app/views/projects/members/_updated_at.html.erb
joshsadam: Perfect, thanks

app/views/projects/members/index.html.erb
joshsadam: Sounds good

app/components/updated_component.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# A component for displaying when a record was updated.
+class UpdatedComponent < ViewComponent::Base
+  attr_reader :description, :updated_at
+
```
joshsadam, line 6: **suggestion (non-blocking):** +I like this a lot, but can you add a preview for it so that it is easy to find if we are looking for it?

app/views/groups/members/_updated_at.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render UpdatedComponent.new(
+  description: t(:"groups.members.index.updated"),
+  updated_at: distance_of_time_in_words(Time.now, member.updated_at)
```
joshsadam, line 3: **thought (blocking):** Wouldn't it make more sense to put the `distance_of_time_in_words` into your component, that way it does not need to be setup every time you use the component?

app/views/projects/members/index.html.erb
```diff
@@ -72,21 +81,36 @@
...
-                      <%= I18n.l(member.updated_at, format: :default) %>
+                      <% if member.created_at < member.updated_at %>
+                        <%= render partial: "projects/members/updated_at", locals: { member: member } %>
+                      <% end %>
                     <% end %>
                   </div>
                 </td>
-                <% if allowed_to?(:destroy_member?, @namespace) %>
-                  <td class="px-4 py-3 text-right">
+                <% if !member_source.key?(:inherited_namespace_path) && allowed_to?(:destroy_member?, @namespace) %>
```
joshsadam, line 112: **suggestion (if-minor):** This is another complicated comparison, can you make a method to simplify?

app/views/groups/members/_updated_at.html.erb
deepsidhu85: I think that would make too much sense :) Updated in 42c31e6fbb8529ab2a3379eeb513c91043cf1e12

app/views/projects/members/_updated_at.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render UpdatedComponent.new(
+  description: t(:"projects.members.index.updated"),
+  updated_at: distance_of_time_in_words(Time.now, member.updated_at)
```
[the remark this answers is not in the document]
deepsidhu85: Removed in 42c31e6fbb8529ab2a3379eeb513c91043cf1e12. As discussed I was trying to render the UpdateComponent within the turbo_stream.update without using a block and it was crashing.

app/views/projects/members/index.html.erb
deepsidhu85: As discussed this comparison really isn't complicated so leaving as is.

app/components/updated_component.rb
deepsidhu85: Added in 18fb25e3483596a49c608dc46c33935c3c888a3d.

app/controllers/concerns/membership_actions.rb
```diff
@@ -90,7 +90,7 @@ def access_levels # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
   def available_users
     # Remove current user from available users as a user cannot add themselves
     @available_users = User.where.not(id: Member.where(
```
ericenns, line 92: Can we change this to `User.where.not(id: Member.select(:user_id).where(namespace: @namespace)).where.not(id: current_user.id)`

app/policies/member_policy.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    if direct_namespace_members.count.positive?
+      inherited_namespace_members = relation.where('namespace_id IN (?) AND user_id NOT IN (?)',
+                                                   namespace_ids,
+                                                   direct_namespace_members.pluck(:user_id)).uniq(&:user_id)
+    else
+      inherited_namespace_members = relation.where(namespace_id: namespace_ids).uniq(&:user_id)
+    end
+
+    Member.where(id: direct_namespace_members.map(&:id)).or(Member.where(id: inherited_namespace_members.map(&:id)))
```
ericenns, line 23: I think this can be changed to `direct_namespace_members + inherited_namespace`. You can add ActiveRecord::Relations together, so no need for an extra query.

app/models/member.rb
```diff
@@ -121,6 +121,26 @@ def user_has_namespace_maintainer_access?(user, namespace)
...
+        source_namespaces = if namespace.project_namespace?
+                              namespace.parent&.self_and_ancestor_ids
+                            else
+                              namespace.self_and_ancestor_ids
+                            end
+
+        source_namespace = Namespace.find_by(id: Member.where(namespace: source_namespaces,
+                                                              user:).order(:access_level).last.namespace_id)
+
+        { inherited_namespace_path: Rails.application.routes.url_helpers.group_path(source_namespace, only_path: true),
```
ericenns, line 140: we should be using the `group_url` instead of `group_path` as this would break deployments under a subpath.

app/policies/member_policy.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Base policy for member authorization
+class MemberPolicy < ApplicationPolicy
+  scope_for :relation do |relation, options|
+    namespace = options[:namespace]
+    direct_namespace_members = relation.where(namespace:).uniq(&:user_id)
```
ericenns, line 7: Instead of querying direct members and then querying indirect members we could just query it all at once:

e.g.

`relation.where(namespace_id: namespace_ids).order(access_level: :desc).uniq(&:user_id)`

Since `access_level` can only be increased down the hierarchy this should prefer direct members over indirect.

app/models/member.rb
```diff
@@ -121,6 +121,26 @@ def user_has_namespace_maintainer_access?(user, namespace)
                        access_level: Member::AccessLevel::MAINTAINER)
       end
     end
+
+    def membership_source(namespace, user)
```
ericenns, line 125: Please pass `member` instead of `user`

app/models/member.rb
```diff
@@ -121,6 +121,26 @@ def user_has_namespace_maintainer_access?(user, namespace)
...
+      if Member.exists?(
```
ericenns, line 126: Passing `member` this can be changed to `member.namespace_id = namespace.id`

app/models/member.rb
```diff
@@ -121,6 +121,26 @@ def user_has_namespace_maintainer_access?(user, namespace)
...
+        namespace:, user:
+      )
+        { label: I18n.t('activerecord.models.member.direct') }
+      else
+        source_namespaces = if namespace.project_namespace?
```
ericenns, line 131: and in the else, we can just use `member.namespace` instead of having to search for the namespace since we already have it in the `member` model.

app/views/groups/members/index.html.erb
```diff
@@ -26,17 +35,16 @@
             "
           >
             <% @members.each do |member| %>
+              <% member_source = Member.membership_source(@namespace, member.user) %>
```
ericenns, line 38: Due to comments above, this will need to pass `member` instead of `member.user`

app/views/projects/members/index.html.erb
```diff
@@ -26,17 +35,17 @@
             "
           >
             <% @members.each do |member| %>
+              <% member_source = Member.membership_source(@namespace, member.user) %>
```
ericenns, line 38: Due to comments above, this will need to pass `member` instead of `member.user`

app/policies/member_policy.rb
ericenns: This comment is superseded by the comment above, which suggests a much better implementation with reduced queries.

app/controllers/concerns/membership_actions.rb
deepsidhu85: Sure! Updated in 0379505201abd0d8b0a317f46c1819e1c82f5c3e

app/models/member.rb
deepsidhu85: Updated in 0379505201abd0d8b0a317f46c1819e1c82f5c3e

app/policies/member_policy.rb
deepsidhu85: Thanks for helping me clean up the queries! Updated in 0379505201abd0d8b0a317f46c1819e1c82f5c3e

app/models/member.rb
deepsidhu85: Updated in 0379505201abd0d8b0a317f46c1819e1c82f5c3e

app/models/member.rb
deepsidhu85: Updated in 0379505201abd0d8b0a317f46c1819e1c82f5c3e

app/models/member.rb
deepsidhu85: Updated in 0379505201abd0d8b0a317f46c1819e1c82f5c3e

app/views/groups/members/index.html.erb
deepsidhu85: Updated in 0379505201abd0d8b0a317f46c1819e1c82f5c3e

app/views/projects/members/index.html.erb
deepsidhu85: Updated in 0379505201abd0d8b0a317f46c1819e1c82f5c3e

app/policies/member_policy.rb
```diff
@@ -0,0 +1,19 @@
...
+  scope_for :relation do |relation, options|
+    namespace = options[:namespace]
+
+    namespace_ids = if namespace.project_namespace?
+                      [namespace.id] + namespace.parent&.self_and_ancestor_ids
+                    else
+                      namespace.self_and_ancestor_ids
+                    end
+
+    relation.where(id: Member.includes(namespace: [:route]).where(namespace_id: namespace_ids)
```
ericenns, line 14: **suggestion (blocking):** This morning I was able to come up with a much better query than I had previously suggested. The following query does not use `uniq` or `pluck` which means that it can all happen in sql at once without having to perform the inner query before executing the outer query.

```ruby
relation.where(id: Member.joins(namespace: [:route]).where(namespace_id: namespace_ids)
                                                        .order(user_id: :asc, 'routes.path': :desc)
                                                        .select('DISTINCT ON (user_id) members.id'))
```

ericenns (changes requested):
This is so close to being merged, I spent a bit of time this morning trying to find a more optimal query here and was able to come up with a solution. See solution directly in the comments.

ericenns (approved):
Looks good to me! Thanks for adding this and making all the requested changes

PR #124, opened 2023-06-21
Member leave group

### LINKED ISSUE #95: Update text on members page for member removal

deepsidhu85:

This PR adds functionality for a member to be able to leave a group or project on their own without needing an owner or maintainer to remove them.

Due to a bad rebase and force update of commits PR #114 ended up with 0 commits and was closed. Opened this PR to replace it.
Resolves #95 and #96 

**Logged in as user that is removing themselves from the group:**

image

image

**Removing a member that is not the current logged in user**

image

image

1. Login as `admin@example.com`
3. Click the ellipsis for the `admin@example.com` user and you should see a link `Leave group`
4. In the confirmation you should see `Are you sure you want to remove yourself from the group?`
5. Try step 3 for another user in the listing and you should see a link `Remove member`
6. In the confirmation you should see `Are you sure you want to remove this member from the group?`
7. Try out the same steps but for a project. Projects from the seed data only have user6 through user8 as members so you will need to log in as one of those users.

## PR acceptance checklist

### DIFF
--- a/app/views/groups/members/_current_user_is_member.html.erb
+++ b/app/views/groups/members/_current_user_is_member.html.erb
@@ -1,0 +1,16 @@
+<%= viral_dropdown(icon: "ellipsis_vertical",
+                                     classes: "member-settings-ellipsis",
+                                     aria: { label: t(:'groups.members.index.actions.member_dropdown_aria_label',
+                                     member: member.user.email) }) do |dropdown| %>
+  <%= dropdown.item(
+    label: (t(:"groups.members.index.leave_group")),
+    url: group_member_path(@namespace, member),
+    data: {
+      turbo_method: :delete,
+      turbo_confirm: (t(:"groups.members.index.leave_confirmation"))
+    },
+    aria: {
+      label: (t(:"groups.members.index.actions.link_leave_group_aria_label"))
+    }
+  ) %>
+<% end %>
--- a/app/views/groups/members/_current_user_is_member.html.erb
+++ b/app/views/groups/members/_current_user_is_member.html.erb
@@ -1,5 +1,4 @@
 <%= viral_dropdown(icon: "ellipsis_vertical",
-                                     classes: "member-settings-ellipsis",
                                      aria: { label: t(:'groups.members.index.actions.member_dropdown_aria_label',
                                      member: member.user.email) }) do |dropdown| %>
   <%= dropdown.item(
--- a/app/views/groups/members/index.html.erb
+++ b/app/views/groups/members/index.html.erb
@@ -108,24 +108,43 @@
                     <% end %>
                   </div>
                 </td>
-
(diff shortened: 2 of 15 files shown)


app/views/groups/members/_current_user_is_member.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= viral_dropdown(icon: "ellipsis_vertical",
+                                     classes: "member-settings-ellipsis",
```
joshsadam, line 2: **question:** Why do you need this class?

joshsadam (commented):
This looks really good, just one question.

app/views/groups/members/_current_user_is_member.html.erb
deepsidhu85: This was added pretty much at the beginning of when the members pages were developed. It was used to get the ellipsis in the tests. However we no longer use the user defined classes/ids to find stuff on the page. I've removed them in 09df5454052849947625f43e56f308b9b188e427

app/views/groups/members/index.html.erb
```diff
@@ -82,27 +82,12 @@
...
-                        },
-                        aria: {
-                          label:
-                            t(
-                              :"groups.members.index.actions.link_remove_aria_label",
-                              member: member.user.email
-                            )
-                        }
-                      ) %>
+                    <% if current_user == member.user %>
```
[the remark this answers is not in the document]
deepsidhu85: Thanks!

ericenns (approved):
Looks good to me!

PR #138, opened 2023-07-21
Namespace to group controller service

deepsidhu85:

This pull request adds in functionality (controller and service) to allow sharing of a namespace with a group and to unshare a previously shared namespace with a group. This will allow us to share namespaces with other groups with a set max role rather than adding in individual users from other groups if we want to allow access to the namespace.

**Note: Currently the controller actions for share and unshare do a redirect but will be updated to use turbo stream in the UI branch**

1. Open up rails console from the terminal with `bin/rails c`
2. Find a bunch of group and project ids that you will use to link groups to groups and projects with groups
3. Try and create some `NamespaceGroupLink` objects using the `Namespaces::GroupShareService` linking groups with other groups and projects with groups using the ids from step 2. Make sure to try out duplicate entries, entries where the group id doesn't exist in the `namespaces` table
4. Try unsharing the links from step 3 using the `Namespaces::GroupUnhareService`. Make sure to also try out unsharing links that don't exist in the table
5. Ensure the controller tests are working correctly for sharing/unsharing by reviewing the tests

## PR acceptance checklist

- [x] I have evaluated
  the [PR acceptance checklist](https://phac-nml.github.io/irida-next/docs/development/development_processes/code_review#acceptance-checklist) for this PR.


app/models/member.rb
ericenns, line 1: **nitpick (blocking):** The additional methods here, specifically `can_share_group?` and `can_unshare_group?` don't make sense to be when we our `object_namespace` can be a `Group` or a `Project` based on the two places it is called from. 

I see either two solutions:
1. come up with a name that applies to both, but also ensures that `object_namespace` is a `Group` or a `Project`. Things I thought of were `can_share_with_group?` and `can_unshare_with_group?`. (The later one maybe not being that great). Another option is just having it be `can_share?` and `can_unshare?` (not sure if that is too vague though.
2. Add `can_share_project?` and `can_unshare_project?` methods that check if `object_namespace` is a `Project` and the existing one that it is a `Group`.

app/services/namespaces/group_unshare_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      @group_id = group_id
+      @namespace = namespace
+    end
+
+    def execute
+      authorize! namespace, to: :unshare_namespace_with_group?
+
+      namespace_group_link = NamespaceGroupLink.find_by(group_id:, namespace:)
+
+      raise NamespaceGroupUnshareError, 'Namespace to group link does not exist' if namespace_group_link.nil?
```
ericenns, line 20: **nitpick (blocking):** This message should be translated as you are adding it to the errors below.

ericenns (changes requested):
This is looking pretty good, test coverage also looks good. I do have some comments on some of the implementation details though.

app/services/namespaces/group_unshare_service.rb
deepsidhu85: Added translation in 66aaa75edf6c80d24a88cb6abfc8a07a4e4d90a6

app/models/member.rb
deepsidhu85: I decided to come up with a common name as it prevents us from writing some duplicate code. Updated in 66aaa75edf6c80d24a88cb6abfc8a07a4e4d90a6. Also, looking at this further the can_modify method that is called only checks for permissions in the project_namespace and group_namespace, so I've left them out from the actual methods checking if the user can share/unshare the namespace with a group. I have however added a check to the `group_share_service.rb` to check if the namespace is of type group or project and if it is not an error will be raised. Added in 879165c8603555cbdda71701e6b905aec80259b5

ericenns (approved):
Looks good to me! Thanks again for doing this

PR #176, opened 2023-09-11
Nextflow schema UI

joshsadam:

This PR only create a nextflow schema UI component and preview.  **It is not used in the current UI**

Only adds UI components for:

- Text input (text, number, email, etc..)
- Checkboxes

Future PRs will cover select fields for the enum values of the nextflow format.

image

image

Copy the nextflow schema test files into `tmp`/ (https://github.com/nextflow-io/nf-validation/blob/master/plugins/nf-validation/src/testResources/nextflow_schema.json)

View the dynamically created form at http://localhost:3000/rails/lookbook/inspect/nextflow/default?_display=%257B%2522theme%2522%253A%2522dark%2522%257D

## PR acceptance checklist


app/components/nextflow_component.html.erb
```diff
@@ -0,0 +1,57 @@
...
+</h1>
+<p class="mb-4 text-lg font-normal text-gray-500 lg:text-xl dark:text-gray-400">
+  <%= schema["description"] %>
+</p>
+<form>
+  <% schema["definitions"].each do |_item, definition| %>
+    <%- if definition["properties"]
+      .values
+      .filter { |property| property.key?("hidden") }
+      .length != definition["properties"].values.length %>
```
ericenns, line 24: **nitpick (blocking):** I think this if statement should only trigger when the length of non hidden values is > 0 instead

joshsadam: I disagree. If it triggers at > 0 that means if there is 1 out of 5 fields hidden then the whole thing would be hidden - which we do not want.

ericenns: I said non hidden values > 0, which would mean that if 1 out of 5 fields is hidden then it would still show the section with the non hidden value.

ericenns: In this case, I guess what you have achieves the same thing that I was asking for, I just interpreted the code incorrectly.

joshsadam: Updated in 9991fcdcfc718d2f6cafb3823f44aa9e31e63210

ericenns (approved):

PR #178, opened 2023-09-13
Namespace group link UI create

deepsidhu85:
## What does this PR do and why?
_Describe in detail what your merge request does and why._

This PR adds in the UI functionality to create a namespace group link for both groups and projects.

## Screenshots or screen recordings
_Screenshots are required for UI changes, and strongly recommended for all other pull requests._

image

image

image

image

## How to set up and validate locally
_Numbered steps to set up and validate the change are strongly suggested._

1. Login as any user that has groups and projects
2. Go the the members page for a group
3. You should see an `Invite group` button and 2 tabs (Members, Groups)
4. Click the `Invite group` button and select a group to share the namespace with, the group access level, and optionally select an expiration date (currently no restrictions)
5. Click the `Share` button
6. Click the `Groups` tab if you have not already done so
7. The new namespace group link should be listed. Note by default 5 namespace group links are seeded for each group and project. Ensure that the source is direct member for this new namespace group link and that the access level and optional expiration are also listed correctly

## PR acceptance checklist
This checklist encourages us to confirm any changes have been analyzed to reduce risks in quality, performance, reliability, security, and maintainability.

- [X] I have evaluated the [PR acceptance checklist](https://phac-nml.github.io/irida-next/docs/development/development_processes/code_review#acceptance-checklist) for this PR.

PR #179, opened 2023-09-13
Namespace group link UI destroy

deepsidhu85:

This PR add in the UI functionality to delete a namespace group link for both groups and projects.

Merge after #178 

image

image

image

1. Login as any user that groups and projects underneath these groups
2. Go to a group's members page and click on the `Groups` tab
3. Click on the `ellipsis` and then click `Unlink` for a direct member namespace group link. You should only be able to unlink direct members
4. Click `Confirm` in the confirmation window
5. The namespace group link should be deleted and you should see a flash message indicating successful deletion
6. Repeat with a project

## PR acceptance checklist

### DIFF
--- a/app/views/groups/group_links/destroy.turbo_stream.erb
+++ b/app/views/groups/group_links/destroy.turbo_stream.erb
@@ -5,25 +5,5 @@
 <% if @namespace_group_link&.deleted? %>
   <%= turbo_stream.update "members", partial: "invited_groups" %>
 
-  <%= turbo_stream.update "members_pagination" do %>
-    <%= render PaginationComponent.new(
-      info: pagy_info(@pagy),
-      prev_url:
-        (
-          if @pagy.prev
-            group_group_links_url(@namespace, page: @pagy.prev)
-          else
-            nil
-          end
-        ),
-      next_url:
-        (
-          if @pagy.next
-            group_group_links_url(@namespace, page: @pagy.next)
-          else
-            nil
-          end
-        )
-    ) %>
-  <% end %>
+  <%= turbo_stream.update "members_pagination", partial: "pagination" %>
 <% end %>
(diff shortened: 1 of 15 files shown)


joshsadam:
I am still not sure about the inconsistency between the alert used for error and the toast used for success.  What do you think @ericenns

app/views/groups/group_links/destroy.turbo_stream.erb
```diff
@@ -0,0 +1,27 @@
+<%= turbo_stream.append "flashes" do %>
+  <%= viral_flash(type:, data: message) %>
+<% end %>
+
+<%= turbo_stream.update "members", partial: "invited_groups" %>
+
+<%= turbo_stream.update "members_pagination" do %>
+  <%= render PaginationComponent.new(
```
joshsadam, line 8: **suggestion:** This pagination is the same as that is in `create.turbo_stream.erb` and `index.turbo_stream.erb` why don't you just make a partial so you don't need to repeat all that?

ericenns:
> I am still not sure about the inconsistency between the alert used for error and the toast used for success.  What do you think @ericenns.  We also use an alert for the successful addition of a member so there is no consistency here.  I can update those on a different branch as I am sure there are many other places we should update, I just want to know what the two of you think is the best way to handle these/

I'm not sure, I agree that we do not have great consistency. I think I would prefer the toast to be used, we can always change the duration on error so that they show longer. We were using alerts before as we had not yet figured out how to use them with the `turbo-stream` requests, but now we know how to do that. I agree that the other instances should be handled in a separate branch.

app/views/groups/group_links/destroy.turbo_stream.erb
deepsidhu85: Since this is the case for the group_links and members I will do this in a separate pr

joshsadam: Sounds good

joshsadam (approved):
This looks good to me

ericenns (approved):
Looks good to me!

PR #189, opened 2023-09-15
Fix edit project form errors

ksierks:
## What does this PR do and why?
Updated the `Project name and description` form to display errors. Sidebar & breadcrumbs are out of scope.
Fixes #172.

## Screenshots or screen recordings
Screenshot from 2023-09-15 13-02-08
Screenshot from 2023-09-15 13-01-43

## How to set up and validate locally
1. Navigate to a Project
2. Click the `Settings` link in the sidebar
3. In the `Project name and description` form set `Project Name` to `a`
4. Click `Update project`
5. See that the page re-renders with `a` now showing in the sidebar instead of the original project name
6. Reload the page and see that the original project name now displays instead of `a`

## PR acceptance checklist
This checklist encourages us to confirm any changes have been analyzed to reduce risks in quality, performance, reliability, security, and maintainability.

- [x] I have evaluated the [PR acceptance checklist](https://phac-nml.github.io/irida-next/docs/development/development_processes/code_review#acceptance-checklist) for this PR.

PR #197, opened 2023-09-20
Update sidebar & breadcrumbs when editing project

ksierks:
Updates the project name in the layout sidebar and breadcrumbs when a project name is successfully updated.
Pulled from [#189](https://github.com/phac-nml/irida-next/pull/189).

Screenshot from 2023-09-21 16-14-36

1. Navigate to a Project
2. Click the `Settings` link in the sidebar
3. In the `Project name and description` form set `Project Name` to `a`
4. Click `Update project`
5. See that the form displays an error message and the sidebar and breadcrumbs are not updated.
6. Change the `Project Name` to `New project name`.
7. Click `Update project`
8. See a success message and that the sidebar an breadcrumbs have changed to the new name.

Note: Verify that change path refreshes the page to the new URL as expected.

## PR acceptance checklist


deepsidhu85:
Just leaving a general comment on the PR of the changes we discussed:

1) Use html sytnax to create a turbo frame around the breadcrumbs so we don't need to use a helper in the component

2) After setting the `change_path submit button to not use turbo when the path was updated but there was an error it caused the url to change with the `/-/edit` removed from it. Update this to use turbo and the update method in projects_controller to only respond to turbo_stream

3) Since the breadcrumbs are now in a turbo frame the navigation needs to be updated to it happens "outside" of the turbo frame. This can be updated in the breadcrumb component where the link is created to set the data: {turbo_frame: "_top"} for each of the link_to tags

4) Looks like I accidentally deleted the review I had left, but it was just to updated the name of the form_fields partial to a clearer name

joshsadam (approved):
Thanks @ksierks this appears to work the way it is supposed to.  Once @ deepsidhu85 gives it the thumbs up, this is good to go.  We still may want to break each section of the settings page out into its own turbo frame at some point, so you don't need to return it all.

ksierks:
> Just leaving a general comment on the PR of the changes we discussed and that were made:
> 
> 1. Use html sytnax to create a turbo frame around the breadcrumbs so we don't need to use a helper in the component
> 2. After setting the `change_path submit button to not use turbo when the path was updated but there was an error it caused the url to change with the `/-/edit` removed from it. Update this to use turbo and the update method in projects_controller to only respond to turbo_stream
> 3. Since the breadcrumbs are now in a turbo frame the navigation needs to be updated to it happens "outside" of the turbo frame. This can be updated in the breadcrumb component where the link is created to set the data: {turbo_frame: "_top"} for each of the link_to tags
> 4. Looks like I accidentally deleted the review I had left, but it was just to updated the name of the form_fields partial to a clearer name

To address #4, I renamed the form `_edit_name_and_description_form.html.erb` like in the groups PR. Please see aaea28541ed40a2e45c9937e8e0a812428e6cb2d.

test/system/projects_test.rb
```diff
@@ -91,8 +91,20 @@ def setup
     fill_in I18n.t(:'activerecord.attributes.namespaces/project_namespace.name'), with: project_name
```
deepsidhu85, line 91: Can I get you to add a test for a successful path change? We will want to test if the url updated correctly and assert that the success text was displayed in the flash

deepsidhu85 (changes requested):
This all looks great. Just a test to add below

test/system/projects_test.rb
deepsidhu85: You are correct! Totally missed it. Thanks

deepsidhu85 (approved):

PR #204, opened 2023-09-22
Refactor breadcrumbs

ChrisHuynh333:
This PR refactors the breadcrumbs component so that the component no longer requires a ```Route``` object passed as an argument, only the ```context_crumbs```. A ```RouteHelper``` is now available to breakdown ```Routes``` into the ```context_crumbs``` format prior to being passed as the argument when initializing a ```Breadcrumb component```. All existing breadcrumbs should still have the same functionality as prior to this PR.

1. Pull down this branch
2. Clickthrough the groups and projects pages, ensuring breadcrumb names and links are accurate.
3. Compare with breadcrumb names and links on ```main```.

## PR acceptance checklist

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -148,7 +148,8 @@ class GroupsController < Groups::ApplicationController # rubocop:disable Metrics
   end
 
   def context_crumbs
-    @context_crumbs =
+    @context_crumbs = route_to_context_crumbs(@group.route)
+    @context_crumbs +=
       case action_name
       when 'show'
         [{
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -149,19 +149,13 @@ class GroupsController < Groups::ApplicationController # rubocop:disable Metrics
 
   def context_crumbs
     @context_crumbs = route_to_context_crumbs(@group.route)
-    @context_crumbs +=
-      case action_name
-      when 'show'
-        [{
-          name: @group.name,
-          path: group_path
-        }]
-      else
-        [{
-          name: I18n.t('groups.edit.title'),
-          path: group_path
-        }]
-      end
+
+    return unless action_name != 'show'
+
+    @context_crumbs += [{
+      name: I18n.t('groups.edit.title'),
+      path: group_path
+    }]
   end
 
   protected
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -2,6 +2,7 @@
 
 # Controller actions for Groups
 class GroupsController < Groups::ApplicationController # rubocop:disable Metrics/ClassLength
+  include BreadcrumbNavigation
   layout :resolve_layout
   before_action :group, only: %i[edit show destroy update transfer]
   before_action :authorized_namespaces, except: %i[index show destroy]
@@ -31,7 +32,6 @@ class GroupsController < Groups::ApplicationController # rubocop:disable Metrics
 
   def new
     @group = Group.find(params[:parent_id]) if params[:parent_id]
-    context_crumbs
     authorize! @group, to: :create_subgroup? if params[:parent_id]
 
     @new_group = Group.new(parent_id: @group&.id)
@@ -51,7 +51,6 @@ class GroupsController < Groups::ApplicationController # rubocop:disable Metrics
       redirect_to group_path(@new_group.full_path)
     else
       @group = @new_group.parent
-      context_crumbs
       render_new status: :unprocessable_entity
     end
   end
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -2,7 +2,6 @@
 
 # Controller actions for Groups
 class GroupsController < Groups::ApplicationController # rubocop:disable Metrics/ClassLength
-  include BreadcrumbNavigation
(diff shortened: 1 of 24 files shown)


ChrisHuynh333:
I am not entirely happy with my solution for ```groups_controller``` ```new``` and ```create```, as well as ```projects_controller``` ```update``` where I've hard coded in a call to update ```context_crumbs```. The reason for the hard coded ```context_crumbs``` call (rather than a ```before_action```) is the respective ```@group``` or ```@project``` are updated during the actions, and therefore a before_action does not work.

I could see moving ```groups_controller#new```  ```@group = Group.find(params[:parent_id]) if params[:parent_id]``` into the ```def group``` function with an ```if/else``` statement based on the action, but I'm not sure about the other two.

Would love to hear any suggestions if you guys have a better way to solve this, unless I'm overthinking this and the solution is fine as is.

ericenns:
> I am not entirely happy with my solution for `groups_controller` `new` and `create`, as well as `projects_controller` `update` where I've hard coded in a call to update `context_crumbs`. The reason for the hard coded `context_crumbs` call (rather than a `before_action`) is the respective `@group` or `@project` are updated during the actions, and therefore a before_action does not work.
> 
> I could see moving `groups_controller#new` `@group = Group.find(params[:parent_id]) if params[:parent_id]` into the `def group` function with an `if/else` statement based on the action, but I'm not sure about the other two.
> 
> Would love to hear any suggestions if you guys have a better way to solve this, unless I'm overthinking this and the solution is fine as is.

I agree, it took me a bit of research but I came up with the following solution.

Create a Controller Concern called `BreadcrumbNavigation`, within this define a `render` method with the following method signature/content:

```ruby
  def render(*args)
    context_crumbs
    super
  end
```

That will automatically call the `context_crumbs` method just before rendering the view, which would mean the variables you need should be setup.

app/controllers/projects/samples/attachments_controller.rb
```diff
@@ -4,7 +4,6 @@ module Projects
   module Samples
     # Controller actions for Project Samples Attachments
     class AttachmentsController < ApplicationController
```
ericenns, line 6: **nitpick (blocking):** Since you removed the before action here, I believe you need to update this to inherit from `Projects::ApplicationController` instead of just `ApplicationController`

app/controllers/projects/samples/transfers_controller.rb
```diff
@@ -4,7 +4,6 @@ module Projects
   module Samples
     # Controller actions for Project Samples Transfer
     class TransfersController < ApplicationController
```
ericenns, line 6: **nitpick (blocking):** Since you removed the before action here, I believe you need to update this to inherit from `Projects::ApplicationController` instead of just `ApplicationController`

ericenns (changes requested):
Looking good, but I did find a few small issues that need to fixed.

app/controllers/groups_controller.rb
```diff
@@ -4,8 +4,8 @@
 class GroupsController < Groups::ApplicationController # rubocop:disable Metrics/ClassLength
   layout :resolve_layout
   before_action :group, only: %i[edit show destroy update transfer]
-  before_action :context_crumbs, except: %i[index new create show]
   before_action :authorized_namespaces, except: %i[index show destroy]
+  before_action :context_crumbs, except: %i[index new create]
```
ericenns, line 8: **nitpick (blocking):** This before_action is not required anymore with your refactor

ericenns (changes requested):
I found another small issue and also noticed that the lookbook previews are broken e.g. http://localhost:3000/rails/lookbook/inspect/viral_breadcrumb/default and http://localhost:3000/rails/lookbook/inspect/viral_breadcrumb/with_crumbs

app/controllers/groups_controller.rb
```diff
@@ -2,10 +2,11 @@
 
 # Controller actions for Groups
 class GroupsController < Groups::ApplicationController # rubocop:disable Metrics/ClassLength
+  include BreadcrumbNavigation
```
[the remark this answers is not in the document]
ChrisHuynh333: This was fixed in 7f856988fa31564ae415cb0b75ed55eac7adeddc

app/controllers/projects/samples/attachments_controller.rb
ChrisHuynh333: This was fixed in 7f856988fa31564ae415cb0b75ed55eac7adeddc

app/controllers/groups_controller.rb
ChrisHuynh333: The ```before_action``` was removed in both ```groups``` and ```projects``` controllers in 3ddabf0bdb2ffa66ba1cda3f0446ce4e21177ce8

ChrisHuynh333:
> I found another small issue and also noticed that the lookbook previews are broken e.g. http://localhost:3000/rails/lookbook/inspect/viral_breadcrumb/default and http://localhost:3000/rails/lookbook/inspect/viral_breadcrumb/with_crumbs

I've fixed the breadcrumbs in lookbook in 7ec469b3401110aba372f46e8474ff6ee266e3e3

ericenns (approved):
This looks good to me!

deepsidhu85 (approved):
This looks good to me! Tested it out and works as expected!

PR #216, opened 2023-09-26
Refactor distance of time

ChrisHuynh333:
This PR is refactoring the use of the rails function ```distance_of_time_in_words``` into a helper function ```time_ago``` to include an html title attribute of the date the action was performed (created/updated date, etc.), so that the actual date is available for the user upon hovering. The ```time_ago``` helper uses an existing ```viral tooltip``` component and can now be called in the views pages for when ```distance_of_time_in_words``` would be used so that all views will contain this date on hover functionality.

image

1. Check the following pages and hover over the time ago element to check if the tooltip contains the expected date. Create the associated element if needed (group, project, members, etc.).
- Projects and Groups: Dashboard, Group Links, Members
- Sample Attachments

## PR acceptance checklist


joshsadam (changes requested):
This is a really good start.  I would like to see the following changes:

1.  Instead of a partial, can you update this to a view component, I think it could even be a viral component.  Using view components is more performant than partials and allows you to display previews in lookbook and testing.
2.  In the viral tooltip component, can you change the surrounding `div` to a \` span\`.  The `div` causes a line break on the tables putting the "by" word on it's own line, really because the `div` is a block element.  I created the component originally and did not think of that.

ChrisHuynh333:
> This is a really good start.  I would like to see the following changes:
> 
> 1. Instead of a partial, can you update this to a view component, I think it could even be a viral component.  Using view components is more performant than partials and allows you to display previews in lookbook and testing.
> 2. In the viral tooltip component, can you change the surrounding `div` to a ` span`.  The `div` causes a line break on the tables putting the "by" word on it's own line, really because the `div` is a block element.  I created the component originally and did not think of that.
> 
> 

@joshsadam I've changed it from a partial to a viral component and changed the tooltip div into a span in 85b83c2574b39edd113d5ef9dde7d20dff907c4b

app/components/viral/distance_of_time_component.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Viral
+  # Card component for rendering sections of pages.
+  class DistanceOfTimeComponent < Viral::Component
+    attr_reader :time_difference, :original_time
```
joshsadam, line 6: **suggestion (blocking):** Since this is a time ago component it probably does not need to take in the current time, looking through your code you are always passing in `Time.now`.  What I think you should do is set that as the default current_time and if someone wants to override it they can but at least they would not need to pass it in.

app/helpers/view_helper.rb
```diff
@@ -11,6 +11,7 @@ module ViewHelper
     checkbox: 'Viral::Form::CheckboxComponent',
     datepicker: 'Viral::Form::DatepickerComponent',
     dialog: 'Viral::DialogComponent',
+    distance: 'Viral::DistanceOfTimeComponent',
```
joshsadam, line 14: **thought:** I don't know if this is an appropriate name, as it actually implies to me something like the distance to an object.  Can we call it `time_ago` or something similar?  I hate naming things.

joshsadam (changes requested):
Good job updating it to a view component.  Just a couple of suggestions here.  I also need you to add a LookBook preview and tests.  Start by adding a preview in: `test/components/previews`by copying another one like `viral_tooltip_component_preview` , and then for tests do something like what is in `test/components/viral/form/datepicker_component_test.rb`, I ned to update some of the other ones as I did not set them up 100% properly.

app/helpers/view_helper.rb
ChrisHuynh333: Completed in 8a2cbfabe8f262620fcade7c5ea1b47dcc8b332a

app/components/viral/distance_of_time_component.rb
ChrisHuynh333: Completed in ebf6676022ee1938aa3ba12f06ef58ba32d369ae

ChrisHuynh333:
> Good job updating it to a view component.  Just a couple of suggestions here.  I also need you to add a LookBook preview and tests.  Start by adding a preview in: `test/components/previews`by copying another one like `viral_tooltip_component_preview` , and then for tests do something like what is in `test/components/viral/form/datepicker_component_test.rb`, I ned to update some of the other ones as I did not set them up 100% properly.

Previews are added in 2d0ebb4f7a18e83d0a955ab12faa14d6b7f26095
Tests are added in d08f2957ccc5fee26dca667c20948cdc88bea133

test/components/viral/system/time_ago_component_test.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'application_system_test_case'
+
+module System
+  class TimeAgoComponentTest < ApplicationSystemTestCase
+    test 'default' do
+      visit('/rails/view_components/viral_time_ago_component/default')
```
joshsadam, line 8: **nitpick (blocking):** 
Can you make this into 2 tests:
1. The `tooltip` test is good (rename to `default`), but start it by making sure the `15 days ago` is there, then go into testing the tooltip.
2. Make a second test `including current time` to test when the user wants to pass in a custom current time (although I still don't know when that would ever occur).  Then test the same way you did the first one including the value and tooltip.

ChrisHuynh333: Fixed in ebaa3e31e69d1e06fdfd0d9fcc6b28975e70ae59

joshsadam (approved):
This looks great now, will merge once tests pass.

PR #237, opened 2023-10-12
Sign in enhancements

### LINKED ISSUE #190: Sign in Enhancements
## Problem
Local accounts are given priority on the sign in page when other login sources are likely to be the default.

## Solution
Update the sign in page to look like below:

1. When no omniauth providers:

image

2. When omniauth providers:

image

When clicking `Sign in with Local Account` the hidden sign in form appears and the other buttons disappear. e.g.

image

When clicking `Return to sign in selection` the form is hidden and the sign in buttons are visible again

ksierks:
The email & password login form is now being displayed after clicking the local account button when omni auth providers are specified.
Fixes #190.

Screenshot from 2023-10-12 16-13-11
Screenshot from 2023-10-12 16-11-33
Screenshot from 2023-10-12 16-12-21

Test with no omni auth providers:
1. In a terminal run `bin/dev`.
2. In a browser navigate to `http://localhost:3000/.
3. Verify only email & password login is displayed.

Test with omni auth providers:
1. In a terminal run `OMNIAUTH_PROVIDERS="developer,saml,azure_activedirectory_v2" bin/dev`.
2. In a browser navigate to `http://localhost:3000/.
3. Verify the selection page is displayed.
4. Click the `Sign in with Local Account` button.
5. Verify the page refreshes with email & password form login displayed.
6. Click the `Return to sign in selection` button.
7. Verify the selection page is displayed.

## PR acceptance checklist


app/views/devise/sessions/new_with_providers.html.erb
```diff
@@ -0,0 +1,35 @@
+<% if @local_account %>
+  <%= render "devise/shared/form" %>
+<% end %>
+<div class="grid gap-2">
+  <% if @local_account %>
+    <%= render "devise/shared/links" %>
+    <%= button_to t(".return_button"),
+    new_user_session_path,
+    method: :get,
+    params: {
```
deepsidhu85, line 10: We don't need to use this params: {}. We can just add the `local: true` to the `new_user_session_path` above

app/views/devise/sessions/new_with_providers.html.erb
```diff
@@ -0,0 +1,35 @@
...
+        <div class="inline-flex items-center text-center">
+          <%= viral_icon(name: provider, classes: "w-4 h-4 mr-2 -ml-1") %>
+          <%= t(".omniauth", provider: OmniAuth::Utils.camelize(provider)) %>
+        </div>
+      <% end %>
+    <% end %>
+    <%= button_to t(".local_button"),
+    new_user_session_path,
+    method: :get,
+    params: {
```
deepsidhu85, line 30: Same comment as above about not needing to use the params: {}

app/controllers/sessions_controller.rb
```diff
@@ -7,9 +7,18 @@ class SessionsController < Devise::SessionsController
...
-  # end
+  def new
+    self.resource = resource_class.new(sign_in_params)
+    clean_up_passwords(resource)
+    yield resource if block_given?
+
+    if resource_class.omniauth_providers.empty?
+      render :new_with_no_providers
+    else
+      @local_account = params[:local] == 'true'
```
deepsidhu85, line 18: We can just set local_account to params[:local]. Don't need to check if it's true, as we will only be passing in the local param if the link is for a local account

deepsidhu85 (changes requested):
This is looking good! Just a few small comments below

app/views/devise/sessions/new_with_providers.html.erb
deepsidhu85: Disregard and just leave the params in as this would require a route change

app/views/devise/sessions/new_with_providers.html.erb
deepsidhu85: Disregard and just leave the params in as this would require a route change

app/controllers/sessions_controller.rb
ksierks: I was weird and set the return button param to false. Fixed in fd8b1e3626721a48ecc73c1ae3717d6d6ed686ff, but I might have introduced an error message now. Does this happen for you too?
Screenshot from 2023-10-12 14-08-23

app/views/devise/sessions/new_with_providers.html.erb
ksierks: As discussed, please see 2817030b69fb2ca9cedc353957f627a8c9f57a51.

app/views/devise/sessions/new_with_providers.html.erb
ksierks: As discussed, please see 2817030b69fb2ca9cedc353957f627a8c9f57a51.

app/views/devise/confirmations/new.html.erb
```diff
@@ -4,13 +4,25 @@
   <%= render "devise/shared/error_messages", resource: resource %>
 
   <div class="field">
-    <%= f.label :email %><br />
-    <%= f.email_field :email, autofocus: true, autocomplete: "email", value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) %>
+    <%= f.label :email %><br/>
+    <%= f.email_field :email,
```
deepsidhu85, line 8: Since we are doing enhancements to the sign in page and have touched the devise templates, can I get you to add translations for the labels, buttons (submit, etc;), and any other elements that are missing translations, to the templates under `passwords/`, `registrations/`, `unlocks/`

app/views/devise/shared/_links.html.erb
```diff
@@ -1,71 +1,37 @@
...
-                      turbo: false
-                    },
-                    class: "button button--state-default button--size-default w-full" do %>
-        <div class="inline-flex items-center text-center">
-          <%= viral_icon(name: provider, classes: "w-4 h-4 mr-2 -ml-1") %>
-          <%= OmniAuth::Utils.camelize(provider) %>
-        </div>
-      <% end %>
+<%- if devise_mapping.confirmable? && controller_name != 'confirmations' %>
+  <%= link_to "Didn't receive confirmation instructions?",
```
deepsidhu85, line 26: Update to use a translation here and for the link_to below

deepsidhu85 (changes requested):
Found a few things I missed in my initial review. Just a few comments related to translations below

app/views/devise/confirmations/new.html.erb
ksierks: Assuming I can ignore the mailer templates, please see d79b221721c5aac23e23772fa75a825607a41b53.

deepsidhu85: We may as well add translations for the mailer templates as well. We will need them at some point!

ksierks: Please see f721f8e35db852a440823cf56b9db8deb8ff094e.

app/controllers/sessions_controller.rb
```diff
@@ -7,9 +7,18 @@ class SessionsController < Devise::SessionsController
   # before_action :configure_sign_in_params, only: [:create]
 
   # GET /resource/sign_in
-  # def new
-  #   super
-  # end
+  def new
```
ericenns, line 10: **nitpick (blocking):** could you refactor this to call `super` and pass your render logic in as block, this way we know our controller won't have to be updated if this gets updated in Devise.

e.g.
```ruby
def new
  super do
     # custom view logic
     return
  end
end
```

The `yield resource if block_given?` is in the devise controller so that you can do the above.

app/views/devise/unlocks/new.html.erb
```diff
@@ -1,16 +1,18 @@
-<h2>Resend unlock instructions</h2>
+<h2><%= link_to t(".title") %></h2>
```
ericenns, line 1: **nitpick (blocking):** Is their a reason why you are using `link_to` here or was it just a bad copy/paste.

app/views/devise/unlocks/new.html.erb
```diff
@@ -1,16 +1,18 @@
...
 
   <div class="field">
-    <%= f.label :email %><br />
+    <%= f.label :email %><br/>
     <%= f.email_field :email, autofocus: true, autocomplete: "email" %>
   </div>
 
   <div class="actions">
-    <%= f.submit "Resend unlock instructions" %>
+    <%= f.submit link_to t(".submit_button") %>
```
ericenns, line 12: **nitpick (blocking):** Is their a reason why you are using `link_to` here or was it just a bad copy/paste.

ericenns (changes requested):
This is working great, just a few minor changes I would like made.

app/views/devise/unlocks/new.html.erb
ksierks: Yup, lots of translations, bad copy & pasting. Please see 25dbf61e43409bd3e907ae96673d9639764591d9.

app/views/devise/unlocks/new.html.erb
ksierks: Yup, lots of translations, bad copy & pasting. Please see 25dbf61e43409bd3e907ae96673d9639764591d9.

app/controllers/sessions_controller.rb
ksierks: Please see 6fddb0f89573a2c6e2811c66a37cd4ea05ce707f.

Issue #238, opened 2023-10-12
Attachments: Concatenation Service


ericenns:
## Problem
As a User I occasionally have Samples which perform poorly during sequencing. To deal with this poor data, I want to concatenate data from multiple sequencing runs together. To accomplish this currently I would have to download the fastq data separately and perform concatenation myself and then reupload the data.

## Solution
Develop an Attachments::ConcatenationService that accepts multiple single end or sets of paired end files and produces either a single end or a set of paired end files.

## Acceptance Criteria
* Service should accept the Sample that the new concatenated Attachments will belong to
* Service should confirm that the selected Attachments belong to the Sample
* Service should accept either multiple single end or multiple sets of paired end files (no mix)
* Service should produce a single end file or a set of paired end files (depends on input)
* Optionally the service should delete the input files
* Service should accept a basename for the newly concatenated file(s)

PR #243, opened 2023-10-19
Attachments Concatenation Service

deepsidhu85:

This PR adds in the service layer to concatenate single and paired end reads. Single end files can only be concatenated with other single end files, and paired end files can only be concatenated with other paired end files. The service accepts params in the following formats

# single-end files
1) attachment_ids is just a 1 dimensional array and is required
2) basename is the base name for the file
3) delete_originals can be set to true to delete the original files. This is also optional

`params: { attachment_ids: [], basename: "FILE BASENAME", delete_originals: true }`

# paired-end files
1) attachment_ids is just a 2 dimensional array and is required. Each array should be a set of paired-end files
2) basename is the base name for the file
3) delete_originals can be set to true to delete the original files. This is also optional

`params: { attachment_ids: [[],[],...], basename: "FILE BASENAME", delete_originals: true }`

Addresses #238  

N/A

1) Login with any user
2) Create a new project
3) Create a new sample inside this project
4) Upload 2 single end files to sample
5) In the rails console get the attachments for this sample and note their ids
6) Set a variable such as `u` to the user from above
7) Set a variable such as `s` to the sample from above
8) Call `Attachments::ConcatenationService.new(u, s, {attachment_ids: IDS_FROM_STEP_5_IN_ARRAY}).execute`
9) In the browser refresh the page and you should see the newly concatenated file, and the originals as well since we did not set the `delete_originals` flag
10) Remove the concatenated file from the sample and try step 8 again in rails console but this time pass in the `delete_originals: true`
11) Refresh the browser and you should see the newly concatenated file and the originals should be deleted
12) Try out the steps from above using paired-end files
13) Attempt to do the above steps using a `user` which doesn't have permissions to update the sample in a project, you should receive a `Not Authorized` error

## PR acceptance checklist


app/services/attachments/concatenation_service.rb
```diff
@@ -0,0 +1,153 @@
...
+    private
+
+    # Validates attachments to ensure they are of the correct type.
+    # Files can only be concatenated as follows. No mixing of types:
+    # Paired-end -> Paired-end
+    # Single-end -> Single-end
+    def validate_same_types(attachments, is_paired_end, is_single_end) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
+      valid_type = true
+
+      attachments.each do |attachment|
```
ericenns, line 79: **nitpick (blocking):** Reading this code `valid_type` could go from true to false to true. E.g. first attachment set's it to false and second sets it true. Also instead of checking for presence of key we should check the value. I also wonder if we should separate validation of extensions into a separate method to reduce rubocop disables.

app/services/attachments/concatenation_service.rb
```diff
@@ -0,0 +1,153 @@
...
+      raise AttachmentConcatenationError,
+            I18n.t('services.attachments.concatenation.incorrect_file_types')
+    end
+
+    # Concatenates the single end reads into a single-end file
+    def concatenate_single_end_reads(attachments)
+      basename = concatenation_params[:basename] || 'concatenated_file'
+      extension = attachments.first.file.filename.to_s.partition('.').last
+
+      extension = 'fastq.gz' if extension == 'gz'
```
ericenns, line 110: **question (non-blocking):** Do you think that it would help if we added `compression` metadata to `attachment` and have it `none` if no compression and `gzip` if `.gz`?

app/services/attachments/concatenation_service.rb
```diff
@@ -0,0 +1,153 @@
...
+
+    # Concatenates the paired-end reads into a multiple paired-end files
+    def concatenate_paired_end_reads(forward_reads, reverse_reads) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+      basename = concatenation_params[:basename] || 'concatenated_file'
+      extension = forward_reads.first.file.filename.to_s.partition('.').last
+
+      extension = 'fastq.gz' if extension == 'gz'
+
+      files = []
+      if forward_reads.length.positive?
```
ericenns, line 130: **nitpick (blocking):** Due to other comments I think we should remove the `positive?` checks and I think it would also be good to refactor the contents of the `if` and other `if` to be a new method, since this logic is shared between `concatenate_single_end_reads` and this method.

ericenns (changes requested):
This is looking good, but I did find a number of issues and provided suggestions on how to fix.

app/services/attachments/concatenation_service.rb
```diff
@@ -0,0 +1,153 @@
...
+
+    def execute # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
+      # authorize if user can update sample
+      authorize! attachable.project, to: :update_sample? if attachable.instance_of?(Sample)
+
+      attachment_ids = concatenation_params[:attachment_ids]
+
+      unless attachment_ids.all? { |i| i.is_a?(Integer) }
+        # if multi-dimensional array of ids
+        attachment_ids = attachment_ids.first + attachment_ids.last
```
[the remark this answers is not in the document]
deepsidhu85: Updated to allow 2 or more pairs of files to be concatenated in 349d6987f0f58fb6c2481ed054fe66446e6a6090

app/services/attachments/concatenation_service.rb
```diff
@@ -0,0 +1,153 @@
...
+      # single-end params: { attachment_ids = [], basename: basefilename OPTIONAL,
+      #                      delete_originals: true OPTIONAL
+      #                      }
+      # paired-end params: { attachment_ids = [[],[]], basename: basefilename OPTIONAL,
+      #                      delete_originals: true OPTIONAL
+      #                      }
+      @concatenation_params = params
+    end
+
+    def execute # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
```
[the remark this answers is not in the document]
deepsidhu85: Refactored into smaller methods in 349d6987f0f58fb6c2481ed054fe66446e6a6090

app/services/attachments/concatenation_service.rb
```diff
@@ -0,0 +1,153 @@
...
+      # authorize if user can update sample
+      authorize! attachable.project, to: :update_sample? if attachable.instance_of?(Sample)
+
+      attachment_ids = concatenation_params[:attachment_ids]
```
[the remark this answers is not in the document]
deepsidhu85: Added attachment_ids validation before processing in 349d6987f0f58fb6c2481ed054fe66446e6a6090

app/services/attachments/concatenation_service.rb
```diff
@@ -0,0 +1,153 @@
...
+
+      unless attachment_ids.all? { |i| i.is_a?(Integer) }
+        # if multi-dimensional array of ids
+        attachment_ids = attachment_ids.first + attachment_ids.last
+      end
+
+      attachments = attachable.attachments.where(id: attachment_ids, attachable:).order(:id)
+
+      if attachments.length != attachment_ids.length
```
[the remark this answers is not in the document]
deepsidhu85: Yup that is correct. I added a comment to the block of code in 349d6987f0f58fb6c2481ed054fe66446e6a6090

app/services/attachments/concatenation_service.rb
deepsidhu85: Updated this so valid_type is no longer used. Instead we continue going through the attachments unless one of them doesn't match the `if` at which point an error will be raised in 349d6987f0f58fb6c2481ed054fe66446e6a6090

app/services/attachments/concatenation_service.rb
deepsidhu85: Yup definitely would. Will be doing this in a separate commit

app/services/attachments/concatenation_service.rb
deepsidhu85: Refactored the logic to remove the `positive?` checks and moved common code between this method and the one for single end reads into a new method in 349d6987f0f58fb6c2481ed054fe66446e6a6090

app/services/attachments/concatenation_service.rb
deepsidhu85: Added compression metadata to attachments in fc5a524e4a03e80f0bd3b6e0021f4cc842a0acac

app/services/attachments/create_service.rb
```diff
@@ -32,14 +32,15 @@ def execute
 
     private
 
-    def identify_illumina_paired_end_files(attachments) # rubocop:disable Metrics/MethodLength,Metrics/CyclomaticComplexity,Metrics/AbcSize
+    def identify_illumina_paired_end_files(attachments) # rubocop:disable Metrics/MethodLength,Metrics/CyclomaticComplexity,Metrics/AbcSize, Metrics/PerceivedComplexity
```
ericenns, line 35: **nitpick (blocking):** Can you move the `_R1` `_R2` matching into a separate method called `identify_paired_end_files`. And just make them as `type: pe` instead of `illumina_pe` as they may or may not be illumina.

app/services/attachments/concatenation_service.rb
```diff
@@ -0,0 +1,173 @@
...
+    def concatenate_paired_end_reads(attachments) # rubocop:disable Metrics/AbcSize
+      basename = concatenation_params[:basename] || 'concatenated_file'
+
+      extension = attachments.first.metadata['compression'] == 'gzip' ? '.gz' : ''
+
+      forward_reads, reverse_reads = attachments_to_paired_end_directional_reads(attachments)
+
+      files = []
+
+      blobs = retrieve_attachment_blobs(forward_reads)
```
ericenns, line 134: **nitpick (blocking):** I was thinking that the refactoring would be a method like `concatenate_attachments` and accept an array of attachments, and a filename. that way you could do the following:

```ruby
files << concatenate_attachments(forward_reads, "#{basename}_R1.fastq#{extension}").signed_id
```

This method would return the new `ActiveStorage::Blob`

ericenns (changes requested):
A couple more comments that I quickly had.

app/services/attachments/create_service.rb
deepsidhu85: Moved identifying regular pe files into their own method in 5e4fc6115a9cecf5130d295baaf4633dd3e1dcc4

app/services/attachments/concatenation_service.rb
deepsidhu85: Refactored in 5e4fc6115a9cecf5130d295baaf4633dd3e1dcc4

ericenns (approved):
Looks good to me!

PR #251, opened 2023-10-19
Update attachment listings

ChrisHuynh333:
In this PR, metadata fields belonging to attachments are added the sample attachments table to show the additional data (format and type) to users..

image
image

1. Create a project and associated sample.
2. Upload attachments with a variety of formats. Namely a non-fasta/fastq file, a fasta file, and fastq files that are both paired end and non-paired end

## PR acceptance checklist


ChrisHuynh333:
NOTE: paired-end samples only contain a single DELETE button. Therefore, delete functionality is not handled correctly currently. After discussion with @ericenns, deleting of paired end samples will be changed so that the delete function deletes both the paired end attachments. This feature has not been completed.

ChrisHuynh333:
There are, of course, numerous ways to in handling the paired end samples within the table. After a few variations, I settled on the following logic:
- During attachments for loop when populating the table, if the attachment is non-paried end, the row populates as normal. If the attachment has an associated file (Ie: paired end attachments), if the current file is in the forward direction, populate the row with this forward attachments data, and grab the associated reverse attachment data using the forward direction attachment's associated file ID and populate a second row with the reverse data. Otherwise, if the attachment is paired end but in the reverse direction, skip (because the forward paired sample will grab this data).

I decided on this as it keeps the code cleaner than a ton of if else statements throughout `_attachment.html.erb`, as well as I had a bug where attachments did not keep their order by ID, and therefore paired end samples would get split. This iteration of my PR avoids both these issues.

However, a drawback is if somehow the user deletes one of the two paired end samples, if there's belief this can become an issue, I can look into adding more conditionals to prevent possible errors.

app/components/viral/attachment_label_component.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Viral
+  # Viral component for attachments table colored labels
+  class AttachmentLabelComponent < Viral::Component
+    attr_reader :label
+
+    COLORS = {
+      format: {
+        fasta: 'bg-blue-100 text-blue-800 text-base font-medium
```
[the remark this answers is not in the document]
ChrisHuynh333: Okay, my bad. I had thought the font size was a little small, hence the change. I've changed to the pill badge classes. And I assume you meant change the pill to a `span` rather than the `div` that was already there?

Fixed in 4f9fbc4c1f178c4e6fa1e5f394e88f54b37535c2

joshsadam:
I was playing with your branch this morning.  When I upload a pair of files it is displayed twice
image

app/components/viral/pill_component.rb
```diff
@@ -0,0 +1,54 @@
...
+      gray: 'bg-gray-100 text-gray-800 text-xs font-medium mr-2 px-2.5 py-0.5 rounded-full dark:bg-gray-700 dark:text-gray-300',  # rubocop:disable Layout/LineLength
+      red: 'bg-red-100 text-red-800 text-xs font-medium mr-2 px-2.5 py-0.5 rounded-full dark:bg-red-900 dark:text-red-300', # rubocop:disable Layout/LineLength
+      green: 'bg-green-100 text-green-800 text-xs font-medium mr-2 px-2.5 py-0.5 rounded-full dark:bg-green-900 dark:text-green-300', # rubocop:disable Layout/LineLength
+      yellow: 'bg-yellow-100 text-yellow-800 text-xs font-medium mr-2 px-2.5 py-0.5 rounded-full dark:bg-yellow-900 dark:text-yellow-300', # rubocop:disable Layout/LineLength
+      indigo: 'bg-indigo-100 text-indigo-800 text-xs font-medium mr-2 px-2.5 py-0.5 rounded-full dark:bg-indigo-900 dark:text-indigo-300', # rubocop:disable Layout/LineLength
+      purple: 'bg-purple-100 text-purple-800 text-xs font-medium mr-2 px-2.5 py-0.5 rounded-full dark:bg-purple-900 dark:text-purple-300', # rubocop:disable Layout/LineLength
+      pink: 'bg-pink-100 text-pink-800 text-xs font-medium mr-2 px-2.5 py-0.5 rounded-full dark:bg-pink-900 dark:text-pink-300' # rubocop:disable Layout/LineLength
```
joshsadam, line 16: **suggestion:** Can you take out the repetitive class and and them directly down below into the `system_arguments`, that should help alleviate the line length warning and clean the code up a little.

app/components/viral/pill_component.rb
```diff
@@ -0,0 +1,54 @@
...
+      @system_arguments = system_arguments
+      @system_arguments[:classes] = class_names(
+        @system_arguments[:classes],
+        COLORS[@color]
+      )
+    end
+
+    private
+
+    def get_color(text)
```
joshsadam, line 39: **suggestion:** Can you move this logic into the template?  Having it here couples it to files, and we will want to use it in other places.

app/views/projects/samples/attachments/_attachment.html.erb
```diff
@@ -1,19 +1,75 @@
...
+  bg-white
+  border-b
+  dark:bg-slate-800
+  dark:border-slate-700
+  "
 >
+  <% if attachment.associated_attachment %>
+  <th class="px-6 py-4">
+    <div>
+      <div class="flex mb-4">
```
joshsadam, line 13: **nitpick:** Can you add the class `items-center` here and line 26 below, this will center the icon and the text.

app/views/projects/samples/attachments/_attachment.html.erb
```diff
@@ -1,19 +1,75 @@
 <tr
```
joshsadam, line 1: **nitpick:** Can you check your formatters?  This file's indentations are off.  Are you using `erb-formatter`?

joshsadam (changes requested):
Left some small changes I would like to see.  Also noticed an issue when deleting a pair of files, one is not deleted and displayed as a single file.  Then if you try to upload a pair again, it fails with a checksum error, which makes sense since one of the files still exists.  Is this in the scope of the PR @ericenns ?

app/views/projects/samples/attachments/_table.html.erb
```diff
@@ -4,17 +4,21 @@
...
           <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.filename") %></th>
+          <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.format") %></th>
+          <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.type") %></th>
           <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.size") %></th>
           <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.created_at") %></th>
           <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.action") %></th>
         </tr>
       </thead>
       <tbody id="attachments-table-body">
         <% attachments.each do |attachment| %>
```
joshsadam, line 15: **suggestion:** Make this a partial and use it in the create turbo response.  Pass in the persistent attachments to it in the create method on the attachments_controller.  You will have to keep track of which ones are persisted and which ones failed, you can do a filter to find out which ones persisted.

app/views/projects/samples/attachments/_attachment.html.erb
```diff
@@ -1,19 +1,80 @@
-<tr
-  id="<%= dom_id(attachment) %>"
-  class="bg-white border-b dark:bg-slate-800 dark:border-slate-700"
->
+<tr id="<%= dom_id(attachment) %>" class="bg-white border-b dark:bg-slate-800 dark:border-slate-700">
+  <% if attachment.associated_attachment && attachment.metadata['direction'] == 'forward' %>
+  <th class="px-6 py-4">
```
joshsadam, line 3: **nitpick:** These should be `td` not `th`, `th` should only be used within a `thead`

app/components/viral/pill_component.rb
ChrisHuynh333: Fixed in b265302acf47f8d228735e52ddb83bb1b54bfded

app/components/viral/pill_component.rb
ChrisHuynh333: Fixed in b265302acf47f8d228735e52ddb83bb1b54bfded

app/views/projects/samples/attachments/_attachment.html.erb
ChrisHuynh333: Fixed in b265302acf47f8d228735e52ddb83bb1b54bfded

app/views/projects/samples/attachments/_attachment.html.erb
ChrisHuynh333: Fixed in 65b34e490be98d1f85f675ebfd53b0c9c3d11019

app/components/viral/pill_component.erb
```diff
@@ -0,0 +1,3 @@
+<%= render(Viral::BaseComponent.new(tag: 'span', **system_arguments)) do %>
```
joshsadam, line 1: **nitpick:** Can you rename this file to `pill_component.html.erb`

ChrisHuynh333: Good catch. Fixed in 99b2aaa00d86610c8060aca2f5dc2d673736b4ec

app/views/projects/samples/attachments/_attachment.html.erb
```diff
@@ -1,19 +1,80 @@
...
+        ),
+        data: {
+          turbo: false
+        } %>
+      </span>
+      </div>
+    </div>
+  </td>
+  <% else %>
+  <td scope="row" class="px-6 py-4"><%= link_to attachment.file.filename,
```
joshsadam, line 34: **nitpick:** `scope="row"` is deprecated, just remove it.

ChrisHuynh333:
> I was playing with your branch this morning. When I upload a pair of files it is displayed twice image

Yes, I missed a conditional on the create turbo frame. It should be fixed now in b265302acf47f8d228735e52ddb83bb1b54bfded

ChrisHuynh333:
> Left some small changes I would like to see.  Also noticed an issue when deleting a pair of files, one is not deleted and displayed as a single file.  Then if you try to upload a pair again, it fails with a checksum error, which makes sense since one of the files still exists.  Is this in the scope of the PR @ericenns ?

Yes I spoke to Eric about this and it was outside the scope of this PR. It is noted above in a comment after my initial PR post.

joshsadam (commented):
One last thing from me, can you add a component test for the new `viral_pill`

test/components/viral/pill_component_test.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'view_component_test_case'
+
+module Viral
+  class PillComponentTest < ViewComponentTestCase
+    test 'default' do
+      render_preview(:default)
+
+      assert_selector 'span.bg-blue-100'
```
joshsadam, line 10: **nitpick:** Can you update this to be: `assert_selector 'span.bg-blue-100', text: 'This is blue'`, that way you know the text that you are getting is within the selector.  You could also just move the `rounded-full` into each assert

app/views/projects/samples/attachments/_attachment.html.erb
```diff
@@ -2,18 +2,82 @@
...
+        sample_id: @sample.id,
+        id: attachment.id
+      ),
+      data: {
+        turbo: false
+      } %>
+    </td>
+  <% end %>
+
+  <% if attachment.metadata['format'] == 'fasta' %>
```
deepsidhu85, line 48: Let's move this `if/elsif/elsif` block into a new helper under `app/helpers` and then call it with <% format_color = helper_method_name(attachment) %>. Reason for this is we will probably need to list the sample files else where and then we have a shared method to get the correct color.

app/views/projects/samples/attachments/_attachment.html.erb
```diff
@@ -2,18 +2,82 @@
...
+    <% format_color = "blue" %>
+  <% elsif attachment.metadata['format'] == 'fastq' %>
+    <% format_color = "green" %>
+  <% elsif attachment.metadata['format'] == 'unknown' %>
+    <% format_color = "gray" %>
+  <% end %>
+  <td class="px-6 py-4"><%= viral_pill(text: attachment.metadata["format"], color: format_color) %></td>
+
+  <% if attachment.metadata['type'] == 'assembly' %>
```
deepsidhu85, line 57: Same comment as above about moving into a helper. You can put this in a separate method in the same helper file

deepsidhu85 (changes requested):
This is looking good. Just a couple of comments below

app/views/projects/samples/attachments/_attachment.html.erb
```diff
@@ -2,18 +2,82 @@
...
+            ),
+            data: {
+              turbo: false
+            } %>
+          </span>
+        </div>
+        <div class="flex items-center">
+          <%= viral_icon(name: :arrow_left, color: :subdued, classes: "h-6 w-6 ml-0 mr-2") %>
+          <span>
+            <%= link_to attachment.associated_attachment_filename,
```
deepsidhu85, line 24: I don't think we need a special method for this. You should be able to just call `attachment.associated_attachment.file.filename`. Then we can also remove the `associated_attachment_filename` method in `attachment.rb`

deepsidhu85 (changes requested):
One more comment below

test/components/viral/pill_component_test.rb
ChrisHuynh333: I actually hadn't seen this before fixing up the tests, so glad we were on the same page!

I've refactored the test and also added a `with_classes` test that includes the pills with added classes.

These can be found in commits 19f2995c11e145e34ba97df0db471b627bbeb7aa ced54dab21a367a1e4f3b5fd248c167cf38362ac 3fbee4acb627b1d20eef951c3bd70722aaca2af4

joshsadam (dismissed):
Looks good to me

test/components/viral/pill_component_test.rb
joshsadam: Looks great, thanks

app/views/projects/samples/attachments/_attachment.html.erb
ChrisHuynh333: I waffled over whether to make the pill color helper extremely generic to be usable by things outside of attachments (c43dd9224b0a9916ecf5fadf7cbc4e2bd466ae58) but decided maybe it's best to have a method within the helper specific to attachments, for less confusion with the arguments required to be passed.

Let me know what you think f93ae17f6874db9352897610bc2fa5ff92ec5f3a

app/views/projects/samples/attachments/_attachment.html.erb
ChrisHuynh333: Same as above f93ae17f6874db9352897610bc2fa5ff92ec5f3a

app/views/projects/samples/attachments/_attachment.html.erb
ChrisHuynh333: Fixed in 9fde28e1bab0cebc9ae7142848f7aa2f747db687

app/views/projects/samples/attachments/_attachment.html.erb
ChrisHuynh333: Fixed in 9fde28e1bab0cebc9ae7142848f7aa2f747db687

app/views/projects/samples/attachments/_attachment.html.erb
ChrisHuynh333: Fixed in cb2b6029026c29eec18cc930df8af9cdd894feb7

deepsidhu85 (approved):

Issue #226, opened 2023-10-05
Member access level not updated on transfer


deepsidhu85:
## Problem

When a user is a member in Group A as a `Maintainer`, and the user is a member of `Group B` as a `Guest` and then `Group B` is transferred into the `Group A` namespace, the member roles are not updated correctly and the user is left with `Maintainer` access in `Group A` and `Guest` access in `Group B`. This should not be happening as the access level should propagate into descendants starting at the parent

![image](https://github.com/phac-nml/irida-next/assets/25466211/3eadf969-8ff2-4a04-af77-60d31a200901)

![image](https://github.com/phac-nml/irida-next/assets/25466211/9020a085-ac4f-4b94-be5a-0e0234bff4e2)

**Note this issue also applies to a project transfer**

**To verify:**

Create 2 groups Group A and Group B, add `user2@email.com` as a `Maintainer` on Group A, add `user2@email.com` as a `Guest` on `Group B`. Now transfer `Group B` so that it's parent is `Group A`.  Visit the members page of `Group A` and you should see `user2@email.com` as a `Maintainer`. Now visit the members page of `Group B` and you should see `user2@email.com` as a `Guest`

## Solution

In the transfer service for groups and projects, if and only if the group has successfully been transferred to a new parent, we want to get the `Member` relationships for the new parent and it's ancestors. For each one of these member relationships we want go through the descendants of the new parent and update the memberships so that the access level is ascending through the descendants. This should ensure that at no point is the access level higher in an ancestor membership than it is in a descendant membership

**Couple of suggestions:**

1) To get the namespace and ancestor member relationships we can user the `Member.for_namespace_and_ancestors` scope which will accept a project namespace or group

2) To get the descendant relationships we can use `namespace.self_and_descendants`

3) We should most likely be able to use the `update_descendant_memberships` method in the `member` model to update descendant memberships for a user

4) We will probably want to set this up as a job that fires off once the parent has been updated and runs in the background as there could be a very large number of members

5) Break this issue up into smaller pieces ie; updating of group transfer memberships, updating of project transfer memberships, setting up job to update descendant memberships

## Acceptance criteria

After transferring a group or project under a new parent where the user was a member in the previous parent or it's ancestors, and is also a member in the new parent or it's ancestors, any membership for the user in the parent and it's descendants should be in descending order ie; the membership access level in a subgroup or project should be equal to or greater than the highest access level in the parent or it's ancestors

ksierks:
Fixed.

PR #263, opened 2023-10-20
Fix for memberships on group transfer job

ksierks:
The job fixes the memberships after a group transfer, so that the access level is ascending through its descendants.
Project transfers is out of scope in this PR.
Fixes part of https://github.com/phac-nml/irida-next/issues/226.

N/A

1. Verify the UpdateMembershipsJob is called when a group transfer is initiated. 
  a. Please see instructions from #264
2. Call the UpdateMembershipsJob from console.
  a. Open a rails console. Run `rails c` in a terminal.
  b. Call `UpdateMembershipsJob.perform_now` to start a job. Pass it a list of membership ids.
  c. Verify the memberships were updated.

## PR acceptance checklist


deepsidhu85:
After looking through the changes in the PR I think there's a few things we can change to clean up some things like having to skip/set the call back in `member.rb`.

Here is my thoughts:

Since we created a job to update the descendant memberships on group/project transfer, we should have it so that the members update service also just queues a job to update the memberships.
So in `members/update_service.rb` within the execute method we want to get the result of the `member.update` and then queue up a job to update the memberships if the `member.update` was successful. Then we want to return the result of `member.update`
With this in `member.rb` we can remove the callback `after_update` for the descendant memberships

In `test/jobs/update_memberships_job_test.rb` we can remove the skipping of the callback in the setup and remove the teardown method completely.

Instead of setting the access level and then doing a save we should update the calls to use the Members::UpdateService for example

`valid_params = { user: @group_member.user, access_level: Member::AccessLevel::MAINTAINER }`
`Members::UpdateService.new(@group_member, @group_member.namespace, @user, valid_params).execute`

For the tests we can wrap the internals of the test all within a `perform_enqueued_jobs do....end` block. Calling the Members::UpdateService can be put on the lines right above where we expect the values to have changed


With the changes above we will also need to update `test/services/members/update_service_test.rb` so when the Members::UpdateService is called we need to put a `perform_enqueued_jobs` right below it

Also, we can just include the ActiveJob::TestHelper in `test/test_helper.rb`

test/services/groups/transfer_service_test.rb
```diff
@@ -4,82 +4,39 @@
 
 module Groups
   class TransferServiceTest < ActiveSupport::TestCase
+    include ActiveJob::TestHelper
```
deepsidhu85, line 7: As per the general comment on the PR this can be moved into `test/test_helper.rb`

deepsidhu85 (changes requested):
This is working as expected but I provided some feedback in a general comment on the PR for things we can change/cleanup

test/services/groups/transfer_service_test.rb
ksierks: Please see 0ef0afd6a0e156a6483e4434ea25dde01d85eebd.

app/services/groups/transfer_service.rb
```diff
@@ -19,12 +19,12 @@ def execute(new_namespace) # rubocop:disable Metrics/AbcSize
         raise TransferError, I18n.t('services.groups.transfer.namespace_group_exists')
       end
 
-      group_ancestor_members_user_ids = Member.for_namespace_and_ancestors(@group).select(:user_id)
-      update_members = Member.for_namespace_and_ancestors(new_namespace).where(user_id: group_ancestor_members_user_ids)
+      group_ancestor_member_user_ids = Member.for_namespace_and_ancestors(@group).select(:user_id)
+      new_namespace_member_ids = Member.for_namespace_and_ancestors(new_namespace).where(user_id: group_ancestor_member_user_ids).select(&:id) # rubocop:disable Layout/LineLength
```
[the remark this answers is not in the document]
ksierks: Please see 1349935d2b0305ac3128821f7ce5200f27df0bb4.

ksierks:
> @ksierks After looking through the changes in the PR I think there's a few things we can change to clean up some things like having to skip/set the call back in `member.rb`.
> 
> Here is my thoughts:
> 
> Since we created a job to update the descendant memberships on group/project transfer, we should have it so that the members update service also just queues a job to update the memberships. So in `members/update_service.rb` within the execute method we want to get the result of the `member.update` and then queue up a job to update the memberships if the `member.update` was successful. Then we want to return the result of `member.update` With this in `member.rb` we can remove the callback `after_update` for the descendant memberships
> 
> In `test/jobs/update_memberships_job_test.rb` we can remove the skipping of the callback in the setup and remove the teardown method completely.
> 
> Instead of setting the access level and then doing a save we should update the calls to use the Members::UpdateService for example
> 
> `valid_params = { user: @group_member.user, access_level: Member::AccessLevel::MAINTAINER }` `Members::UpdateService.new(@group_member, @group_member.namespace, @user, valid_params).execute`
> 
> For the tests we can wrap the internals of each test within a `perform_enqueued_jobs do....end` block. Calling the Members::UpdateService can be put on the lines right above where we expect the values to have changed
> 
> With the changes above we will also need to update `test/services/members/update_service_test.rb` so when the Members::UpdateService is called we need to put a `perform_enqueued_jobs` right below it
> 
> Also, we can just include the ActiveJob::TestHelper in `test/test_helper.rb`

Hopefully, I addressed everything in e500611feb682f07b21d52e0b35a28adec66dc25, except "Instead of setting the access level and then doing a save we should update the calls to use the Members::UpdateService". I am only wanting to test the job in update_memberships_job_test.rb. If I call Members::UpdateService.execute, it calls the job that I'm trying to test. It's a little too cyclical.

deepsidhu85 (approved):

ChrisHuynh333 (approved):
Works as described, nothing to add on my end!

PR #291, opened 2023-11-01
Attachment destroy service

ChrisHuynh333:
This PR creates the destroy service for attachments, moving the destroy functionality away from the attachments controller. In addition to the destroy service, deletion of paired sample attachments is now also fixed whereby the shared delete button on the samples attachment page for paired attachments will now delete both associated attachments. 

[Screencast from 2023-11-03 09:23:46 AM.webm](https://github.com/phac-nml/irida-next/assets/82407232/49204e67-e806-486f-921c-007aa0773132)

Upload both single and paired attachments to a sample and delete, ensuring delete functionality is handled as expected.

## PR acceptance checklist

### DIFF
--- a/app/controllers/projects/samples/attachments_controller.rb
+++ b/app/controllers/projects/samples/attachments_controller.rb
@@ -38,23 +38,32 @@ module Projects
         end
       end
 
-      def destroy
+      def destroy # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
         authorize! @project, to: :destroy_sample?
-        return destroy_error if @attachment.attachable_type != 'Sample' || @attachment.attachable_id != @sample.id
-
-        @destroyed_attachments = ::Attachments::DestroyService.new(@sample, @attachment, current_user).execute
-
-        return unless @destroyed_attachments
-
-        status = if @destroyed_attachments.count.positive?
-                   destroy_status(@attachment, @destroyed_attachments.length)
-                 else
-                   :unprocessable_entity
-                 end
-
         respond_to do |format|
-          format.turbo_stream do
-            render status:, locals: { destroyed_attachments: @destroyed_attachments }
+          if @attachment.attachable_type != 'Sample' || @attachment.attachable_id != @sample.id
+            format.turbo_stream do
+              render status: :bad_request,
+                     locals: { type: 'alert',
+                               message: t('.error',
+                                          filename: @attachment.file.filename,
+                                          errors: "Attachment does not belong to #{@sample.name}"),
+                               destroyed_attachments: nil }
+            end
+          else
+            @destroyed_attachments = ::Attachments::DestroyService.new(@sample, @attachment, current_user).execute
+
+            return unless @destroyed_attachments
+
+            status = if @destroyed_attachments.count.positive?
+                       destroy_status(@attachment, @destroyed_attachments.length)
+                     else
+                       :unprocessable_entity
+                     end
+
+            format.turbo_stream do
+              render status:, locals: { destroyed_attachments: @destroyed_attachments }
+            end
           end
         end
       end
@@ -79,19 +88,6 @@ module Projects
         return count == 2 ? :ok : :multi_status if attachment.associated_attachment
 
         count == 1 ? :ok : :unprocessable_entity
-      end
-
-      def destroy_error
-        respond_to do |format|
-          format.turbo_stream do
-            render status: :bad_request,
-                   locals: { type: 'alert',
-                             message: t('.error',
-                                        filename: @attachment.file.filename,
-                                        errors: "Attachment does not belong to #{@sample.name}"),
-                             destroyed_attachments: nil }
-          end
-        end
       end
     end
   end
--- a/app/controllers/projects/samples/attachments_controller.rb
+++ b/app/controllers/projects/samples/attachments_controller.rb
@@ -40,17 +40,18 @@ module Projects
 
       def destroy
         authorize! @project, to: :update_sample?
+
+        @destroyed_attachments = ::Attachments::DestroyService.new(@sample, @attachment, current_user).execute
+
+        return unless @destroyed_attachments
+
+        status = if @destroyed_attachments.count.positive?
+                   destroy_status(@attachment, @destroyed_attachments.length)
+                 else
+                   :unprocessable_entity
+                 end
+
         respond_to do |format|
-          @destroyed_attachments = ::Attachments::DestroyService.new(@sample, @attachment, current_user).execute
-
-          return unless @destroyed_attachments
-
-          status = if @destroyed_attachments.count.positive?
-                     destroy_status(@attachment, @destroyed_attachments.length)
-                   else
-                     :unprocessable_entity
-                   end
-
           format.turbo_stream do
             render status:, locals: { destroyed_attachments: @destroyed_attachments }
           end
--- a/app/controllers/projects/samples/attachments_controller.rb
+++ b/app/controllers/projects/samples/attachments_controller.rb
@@ -38,22 +38,24 @@ module Projects
         end
       end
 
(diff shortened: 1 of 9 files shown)


app/controllers/projects/samples/attachments_controller.rb
```diff
@@ -28,12 +28,23 @@ def create
       end
 
       def destroy
-        authorize! @project, to: :update_sample?
+        authorize! @project, to: :destroy?
```
deepsidhu85, line 31: We should use the `:destroy_sample?` in the project policy for this authorization

app/controllers/projects/samples/attachments_controller.rb
```diff
@@ -28,12 +28,23 @@ def create
...
+        return unless @attachment.attachable_type == 'Sample' && @attachment.attachable_id == @sample.id
```
deepsidhu85, line 32: We should return an error to display to the user instead of just just returning

app/services/attachments/destroy_service.rb
```diff
@@ -0,0 +1,27 @@
...
+  class DestroyService < BaseService
+    def initialize(attachable, attachment, user = nil)
+      super(user)
+      @attachable = attachable
+      @attachment = attachment
+    end
+
+    def execute
+      authorize! @attachable.project, to: :destroy?
+
```
deepsidhu85, line 14: Before deleting we should make sure that the attachment (and it's associated attachment if paired-end) belongs to the sample. If not an appropriate error should be attached to the `@attachment`. You can take a look at the other services to see examples of how the errors are attached

deepsidhu85 (changes requested):
This is working great! Just a few comments for now

app/controllers/projects/samples/attachments_controller.rb
ChrisHuynh333: Fixed in d9f8225c5d145849cec6a8481c7897234d4b8b2d

app/controllers/projects/samples/attachments_controller.rb
ChrisHuynh333: Fixed in d9f8225c5d145849cec6a8481c7897234d4b8b2d

app/services/attachments/destroy_service.rb
ChrisHuynh333: Fixed in 4376c31d10cd19542a8c6f5a924ecc659f51f78c

app/controllers/projects/samples/attachments_controller.rb
```diff
@@ -27,13 +27,33 @@ def create
         end
       end
 
-      def destroy
-        authorize! @project, to: :update_sample?
+      def destroy # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+        authorize! @project, to: :destroy_sample?
+        respond_to do |format|
+          if @attachment.attachable_type != 'Sample' || @attachment.attachable_id != @sample.id
```
ericenns, line 33: I'm not sure how this if condition could ever be hit. The reason being that `@attachment` is queried from the `@sample`, so it should only be set if the attachment belongs to the sample. There is also no test for this condition, so that would be required if this if statement is possible to be met which I cannot see it ever happening. If the user were trying to delete an attachment that doesn't belong to the sample they would get `not_found` as defined within the `attachment` method in this controller.

ericenns (dismissed):
Generally this looks good, but I did find an if statement that would be impossible to hit which is noted in my comments.

app/controllers/projects/samples/attachments_controller.rb
ericenns: Following up, this is what happens when I change the attachment id in the `delete` form button to point to an attachment that belongs to another Sample.

image

Which confirms my comments that this condition can never be met.

app/controllers/projects/samples/attachments_controller.rb
deepsidhu85: Sorry Chris I made a mistake here, we should change this back to `update_sample?`.  The `destroy_sample?` authorization method only allows a member who is an owner to delete attachments, but we want to allow maintainers to delete attachments as well

app/controllers/projects/samples/attachments_controller.rb
ChrisHuynh333: > Sorry Chris I made a mistake here, we should change this back to `update_sample?`. The `destroy_sample?` authorization method only allows a member who is an owner to delete attachments, but we want to allow maintainers to delete attachments as well

I believe this should be up here. Fixed in ffb0a839fee97a3e3eb2bf7596e329fc451758ee

app/controllers/projects/samples/attachments_controller.rb
ChrisHuynh333: > I'm not sure how this if condition could ever be hit. The reason being that `@attachment` is queried from the `@sample`, so it should only be set if the attachment belongs to the sample. There is also no test for this condition, so that would be required if this if statement is possible to be met which I cannot see it ever happening. If the user were trying to delete an attachment that doesn't belong to the sample they would get `not_found` as defined within the `attachment` method in this controller.

Based on @ericenns comment below, I've removed the check in ffb0a839fee97a3e3eb2bf7596e329fc451758ee

app/controllers/projects/samples/attachments_controller.rb
ChrisHuynh333: Removed in ffb0a839fee97a3e3eb2bf7596e329fc451758ee

deepsidhu85 (changes requested):
This is looking good overall just a few more comments

test/services/attachments/destroy_service_test.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
```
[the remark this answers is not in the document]
ChrisHuynh333: Added in 3f2b8e27a44b3a94983c26198cee222bb520f77e

app/services/attachments/destroy_service.rb
```diff
@@ -0,0 +1,37 @@
...
+    def execute # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+      authorize! @attachable.project, to: :destroy?
+
+      if @attachment.attachable_id != @attachable.id || !@attachable.is_a?(@attachment.attachable_type.constantize)
+        raise AttachmentsDestroyError, I18n.t('services.attachments.destroy.does_not_belong_to_attachable')
+      end
+
+      destroyed_attachments = []
+      if @attachable.instance_of?(Sample) && @attachment.associated_attachment
```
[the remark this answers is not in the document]
ChrisHuynh333: Fixed in 3f2b8e27a44b3a94983c26198cee222bb520f77e

app/services/attachments/destroy_service.rb
```diff
@@ -0,0 +1,37 @@
...
+    def initialize(attachable, attachment, user = nil)
+      super(user)
+      @attachable = attachable
+      @attachment = attachment
+    end
+
+    def execute # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+      authorize! @attachable.project, to: :destroy?
+
+      if @attachment.attachable_id != @attachable.id || !@attachable.is_a?(@attachment.attachable_type.constantize)
```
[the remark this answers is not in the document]
ChrisHuynh333: Fixed in 3f2b8e27a44b3a94983c26198cee222bb520f77e

app/services/attachments/destroy_service.rb
ChrisHuynh333: As mentioned below, fixed in 3f2b8e27a44b3a94983c26198cee222bb520f77e

deepsidhu85 (approved):

PR #311, opened 2023-11-20
Metadata update service

ChrisHuynh333:
This PR creates the new project sample metadata update service. 

1. Open the rails console with `rails c`
2. Either create or assign a Sample (preferably that contains an empty metadata field) and assign it to variable `sample`.
3. Assign the sample's associated project to `project`.
4. Assign a user that is authorized to update the sample to `user`

**Keep the console values for each testing phase**

5. Test adding a new metadata hash to the sample and metadata_provenance with user:
a) Assign a new variable `params_1` with a new hash `metadata` (such as `params_1 = {'metadata' => {'key1': 'value1', 'key2': 'value2'}}`)
b) Execute the update service with `Samples::Metadata::UpdateService.new(project, sample, user, params_1).execute`
c) Check the metadata field of the `sample` to ensure it equals your `metadata` value, and `sample.metadata_provenance` equals `id: user.id` and `source: user` for each key.

6. Test merging a new metadata hash to a sample containing metadata, using an analysis for metadata_provenance:
a) Create and assign a new `params_2` hash which includes both `metadata` and `analysis_id`. Ensure it has an overlapping metadata key with a different value that exists in `sample.metadata`. For example, for the above sample, you could use `params_2 = {'metadata' => {'key3': 'value3', 'key1': 'value4'}, 'analysis_id' => 1}`.
b) Execute the update service with `Samples::Metadata::UpdateService.new(project, sample, user, params_2).execute`.
c) Check that the overlapping key of `sample.metadata` now has the value from `params_2['metadata']` and not from `params_1['metadata']` and any new key value pairs you may have assigned. Ensure any overwritten and new metadata fields have the accompanying `metadata_provenance` fields with `id: 1` and `source: analysis`.

7. Test merging a new metadata hash to a sample containing metadata, using an user for metadata_provenance:
Note: A user should not be able to overwrite any metadata added by an `analysis`
a) Create and assign a new `params_3` hash which includes `metadata`. Ensure it has an overlapping metadata key with a different value that exists in `sample.metadata`, and that metadata key exists in `sample.metadata_provenance` and was added by an analysis. For example, for the above sample, you could use `params_3 = {'metadata' => {'key3': 'value6', 'key5': 'value5'}}`.
b) Execute the update service with `Samples::Metadata::UpdateService.new(project, sample, user, params_3).execute`.
c) Check that the overlapping key of `sample.metadata` does NOT have the new value from `params_3['metadata']`, but any new key value pairs should now be assigned. Ensure new metadata fields have the corresponding `user` source in `metadata_provenance` and any fields that were not overwritten do not have changes to `metadata_provenance`.

8. Test removal of a key value pair from `sample.metadata`:
a) Assign a params hash containing a metadata key that exists in `sample.metadata` but with an empty string value, like `params_4 = {'metadata' => {'key3': ''}}`
b) Execute the update service with  `Samples::Metadata::UpdateService.new(project, sample, user, params_4).execute`.
c) Ensure the `key` passed to the service was removed from `sample.metadata` and `sample.metadata_provenance`. Ensure to test both `user` and `analysis` source can delete both a `user` and `analysis` added metadata field.

## PR acceptance checklist

### DIFF
--- a/app/services/samples/metadata/update_service.rb
+++ b/app/services/samples/metadata/update_service.rb
@@ -15,9 +15,8 @@ module Samples
         @analysis_id = params['analysis_id']
       end
 
-      def execute # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+      def execute
         authorize! sample.project, to: :update_sample?
-        metadata_fields_update_status = { updated: [], not_updated: [] }
 
         validate_sample_in_project
 
@@ -25,29 +24,15 @@ module Samples
 
         transform_metadata_keys
 
-        @metadata.each do |key, value|
-          if value.blank?
-            if @sample['metadata'].key?(key)
-              @sample['metadata'].delete(key)
-              @sample['metadata_provenance'].delete(key)
-              metadata_fields_update_status[:updated].append(key)
-            end
-          else
-            metadata_updated = update_metadata(key, value)
-            if metadata_updated
-              metadata_fields_update_status[:updated].append(key)
-            else
-              metadata_fields_update_status[:not_updated].append(key)
-            end
-          end
-        end
+        metadata_fields_update_status = perform_metadata_update
+
         @sample.save
 
         check_not_updated_metadata_fields(metadata_fields_update_status[:not_updated])
         metadata_fields_update_status[:updated]
       rescue Samples::Metadata::UpdateService::SampleMetadataUpdateError => e
         @sample.errors.add(:base, e.message)
-        metadata_fields_update_status[:updated]
+        metadata_fields_update_status ? metadata_fields_update_status[:updated] : []
       end
 
       private
@@ -72,7 +57,24 @@ module Samples
         @metadata = @metadata.transform_keys(&:to_s)
       end
 
-      def update_metadata(key, value)
+      def perform_metadata_update
+        update_status = { updated: [], not_updated: [] }
+        @metadata.each do |key, value|
+          if value.blank?
+            if @sample['metadata'].key?(key)
+              @sample['metadata'].delete(key)
+              @sample['metadata_provenance'].delete(key)
+              update_status[:updated].append(key)
+            end
+          else
+            metadata_updated = assign_metadata_to_sample(key, value)
+            metadata_updated ? update_status[:updated].append(key) : update_status[:not_updated].append(key)
+          end
+        end
+        update_status
+      end
+
+      def assign_metadata_to_sample(key, value)
         # We don't overwrite existing @sample['metadata_provenance'] or @sample['metadata']
         # that has a {source: 'analysis'} with a user
         if @sample['metadata_provenance'].key?(key) && @analysis_id.nil? &&
--- a/app/services/samples/metadata/update_service.rb
+++ b/app/services/samples/metadata/update_service.rb
@@ -56,20 +56,20 @@ module Samples
 
       private
 
-      def update_metadata_provenance(key) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity, Metrics/MethodLength
+      def update_metadata_provenance(key) # rubocop:disable Metrics/MethodLength
         if @sample['metadata_provenance'].key?(key)
           # We don't overwrite existing @sample['metadata_provenance'] or @sample['metadata']
           # that has a {source: 'analysis'} with a user
-          if @analysis_id
-            @sample['metadata_provenance'][key] = { source: 'analysis', id: @analysis_id }
-            true
-          elsif @analysis_id.nil?
+          if @analysis_id.nil?
             if @sample['metadata_provenance'][key]['source'] == 'user'
               @sample['metadata_provenance'][key] = { source: 'user', id: current_user.id }
               true
-            elsif @sample['metadata_provenance'][key]['source'] == 'analysis'
+            else
               false
             end
+          else
+            @sample['metadata_provenance'][key] = { source: 'analysis', id: @analysis_id }
+            true
           end
         else
           @sample['metadata_provenance'][key] =
--- a/app/services/samples/metadata/update_service.rb
+++ b/app/services/samples/metadata/update_service.rb
@@ -24,7 +24,7 @@ module Samples
 
         transform_metadata_keys
 
-        metadata_fields_update_status = metadata_modification
+        metadata_fields_update_status = perform_metadata_update
 
         @sample.save
 
@@ -57,13 +57,13 @@ module Samples
         @metadata = @metadata.transform_keys(&:to_s)
       end
 
-      def metadata_modification
+      def perform_metadata_update
         update_status = { updated: [], not_updated: [] }
         @metadata.each do |key, value|
           if value.blank?
-            if @sample['metadata'].key?(key)
-              @sample['metadata'].delete(key)
-              @sample['metadata_provenance'].delete(key)
+            if @sample.metadata.key?(key)
+              @sample.metadata.delete(key)
+              @sample.metadata_provenance.delete(key)
               update_status[:updated].append(key)
             end
           else
@@ -75,15 +75,15 @@ module Samples
       end
 
       def assign_metadata_to_sample(key, value)
-        # We don't overwrite existing @sample['metadata_provenance'] or @sample['metadata']
+        # We don't overwrite existing @sample.metadata_provenance or @sample.metadata
         # that has a {source: 'analysis'} with a user
-        if @sample['metadata_provenance'].key?(key) && @analysis_id.nil? &&
-           @sample['metadata_provenance'][key]['source'] == 'analysis'
+        if @sample.metadata_provenance.key?(key) && @analysis_id.nil? &&
+           @sample.metadata_provenance[key]['source'] == 'analysis'
           false
         else
-          @sample['metadata_provenance'][key] =
+          @sample.metadata_provenance[key] =
             @analysis_id.nil? ? { source: 'user', id: current_user.id } : { source: 'analysis', id: @analysis_id }
-          @sample['metadata'][key] = value
+          @sample.metadata[key] = value
           true
         end
       end
--- a/app/services/samples/metadata/update_service.rb
+++ b/app/services/samples/metadata/update_service.rb
@@ -34,12 +34,8 @@ module Samples
                 @sample['metadata'].delete(key) && @sample['metadata_provenance'].delete(key)
               end
             else
-              provenance_updated = update_metadata_provenance(key)
-              if provenance_updated
-                assign_metadata_value(key, value)
-              else
-                metadata_fields_not_updated.append("#{key}: #{value}")
-              end
+              metadata_updated = update_metadata(key, value)
+              !metadata_updated && metadata_fields_not_updated.append("#{key}: #{value}")
             end
           end
           @sample.update(id: @sample.id)
@@ -60,30 +56,18 @@ module Samples
 
       private
 
-      def update_metadata_provenance(key) # rubocop:disable Metrics/MethodLength
-        if @sample['metadata_provenance'].key?(key)
-          # We don't overwrite existing @sample['metadata_provenance'] or @sample['metadata']
-          # that has a {source: 'analysis'} with a user
-          if @analysis_id.nil?
-            if @sample['metadata_provenance'][key]['source'] == 'user'
-              @sample['metadata_provenance'][key] = { source: 'user', id: current_user.id }
-              true
-            else
-              false
-            end
-          else
-            @sample['metadata_provenance'][key] = { source: 'analysis', id: @analysis_id }
-            true
-          end
+      def update_metadata(key, value)
+        # We don't overwrite existing @sample['metadata_provenance'] or @sample['metadata']
+        # that has a {source: 'analysis'} with a user
+        if @sample['metadata_provenance'].key?(key) && @analysis_id.nil? &&
+           @sample['metadata_provenance'][key]['source'] == 'analysis'
+          false
         else
           @sample['metadata_provenance'][key] =
             @analysis_id.nil? ? { source: 'user', id: current_user.id } : { source: 'analysis', id: @analysis_id }
+          @sample['metadata'][key] = value
           true
         end
-      end
-
-      def assign_metadata_value(key, value)
-        @sample['metadata'][key] = value
       end
     end
   end
--- a/app/services/samples/metadata/update_service.rb
+++ b/app/services/samples/metadata/update_service.rb
@@ -7,15 +7,22 @@ module Samples
       SampleMetadataUpdateError = Class.new(StandardError)
       attr_accessor :sample, :metadata, :metadata_key
 
-      def initialize(sample, user = nil, params = {}, metadata = nil, metadata_key = nil) # rubocop:disable Metrics/ParameterLists
+      def initialize(project, sample, user = nil, params = {}, metadata = nil, metadata_key = nil) # rubocop:disable Metrics/ParameterLists
         super(user, params.except(:sample, :id))
+        @project = project
         @sample = sample
         @metadata = metadata
         @metadata_key = metadata_key
       end
 
-      def execute
+      def execute # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
         authorize! sample.project, to: :update_sample?
+
+        if @project.id != @sample.project.id
+          raise SampleMetadataUpdateError,
+                I18n.t('services.samples.metadata.sample_does_not_belong_to_project', sample_name: @sample.name,
+                                                                                      project_name: @project.name)
+        end
 
         if @metadata
           @metadata = @metadata.transform_keys(&:to_s)
--- a/app/services/samples/metadata/update_service.rb
+++ b/app/services/samples/metadata/update_service.rb
@@ -7,12 +7,12 @@ module Samples
       SampleMetadataUpdateError = Class.new(StandardError)
       attr_accessor :sample, :metadata, :analysis_id
 
-      def initialize(project, sample, user = nil, params = {}, metadata = nil, analysis_id = nil) # rubocop:disable Metrics/ParameterLists
+      def initialize(project, sample, user = nil, params = {})
         super(user, params)
         @project = project
         @sample = sample
-        @metadata = metadata
-        @analysis_id = analysis_id
+        @metadata = params['metadata']
+        @analysis_id = params['analysis_id']
       end
 
       def execute # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/MethodLength
@@ -25,14 +25,16 @@ module Samples
         end
 
         if @metadata
+          # Without transforming keys, issues with overwritting can occur and multiples of the same key can appear
           @metadata = @metadata.transform_keys(&:to_s)
-          @metadata.each do |k, v|
-            if v.blank?
-              @sample['metadata'].delete(k) && @sample['metadata_provenance'].delete(k) if @sample['metadata'].key?(k)
+          @metadata.each do |key, value|
+            if value.blank?
+              if @sample['metadata'].key?(key)
+                @sample['metadata'].delete(key) && @sample['metadata_provenance'].delete(key)
+              end
             else
-              @sample['metadata'][k] = v
-              @sample['metadata_provenance'][k] =
-                @analysis_id.nil? ? { source: 'user', id: current_user.id } : { source: 'analysis', id: @analysis_id }
+              @sample['metadata'][key] = value
+              update_sample_metadata_provenance(key, current_user)
             end
           end
         else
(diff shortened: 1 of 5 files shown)


ChrisHuynh333:
There's a couple things that could be added/changed depending how we end up using the service within the website
- Params I haven't really implemented depending on which params end up getting passed
- I assume a `bool` can be returned to get feedback for update/error messages

Let me know if you guys already have an idea how these will be implemented and I can do that now, or we can wait until the sample metadata implementation is further fleshed out!

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Samples
+  module Metadata
+    # Service used to Update Samples::Metadata
+    class UpdateService < BaseService
+      SampleMetadataUpdateError = Class.new(StandardError)
+      attr_accessor :sample, :metadata, :metadata_key
```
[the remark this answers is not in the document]
ChrisHuynh333: Fixed in 3cdb2c8c2344064f0eb49ce032137ffb099a7bf2 as well as `metadata_provenance` was added

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,49 @@
...
+                                                                                      project_name: @project.name)
+        end
+
+        if @metadata
+          @metadata = @metadata.transform_keys(&:to_s)
+          @metadata.each do |k, v|
+            if v.blank?
+              @sample['metadata'].delete(k) && @sample['metadata_provenance'].delete(k) if @sample['metadata'].key?(k)
+            else
+              @sample['metadata'][k] = v
```
deepsidhu85, line 33: We probably should update the code here to not allow a user to be able to overwrite metadata fields/values that were previously added by an analysis. Another analysis could update it, but if it's a user update then we should probably not allow that. Thoughts @ericenns

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,49 @@
...
+        authorize! sample.project, to: :update_sample?
+
+        if @project.id != @sample.project.id
+          raise SampleMetadataUpdateError,
+                I18n.t('services.samples.metadata.sample_does_not_belong_to_project', sample_name: @sample.name,
+                                                                                      project_name: @project.name)
+        end
+
+        if @metadata
+          @metadata = @metadata.transform_keys(&:to_s)
```
deepsidhu85, line 28: Is there a reason to transform the keys here?

deepsidhu85 (changes requested):
This is looking good, just a few comments below

app/services/samples/metadata/update_service.rb
ericenns: Yeah I agree a separate method to filter out the keys where the `metadata_provenance` indicates that it was last touched by an `analysis` should not overwritten by a user. Though a `User` should be able to remove those keys.

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module Samples
+  module Metadata
+    # Service used to Update Samples::Metadata
+    class UpdateService < BaseService
+      SampleMetadataUpdateError = Class.new(StandardError)
+      attr_accessor :sample, :metadata, :analysis_id
+
+      def initialize(project, sample, user = nil, params = {}, metadata = nil, analysis_id = nil) # rubocop:disable Metrics/ParameterLists
```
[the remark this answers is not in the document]
ChrisHuynh333: Yup, makes sense. Updated in e64d71ce473cbb8064cd28816d776d62536d2d07

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Updated in e64d71ce473cbb8064cd28816d776d62536d2d07 and c789fc25fe68ba49a34ea3f76540e6597d3eed2d

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,74 @@
...
+        if @metadata
+          # Without transforming keys, issues with overwritting can occur and multiples of the same key can appear
+          @metadata = @metadata.transform_keys(&:to_s)
+          @metadata.each do |key, value|
+            if value.blank?
+              if @sample['metadata'].key?(key)
+                @sample['metadata'].delete(key) && @sample['metadata_provenance'].delete(key)
+              end
+            else
+              update_metadata(key, value, current_user)
```
deepsidhu85, line 36: We shouldn't need to pass the `current_user` as an argument since it's already set as an attribute accessor in the base service. You should be able to just use `current_user` in the `update_metadata` method

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,74 @@
...
+
+        @sample.update(id: @sample.id)
+      rescue Samples::Metadata::UpdateService::SampleMetadataUpdateError => e
+        @sample.errors.add(:base, e.message)
+      end
+
+      private
+
+      def update_metadata(key, value, user)
+        if @sample['metadata_provenance'].key?(key)
```
deepsidhu85, line 52: In this method we should probably return which key -> value wasn't updated due to the previous value being set by an analysis and now a user is trying to update it. Then in the `execute` method above within the .each loop keep track of the keys that are returned from the `update_metadata` method. We can then attach an error message which lists the keys that weren't updated to `samples.errors -> base`.

deepsidhu85 (changes requested):
Just a few comments below before I test it out

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Fixed in 3b73978b5eaf0aa163e17bb5034f7b9b62c8bf9b

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Updated in 3b73978b5eaf0aa163e17bb5034f7b9b62c8bf9b
Message updated in 0e24012480f59df2a7803e01baa1d8a8f8192615

ChrisHuynh333: I also have both the metadata field and value in the message, so it will read: `Metadata fields 'key1: value1, key2: value2, etc.' of sample {sample_name} cannot be overwritten by a user as it was previously updated with an analysis. `

In terms of error msg readability, I'm not sure value is necessary here and the metadata field is sufficient for letting the user know it didn't update. Thoughts?

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,74 @@
...
+          @metadata = @metadata.transform_keys(&:to_s)
+          metadata_fields_not_updated = []
+          @metadata.each do |key, value|
+            if value.blank?
+              if @sample['metadata'].key?(key)
+                @sample['metadata'].delete(key) && @sample['metadata_provenance'].delete(key)
+              end
+            else
+              metadata_updated = update_metadata(key, value)
+              !metadata_updated && metadata_fields_not_updated.append("#{key}: #{value}")
```
deepsidhu85, line 38: I think appending just the key should be good enough. The value portion doesn't really add anything to the error message.

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,74 @@
...
+            end
+          end
+          @sample.update(id: @sample.id)
+        else
+          raise SampleMetadataUpdateError,
```
deepsidhu85, line 43: Just thinking about the next part of this which is going to be the metadata controller, we will need the service to return true/false if it was updated or not. The `@sample.update(id: @sample.id)` will return a boolean by default, but here we are just raising an error and not returning anything. I still need to test it out but I believe we can raise the error and do a return in this case we would return false after the error is raised. Then in the controller we can check if it was updated or not and if it was updated successfully we can do a secondary check to see if we need to return an error message for any keys that weren't updated

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Updated in 8a632ec0a9396186a5fa9bbadacc4b27e3f8c76b

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Updated in 8a632ec0a9396186a5fa9bbadacc4b27e3f8c76b

I explicitly returned a `true` if `metadata_fields_not_updated.count == 0` as it seemed without it, the expected return was not behaving as expected. The problem I can see is if partial metadata fields were updated, and others not, in the same execution. 
I imagine in the view, we would want to provide both a success and failure feedback to the user via flash msg. Currently, by returning false if we have an error msg (ie: empty metadata, or partial metadata not updated), I don't see a clean way of still providing a success to the user that some fields were updated.

Would returning a list of updated metadata fields work? Return updated metadata fields to controller, and still check for errors in the controller.
If updated metadata count == 0 : Display error (whether it's empty metadata, or all fields not updated)
If updated metadata count positive? : still check error and run both success and error if error (captures both all metadata updated, or partial updated/not updated)

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,84 @@
...
+            else
+              metadata_updated = update_metadata(key, value)
+              if metadata_updated
+                metadata_fields_update_status[:updated].append(key)
+              else
+                metadata_fields_update_status[:not_updated].append(key)
+              end
+            end
+          end
+          @sample.update(id: @sample.id)
```
ericenns, line 50: **nitpick (blocking):** What is this link supposed to do? if you just want to persist changes to the metadata after the if then just call `@sample.save`.

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      def initialize(project, sample, user = nil, params = {})
+        super(user, params)
+        @project = project
+        @sample = sample
+        @metadata = params['metadata']
+        @analysis_id = params['analysis_id']
+      end
+
+      def execute # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/MethodLength
```
ericenns, line 18: **nitpick (blocking):** I'd like you to split up this method into smaller ones to reduce number of rubocop disable and also have it more readable. I will include comments on specific areas that I think can be refactored.

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,84 @@
...
+        authorize! sample.project, to: :update_sample?
+        metadata_fields_update_status = { updated: [], not_updated: [] }
+
+        if @project.id != @sample.project.id
```
ericenns, line 22: **nitpick (blocking):** Can you move this if statement into a private method named `validate_sample_in_project`.

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,84 @@
...
+          raise SampleMetadataUpdateError,
+                I18n.t('services.samples.metadata.sample_does_not_belong_to_project', sample_name: @sample.name,
+                                                                                      project_name: @project.name)
+        end
+
+        if @metadata.nil? || @metadata == {}
```
ericenns, line 28: **nitpick (blocking):** can you move the first part of this if statement into `validate_metadata_param`. After doing that you won't need the if/else logic since the if statement would be in the `validate_metadata_param`.

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,84 @@
...
+                         sample_name: @sample.name,
+                         metadata_fields: metadata_fields_update_status[:not_updated].join(', '))
+          else
+            metadata_fields_update_status[:updated]
+          end
+
+        end
+      rescue Samples::Metadata::UpdateService::SampleMetadataUpdateError => e
+        @sample.errors.add(:base, e.message)
+        metadata_fields_update_status[:updated]
```
ericenns, line 64: **nitpick (blocking):** Should this have `:updated` status when there is an error?

test/services/samples/metadata/update_service_test.rb
ericenns, line 1: **nitpick (blocking):** Can your tests be updated to test the return values from the service as from reading the service above and these test I don't understand what the expected output is.

ericenns (changes requested):
I have a few comments below, after addressing these I may have some more if we still have a lot of rubocop disable comments. In general the logic looks right, but I would like some refactoring to be able to better read the code.

app/services/samples/metadata/update_service.rb
ChrisHuynh333: I discussed this above with Deep https://github.com/phac-nml/irida-next/pull/311#discussion_r1417766067

This is mainly to capture if metadata was passed and some fields updated, while others did not (mainly by if user tries to update a field already updated by an analysis). This is under the assumption we want to provide the user with both a success and failure feedback for the fields that were updated and not updated via the controller. Without returning the `updated` list, I couldn't think of a way to check if some fields were updated. So this way, we can check if any fields were updated with the returned list, and anything `not_updated` would be contained in the `@sample.errors`.

Would love to hear if you have another idea on how to approach this solution if one comes to mind!

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Fixed in b78174939f9a80456d3eb8fbc8342885608f718e

ChrisHuynh333: Fixed in b78174939f9a80456d3eb8fbc8342885608f718e
Moved `key_transformation` and checking `not_updated_metadata_fields` to their own private methods too, was able to get rid of `cyclomatic` and `perceived complexity` disables.

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Fixed in b78174939f9a80456d3eb8fbc8342885608f718e
Moved `key_transformation` and checking `not_updated_metadata_fields` to their own private methods too, was able to get rid of `cyclomatic` and `perceived complexity` disables.

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Fixed in b78174939f9a80456d3eb8fbc8342885608f718e

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Fixed in b78174939f9a80456d3eb8fbc8342885608f718e

test/services/samples/metadata/update_service_test.rb
ChrisHuynh333: Fixed in b78174939f9a80456d3eb8fbc8342885608f718e

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,99 @@
...
+        authorize! sample.project, to: :update_sample?
+        metadata_fields_update_status = { updated: [], not_updated: [] }
+
+        validate_sample_in_project
+
+        validate_metadata_param
+
+        transform_metadata_keys
+
+        @metadata.each do |key, value|
```
deepsidhu85, line 28: We should probably move this block into it's own method as well which will return the ` metadata_fields_update_status`

deepsidhu85 (changes requested):
Still looking through the refactor, but have one comment below for now

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Updated in f28954cde8515abffe12e9ac9b777ca728ecce83

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        metadata_fields_update_status = metadata_modification
+
+        @sample.save
+
+        fields_not_updated(metadata_fields_update_status)
+        metadata_fields_update_status
+      rescue Samples::Metadata::UpdateService::SampleMetadataUpdateError => e
+        @sample.errors.add(:base, e.message)
+        metadata_fields_update_status || { updated: [], not_updated: [] }
```
deepsidhu85, line 35: You should be able to just return `metadata_fields_update_status` here as well. In the controller we can always check for a nil value as well

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,101 @@
...
+        else
+          @sample['metadata_provenance'][key] =
+            @analysis_id.nil? ? { source: 'user', id: current_user.id } : { source: 'analysis', id: @analysis_id }
+          @sample['metadata'][key] = value
+          true
+        end
+      end
+
+      def fields_not_updated(metadata_fields_update_status)
+        return unless metadata_fields_update_status.count.positive?
```
deepsidhu85, line 92: Since we are only interested in the `not_updated` fields, assign `metadata_fields_update_status[:not_updated]` to a variable then check the count on that, and then use the variable below when setting the metadata_fields in the error

deepsidhu85 (changes requested):
This is almost good to go. Just a couple of small comments

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Updated in 5ab62ea9e785867b0a42761d594ba86b2f4f0a82

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Updated in 5ab62ea9e785867b0a42761d594ba86b2f4f0a82

app/services/samples/metadata/update_service.rb
deepsidhu85: This has been updated to just return the `metadata_fields_update_status` which is a hash that includes updated: [] and not_updated: []. In the controller we can just check to see if `updated` has a length to determine whether it will go into the success block or the error block of the if/else in the controller

deepsidhu85 (dismissed):
This looks good to me! Thanks for making all the changes

test/services/samples/metadata/update_service_test.rb
ericenns, line 1: **nitpick (blocking):** Can you update your asserts in this test, currently they are in the wrong order. They are supposed to be `assert EXPECTED, ACTUAL` where you have it flipped.

https://guides.rubyonrails.org/testing.html#available-assertions

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,102 @@
...
+        raise SampleMetadataUpdateError,
+              I18n.t('services.samples.metadata.empty_metadata', sample_name: @sample.name)
+      end
+
+      def transform_metadata_keys
+        # Without transforming keys, issues with overwritting can occur and multiples of the same key can appear
+        @metadata = @metadata.transform_keys(&:to_s)
+      end
+
+      def metadata_modification
```
ericenns, line 60: **nitpick (blocking):** Can you rename this method to `perform_metadata_update`, I think it is a more descriptive name, and I think you had this originally.

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,102 @@
...
+        update_status = { updated: [], not_updated: [] }
+        @metadata.each do |key, value|
+          if value.blank?
+            if @sample['metadata'].key?(key)
```
ericenns, line 64: **nitpick (blocking):** can you change all instances of `@sample['metadata']` to `@sample.metadata` as that is the normal way to access a models attributes vs the hash like interface your using.

app/services/samples/metadata/update_service.rb
```diff
@@ -0,0 +1,102 @@
...
+              @sample['metadata'].delete(key)
+              @sample['metadata_provenance'].delete(key)
```
ericenns, line 66: **praise:** can you change all instances of `@sample['metadata_provenance']` to `@sample.metadata_provenance` as that is the normal way to access a models attributes vs the hash like interface your using.

ericenns (changes requested):
I couple formatting comments and some testing comments.

test/services/samples/metadata/update_service_test.rb
ChrisHuynh333: Thanks, fixed in 38058670dd23177ebf0267bb3f1c3fa5166647d9

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Fixed in 38058670dd23177ebf0267bb3f1c3fa5166647d9

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Fixed in 38058670dd23177ebf0267bb3f1c3fa5166647d9

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Fixed in 38058670dd23177ebf0267bb3f1c3fa5166647d9

ericenns (approved):
Looks good to me

Issue #246, opened 2023-10-19
Workflow: Preparation Service


ericenns:
## Summary
Since we store `Attachments` without using the original filename on the storage backend we need to prepare our `WorkflowExecution` before we submit it to our WES api.

## Acceptance Criteria
* Create a `WorkflowExecutions::PreparationService`
  * Accept a `workflow_execution` as parameters
  * Validate that the `workflow_execution` is in `STATE_TO_BE_DETERMINED`
  * Create a directory on the storage backend
  * Copy the blobs from the `SamplesWorkflowExecution` `samplesheet_params` to a subdirectory in the created directory
  * Create a `samplesheet.csv` that uses the `samplesheet_params` from each `SamplesWorkflowExecution` and fill in the `sample` column with the `Sample` signed gid (https://github.com/rails/globalid#signed-global-ids) and update the other file params to point to the copied files in the subdirectory
  * Mark the `worfklow_execution` as `prepared`

ericenns:
Completed in #321

PR #321, opened 2023-11-30
Workflow Executions: Preparation Service

ericenns:

## PR acceptance checklist


app/services/workflow_executions/preparation_service.rb
```diff
@@ -0,0 +1,120 @@
...
+        attachments[key] = GlobalID.find(gid)
+      end
+      attachments
+    end
+
+    def generate_input_key(filename, prefix = '')
+      format('%<run_dir>s/%<prefix>s%<filename>s', run_dir: @run_dir, filename:, prefix:)
+    end
+
+    def copy_attachment_to_run_dir(attachment, attachable) # rubocop:disable Metrics/AbcSize
```
JeffreyThiessen, line 74: Line 75-88 only uses `attachment` and not `attachable`, it could be refactored to make it more readable and not trigger the Metrics/AbcSize rule.

I'd say lines 75-88 could be it's own function, or if you want key generation to exist in `copy_attachment_to_run_dir` then make a function for lines 79-88 called `generate_attachment_blob(key, attachment)` that returns `blob`.

app/services/workflow_executions/preparation_service.rb
```diff
@@ -0,0 +1,120 @@
...
+      @samplesheet_headers = %w[sample fastq_1 fastq_2]
+      @samplesheet_rows = []
+    end
+
+    def execute # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
+      # confirm params/permissions
+      # build workflow execution run directory
+      generate_run_dir
+
+      # does it make sense to rename the SamplesWorkflowExecution model to something better
```
JeffreyThiessen, line 21: I think this depends on how we want to think of the SamplesWorkflowExecutions. Right now it's just a table join with extra samplesheet params and the name works fine for that, if we want it to handle more data or extra logic than a rename would make sense.

app/services/workflow_executions/preparation_service.rb
```diff
@@ -0,0 +1,120 @@
...
+
+        attachments.each do |key, attachment|
+          samplesheet_params[key] = copy_attachment_to_run_dir(attachment, sample_workflow_execution)
+        end
+
+        @samplesheet_rows << samplesheet_params
+      end
+
+      # persist samplesheet in run dir
+      @workflow_execution.inputs.attach(io: samplesheet_file, key: generate_input_key('samplesheet.csv'),
```
JeffreyThiessen, line 35: Since we are creating and managing a new csv file `samplesheet.csv` which contains the sample sheet data we should name it such that it won't accidentally be confused with the Illumina formatted csv for fastq conversion `SampleSheet.csv`.

I'd say name it something like `samplesheet_params.csv` or `samplesheet_data.csv`

app/services/workflow_executions/preparation_service.rb
```diff
@@ -0,0 +1,120 @@
...
+        copied_blob.compose([attachment.file.key])
+        copied_blob.save!
+      end
+
+      attachable.inputs.attach(blob.signed_id)
+
+      blob_to_service_path(blob)
+    end
+
+    def blob_to_service_path(blob)
```
JeffreyThiessen, line 95: Will this blob to service path conversion be used other places too? Maybe this function should be in a active storage helper file somewhere?

JeffreyThiessen (changes requested):
The code looks good. I just have a few questions/comments.

test/services/workflow_executions/preparation_service_test.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+
+      assert_equal 1, @workflow_execution.inputs.size
+      assert_equal 1, @workflow_execution.samples_workflow_executions.first.inputs.size
+
+      assert @workflow_execution.workflow_params.key? '--input'
+      assert @workflow_execution.workflow_params.key? '--outdir'
+      assert_equal @workflow_execution.inputs.first.blob.key, @workflow_execution.workflow_params['--input']
+
+      assert_equal 'prepared', @workflow_execution.state
```
JeffreyThiessen, line 24: We should test the state before the preparation is executed as listed in https://github.com/phac-nml/irida-next/issues/246

We don't have a state for it right now but we could just add an assert state not prepared.

app/services/workflow_executions/preparation_service.rb
ericenns: I suspect it will only be used in this place, if eventually other places need it then it can be refactored later.

app/services/workflow_executions/preparation_service.rb
ericenns: This file is just created to run the pipeline in the `CompletionService` we will remove it, it just temporary and we won't be storing Illumina samplesheets so I don't see any chance of confusion.

deepsidhu85, line 35: Assign `generate_input_key('samplesheet.csv')` since it's being used here and on line 40 below to a variable and update to use this variable

app/services/workflow_executions/preparation_service.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
+
+require 'csv'
+require 'tempfile'
+
+module WorkflowExecutions
```
deepsidhu85, line 6: Can you add some basic error checking for now such as the samplesheet isn't empty, directory was created successfully, attachments copied successfully etc

deepsidhu85 (changes requested):
This looks good just have a few comments below before testing

app/services/workflow_executions/preparation_service.rb
ericenns: Fixed in 6116512b

app/services/workflow_executions/preparation_service.rb
ericenns: I removed this comment for now, I think a rename will likely come into play when we decide how to store results for a Sample as this model reference might be the correct one.

app/services/workflow_executions/preparation_service.rb
ericenns: Not sure what you are asking for? If adding it to the actual service this would be a pain as I would have to use the `ActiveStorage::Service` directly and handle it differently for each service. Or are you suggesting for the test?

test/services/workflow_executions/preparation_service_test.rb
ericenns: Fixed in f986ce2d

Issue #247, opened 2023-10-19
Workflow: Submission Service


ericenns:
## Summary
After a `WorfklowExecution` has been prepared. we need to submit it to our WES api.

## Acceptance Criteria
* Create a `WorkflowExecutions::SubmissionService`
  * Accept a `workflow_execution` as parameters
  * Validate that the `workflow_execution` is in `STATE_TO_BE_DETERMINED`
  * Submit the `workflow_execution` to the WES api
  * Store the returned `run_id`
  * Mark the `worfklow_execution` as `prepared`

ericenns:
Closed via #325

PR #325, opened 2023-12-14
Workflow Executions: Submission Service

ericenns:

This PR adds in the WorkflowExecutions::PreparationService as described in https://github.com/phac-nml/irida-next/issues/247

This first checks that the `WorkflowExecution` is in the `prepared` state, if not it returns false.

If `prepared` it continues on and submits to WES run_workflow endpoint, and stores the `run_id` and updates the state `submitted`.

## PR acceptance checklist

### DIFF
--- a/app/services/workflow_executions/submission_service.rb
+++ b/app/services/workflow_executions/submission_service.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module WorkflowExecutions
+  # Service used to Prepare a WorkflowExecution
+  class SubmissionService < BaseService
+    def initialize(workflow_execution, wes_connection, user = nil, params = {})
+      super(user, params)
+      @workflow_execution = workflow_execution
+      @wes_client = Integrations::Ga4ghWesApi::V1::Client.new(conn: wes_connection)
+    end
+
+    def execute
+      run = @wes_client.run_workflow(**@workflow_execution.as_wes_params)
+
+      @workflow_execution.run_id = run[:run_id]
+
+      # mark workflow execution as submitted
+      @workflow_execution.state = 'submitted'
+
+      @workflow_execution.save
+    end
+  end
+end
--- a/app/services/workflow_executions/submission_service.rb
+++ b/app/services/workflow_executions/submission_service.rb
@@ -10,6 +10,8 @@ module WorkflowExecutions
     end
 
     def execute
+      return false unless @workflow_execution.prepared?
+
       run = @wes_client.run_workflow(**@workflow_execution.as_wes_params)
 
       @workflow_execution.run_id = run[:run_id]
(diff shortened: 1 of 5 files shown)


JeffreyThiessen (commented):
Overall the code looks good.

I think we should consider changing `state` from a `string` to an `enum` in the database model.
By defining it as an `enum`, we will have access to dynamic methods like `workflow_execution.prepared?` and `workflow_execution.submitted?`, and can set the state with the `!` equivalent methods. 
It also would prevent string comparison which can be annoying to work with.

Article on this topic: https://blog.saeloun.com/2022/01/05/how-to-use-enums-in-rails/

ericenns:
> Overall the code looks good.
> 
> I think we should consider changing `state` from a `string` to an `enum` in the database model. By defining it as an `enum`, we will have access to dynamic methods like `workflow_execution.prepared?` and `workflow_execution.submitted?`, and can set the state with the `!` equivalent methods. It also would prevent string comparison which can be annoying to work with.
> 
> Article on this topic: https://blog.saeloun.com/2022/01/05/how-to-use-enums-in-rails/

I agree, I just haven't fully though what the states should be so just holding off on that for now. The enums will certainly make it alot easier to use later on though.

JeffreyThiessen (approved):
Looks good!

app/services/workflow_executions/submission_service.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module WorkflowExecutions
+  # Service used to Prepare a WorkflowExecution
+  class SubmissionService < BaseService
+    def initialize(workflow_execution, wes_connection, user = nil, params = {})
+      super(user, params)
+      @workflow_execution = workflow_execution
+      @wes_client = Integrations::Ga4ghWesApi::V1::Client.new(conn: wes_connection)
```
deepsidhu85, line 9: I gather that we will be using `Integrations::Ga4ghWesApi::V1::Client.new(conn: wes_connection)` in multiple services after this one. I think it would be better to have this in as a common method to which we can pass in the `wes_connection`. This way if the version changes we only need to update it in one place

app/services/workflow_executions/submission_service.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def execute
+      return false unless @workflow_execution.prepared?
```
deepsidhu85, line 13: Not sure if this is possible, but can we check if the client was initialized successfully and if it wasn't throw an error? We can always add that stuff in later once we know exactly where/how the service will be called from

deepsidhu85 (changes requested):
Looks good just a couple of comments below

app/services/workflow_executions/submission_service.rb
ericenns: I agree, but let's see how the cards play first.

app/services/workflow_executions/submission_service.rb
ericenns: I think an exception might be raised in this case, but currently no handling here. That handling can be added later as we actually see what happens.

app/services/workflow_executions/submission_service.rb
deepsidhu85: Lol yeah sounds good!

deepsidhu85 (approved):

PR #328, opened 2023-12-14
Workflow Executions Cancellation Service

### LINKED ISSUE #248: Workflow: Cancellation Service
## Summary
After a `WorfklowExecution` has been submitted, a user may want to cancel it. This would first occur by a user clicking a button and setting the state of the `WorkflowExecution` to `cancelling`, after that a call would occur to a new `WorkflowExecutions::CancellationService`, this would use the WES api to cancel the execution and then set the state to `cancelled`.

## Acceptance Criteria
* Create a `WorkflowExecutions::CancellationService`
  * Accept a `workflow_execution` as parameters
  * Validate that the `workflow_execution` is in `cancelling` state
  * Use the WES api to cancel the `workflow_execution` using the `run_id` attribute.
  * Mark the `worfklow_execution` as `cancelled`

JeffreyThiessen:
Fixes #248 
This PR adds in the WorkflowExecutions::CencellationService as described in #248 

This first checks that `WorkflowExecution` is in the `submitted` state, if not it returns false.
In the future if we have separate states for submitted and running, this will need to be changed.

If `submitted` it continues on and sends a cancel request to the WES `runs/<id>/cancel` endpoint, then on success it updates the state to `cancelled`.

## PR acceptance checklist


app/models/workflow_execution.rb
```diff
@@ -23,6 +23,14 @@ def prepared?
     state == 'prepared'
   end
 
+  def submitted?
+    state == 'submitted'
+  end
+
+  def cancelled?
```
deepsidhu85, line 30: This should be `canceled?` as well as `state == 'canceled'` This will need to be updated in the service as well.

app/services/workflow_executions/cancellation_service.rb
```diff
@@ -0,0 +1,24 @@
...
+      super(user, params)
+      @workflow_execution = workflow_execution
+      @wes_client = Integrations::Ga4ghWesApi::V1::Client.new(conn: wes_connection)
+    end
+
+    def execute
+      return false unless @workflow_execution.submitted?
+
+      # throws exception if failed
+      @wes_client.cancel_run(@workflow_execution.run_id)
```
deepsidhu85, line 16: If an exeception is thrown, we should attach the exeception message to the @workflow_execution so that we can return the error. This service isn't going to be kicked off as a job correct? It will just be the user clicking a button to cancel which will fire off immediately?

deepsidhu85 (changes requested):
This looks good, just have a couple of comments below

app/models/workflow_execution.rb
deepsidhu85: Decided to leave this as is

JeffreyThiessen: Canadian spelling vs American spelling. I'm not sure if there's a reason to prefer one over the other in code. Maybe @ericenns has input?

app/services/workflow_executions/cancellation_service.rb
deepsidhu85: @ericenns Thoughts on this?

JeffreyThiessen: All of the `wes_client` calls will throw appropriate exceptions from `lib/integrations/api_exceptions.rb`. I added the note in this PR but all of the other services that use the ga4gh wes api will throw exceptions on failure.

deepsidhu85: Sounds good thanks. Is it possible to show the exception in the test `cancel unsubmitted workflow_execution`?

deepsidhu85: Will add in the exception handling in a separate PR as this applies to all the services where the wes client is used

app/services/workflow_executions/cancellation_service.rb
```diff
@@ -0,0 +1,24 @@
...
+  class CancellationService < BaseService
+    def initialize(workflow_execution, wes_connection, user = nil, params = {})
+      super(user, params)
+      @workflow_execution = workflow_execution
+      @wes_client = Integrations::Ga4ghWesApi::V1::Client.new(conn: wes_connection)
+    end
+
+    def execute
+      return false unless @workflow_execution.submitted?
+
```
deepsidhu85, line 14: As per the issue #248 we want to validate that the workflow is in the `cancelling` state before making the call to the WES api to actually cancel the workflow. We will need to update the test as well to assert that it's in the `cancelling` state before calling the cancellation service

deepsidhu85 (changes requested):
One last comment

app/services/workflow_executions/cancellation_service.rb
JeffreyThiessen: Fixed in https://github.com/phac-nml/irida-next/pull/328/commits/eb38099002599a4df1ae9329fb92ab779c612d95

deepsidhu85 (approved):
Looks good to me!

ericenns (approved):
Looks good to me!

Issue #334, opened 2023-12-19
Project Sample Metadata - Update Metadata Summary


deepsidhu85:
# Summary
As a user I want to be able to see a summary of the metadata that is in my projects, as well as, at the group level. The db representation has been setup on the `Namespace` model but we only want to target the group model and the project_namespace model. The summary will just contain the metadata fields keys and the count of how many times that metadata field key appears in the group and/or project. **Note: for this first part we will not be adding in linked/unlinked namespace metadata summaries**

# Solution
1) Add a method to the group model which will update the metadata summary for a group
2) Add a method to the project namespace model which will update the metadata summary for a project
3) The above methods should be called on the following actions: `Deleting metadata from a sample`, `Adding metadata to a sample`, `Transferring a groups and/or projects`, `Transferring samples`

Here is a resource that can be used to help figure this out https://apidock.com/rails/ActiveModel/Dirty/previous_changes 

# Acceptance Criteria
The metadata summary for a group and/or project is updated correctly and has the correct fields and counts.
The metadata summary follows this format
`{
  key1: 3, key2: 1,......
}`

Testing added to test the following:

Project under a group
Projects under both groups and subgroups

PR #339, opened 2023-12-20
Metadata Summary Update by Metadata::UpdateService

ChrisHuynh333:
As part of [issue 334](https://github.com/phac-nml/irida-next/issues/334), this PR implements `group` and `project` `metadata_summary` updating when metadata for a sample is changed via the `Samples::Metadata::UpdateServce`. 

When `sample_metadata` is updated by the `UpdateService`, the `project.namespace` and any `group` parents will have a count updated to track how many samples contain each metadata_field in their respective `update_summary`.

Note: updating `metadata_summary` by `project/group/sample` `transfer` will be implemented in a separate PR.
Note: Although not too clear in this PR, `subtract_from_metadata_summary` and `add_to_metadata_summary` were refactored to include the `update_by_one` argument in order for these functions to be usable by all `metadata_summary` updating (ie: transfers and deletions). See [PR 350](https://github.com/phac-nml/irida-next/pull/350/files) for context.

Either create or utilize `groups/projects/samples` within `db:seed` with empty `metadata_summary` that would have a 'master' group, with separate descendent groups and/or projects.

I will use the following:
Bacillus 
     V
Bacillus cereus > Outbreak 2021 (var: `project1`) > Sample 10 (var: `sample1`)
             |
              ---------------- > Outbreak 2022 (var: `project2`) > Sample 10 (var: `sample2`)

1. In the rails console, assign the above variables `sample1`, `sample2`, `project1`, `project2`.
2. Assign a `user`  with permission to edit the above.
3. Test adding metadata
a) Assign `params1 = {'metadata' => {'metadatafield1' => 'value1'}}`
b) Run `Samples::Metadata::UpdateService.new(project1, sample1, user, params1).execute`
c) Check that Outbreak 2021, Bacillus cereus, and Bacillus now have an `metadata_summary = {'metadatafield1': 1}`. Outbreak 2022 `metadata_summary` should still be empty.
d) Run `Samples::Metadata::UpdateService.new(project2, sample2, user, params1).execute`
e) Check Bacillus cereus and Bacillus now have an `metadata_summary = {'metadatafield1': 2}`, and Outbreak 2021 and Outbreak 2022 `metadata_summary = {'metadatafield1': 1}` ` .

4. Test adding and deleting metadata
a) Assign `params2 = {'metadata' => {'metadatafield1' => '', 'metadatafield2' => 'value2'}}`
b) Run `Samples::Metadata::UpdateService.new(project1, sample1, user, params2).execute`
c) Check that Bacillus cereus, and Bacillus now have an `metadata_summary = {'metadatafield1': 1, 'metadatafield2': 1}`, Outbreak 2021 `metadata_summary = {'metadatafield2': 1}` and Outbreak 2022 `metadata_summary = {'metadatafield1': 1}`

## PR acceptance checklist

### DIFF
--- a/app/controllers/projects/samples/metadata_controller.rb
+++ b/app/controllers/projects/samples/metadata_controller.rb
@@ -9,8 +9,8 @@ module Projects
         respond_to do |format|
           metadata_fields = ::Samples::Metadata::UpdateService.new(@project, @sample, current_user,
                                                                    metadata_params).execute
-          changed_metadata_fields = metadata_fields[:added] + metadata_fields[:updated] + metadata_fields[:deleted]
-          if changed_metadata_fields.count.positive?
+          modified_metadata = metadata_fields[:added] + metadata_fields[:updated] + metadata_fields[:deleted]
+          if modified_metadata.count.positive?
             flash[:success] =
               t('.success', metadata_fields: metadata_fields[:updated].join(', '), sample_name: @sample.name)
           end
--- a/app/controllers/projects/samples/metadata_controller.rb
+++ b/app/controllers/projects/samples/metadata_controller.rb
@@ -9,7 +9,8 @@ module Projects
         respond_to do |format|
           metadata_fields = ::Samples::Metadata::UpdateService.new(@project, @sample, current_user,
                                                                    metadata_params).execute
-          if !metadata_fields.nil? && metadata_fields[:updated].count.positive?
+          changed_metadata_fields = metadata_fields[:added] + metadata_fields[:updated] + metadata_fields[:deleted]
+          if changed_metadata_fields.count.positive?
             flash[:success] =
               t('.success', metadata_fields: metadata_fields[:updated].join(', '), sample_name: @sample.name)
           end
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -226,7 +226,7 @@ class Namespace < ApplicationRecord # rubocop:disable Metrics/ClassLength
     Route.restore(Route.only_deleted.find_by(source_id: id).id, recursive: true)
   end
 
-  def subtract_from_metadata_summary(namespaces, metadata, update_by_one)
+  def subtract_from_metadata_summary_count(namespaces, metadata, update_by_one)
     namespaces.each do |namespace|
       metadata.each do |metadata_field, value|
         value = 1 if update_by_one
@@ -240,7 +240,7 @@ class Namespace < ApplicationRecord # rubocop:disable Metrics/ClassLength
     end
   end
 
-  def add_to_metadata_summary(namespaces, metadata, update_by_one)
+  def add_to_metadata_summary_count(namespaces, metadata, update_by_one)
     namespaces.each do |namespace|
       metadata.each do |metadata_field, value|
         value = 1 if update_by_one
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -225,4 +225,32 @@ class Namespace < ApplicationRecord # rubocop:disable Metrics/ClassLength
   def restore_routes
     Route.restore(Route.only_deleted.find_by(source_id: id).id, recursive: true)
   end
+
+  def subtract_from_metadata_summary(namespaces, metadata, update_by_one)
+    namespaces.each do |namespace|
+      metadata.each do |metadata_field, value|
+        value = 1 if update_by_one
+        if namespace.metadata_summary[metadata_field] == value
+          namespace.metadata_summary.delete(metadata_field)
+        else
+          namespace.metadata_summary[metadata_field] -= value
+        end
+      end
+      namespace.save
+    end
+  end
+
+  def add_to_metadata_summary(namespaces, metadata, update_by_one)
+    namespaces.each do |namespace|
+      metadata.each do |metadata_field, value|
+        value = 1 if update_by_one
+        if namespace.metadata_summary.key?(metadata_field)
+          namespace.metadata_summary[metadata_field] += value
+        else
+          namespace.metadata_summary[metadata_field] = value
+        end
+      end
+      namespace.save
+    end
+  end
 end
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -248,15 +248,17 @@ class Namespace < ApplicationRecord # rubocop:disable Metrics/ClassLength
   # sample_transfer (original parents), sample_deletion, and update_service when metadata is deleted from a sample.
   def subtract_from_metadata_summary_count(namespaces, metadata, update_by_one)
     namespaces.each do |namespace|
-      metadata.each do |metadata_field, value|
-        value = 1 if update_by_one
-        if namespace.metadata_summary[metadata_field] == value
-          namespace.metadata_summary.delete(metadata_field)
-        else
-          namespace.metadata_summary[metadata_field] -= value
+      namespace.with_lock do
+        metadata.each do |metadata_field, value|
+          value = 1 if update_by_one
+          if namespace.metadata_summary[metadata_field] == value
+            namespace.metadata_summary.delete(metadata_field)
+          else
+            namespace.metadata_summary[metadata_field] -= value
+          end
         end
+        namespace.save
       end
-      namespace.save
     end
   end
 
@@ -264,15 +266,17 @@ class Namespace < ApplicationRecord # rubocop:disable Metrics/ClassLength
   # and update_service when metadata is added to a sample.
   def add_to_metadata_summary_count(namespaces, metadata, update_by_one)
     namespaces.each do |namespace|
-      metadata.each do |metadata_field, value|
-        value = 1 if update_by_one
-        if namespace.metadata_summary.key?(metadata_field)
-          namespace.metadata_summary[metadata_field] += value
-        else
-          namespace.metadata_summary[metadata_field] = value
+      namespace.with_lock do
+        metadata.each do |metadata_field, value|
+          value = 1 if update_by_one
+          if namespace.metadata_summary.key?(metadata_field)
+            namespace.metadata_summary[metadata_field] += value
+          else
+            namespace.metadata_summary[metadata_field] = value
+          end
         end
+        namespace.save
       end
-      namespace.save
     end
   end
 end
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -226,6 +226,26 @@ class Namespace < ApplicationRecord # rubocop:disable Metrics/ClassLength
     Route.restore(Route.only_deleted.find_by(source_id: id).id, recursive: true)
   end
 
+  # The add and subtract_from_metadata_summary count functions are used by all update_metadata_summary functions
+  # (update_service, namespace_transfer, namespace_deletion, sample_transfer, sample_deletion)
+  # to update parental namespace metadata_summary attributes.
+  #
+  # When a sample is transferred, deleted, or updated, namespaces are the non-user parental namespaces, metadata is
+  # equal to sample.metadata, and update_by_one is assigned true. Because we are passing actual metadata values
+  # (ie: sample.metadata = {metadatafield1: metadatavalue1}), we assign true to update_by_one and value = 1 when
+  # calling these functions because we will update the samples' parents by a count of 1, regardless of what the actual
+  # metadata value is (ie: sample.metadata above with a value of metadatavalue1 will result in
+  # metadata_summary['metadatafield1'] += 1 for all parental namespaces).
+  #
+  # When namespaces are transferred or deleted, namespaces are the non-user parental namespaces, metadata is equal to
+  # the namespace.metadata_summary, and update_by_one is false. With namespaces, we will pass metadata_summary
+  # (ie: project.namespace.metadata_summary = {metadatafield1: 5, metadatafield2: 10}), and we can iterate over the
+  # summary and utilize the hash values to update all parental namespaces, so setting value = 1 is unnecessary and
+  # we assign false to update_by_one (ie: all parental namespaces will have their metadata_summary updated by
+  # 5 for metadatafield1 and 10 for metadatafield2).
+  #
+  # Subtract will be called for namespace_transfer (original parents), namespace_deletion,
+  # sample_transfer (original parents), sample_deletion, and update_service when metadata is deleted from a sample.
   def subtract_from_metadata_summary_count(namespaces, metadata, update_by_one)
     namespaces.each do |namespace|
       metadata.each do |metadata_field, value|
@@ -240,6 +260,8 @@ class Namespace < ApplicationRecord # rubocop:disable Metrics/ClassLength
     end
   end
 
+  # Add will be called for namespace_transfer (new parents), sample_transfer (new parents),
+  # and update_service when metadata is added to a sample.
   def add_to_metadata_summary_count(namespaces, metadata, update_by_one)
     namespaces.each do |namespace|
       metadata.each do |metadata_field, value|
--- a/app/models/namespaces/project_namespace.rb
+++ b/app/models/namespaces/project_namespace.rb
@@ -34,5 +34,42 @@ module Namespaces
     def self.sti_name
       'Project'
     end
+
+    def update_metadata_summary(metadata_to_delete, metadata_to_add)
+      namespaces_to_update = self_and_parents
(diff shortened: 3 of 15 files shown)


app/models/namespace.rb
```diff
@@ -225,4 +225,32 @@ def validate_nesting_level
...
   end
+
+  def subtract_from_metadata_summary(namespaces, metadata, update_by_one)
+    namespaces.each do |namespace|
+      metadata.each do |metadata_field, value|
+        value = 1 if update_by_one
+        if namespace.metadata_summary[metadata_field] == value
+          namespace.metadata_summary.delete(metadata_field)
+        else
+          namespace.metadata_summary[metadata_field] -= value
```
ksierks, line 236: Question: Is it possible for update_by_one to be false and the value not an integer?

app/models/namespace.rb
```diff
@@ -225,4 +225,32 @@ def validate_nesting_level
...
+      namespace.save
+    end
+  end
+
+  def add_to_metadata_summary(namespaces, metadata, update_by_one)
+    namespaces.each do |namespace|
+      metadata.each do |metadata_field, value|
+        value = 1 if update_by_one
+        if namespace.metadata_summary.key?(metadata_field)
+          namespace.metadata_summary[metadata_field] += value
```
ksierks, line 248: Question: Same as above comment. Is it possible for update_by_one to be false and the value not an integer?

app/models/namespaces/project_namespace.rb
```diff
@@ -34,5 +34,19 @@ def of_ancestors_and_self
     def self.sti_name
       'Project'
     end
+
+    def self_and_parents
```
ksierks, line 38: Question: Should this be returning all ancestors that are not UserNamespaces?

app/models/namespace.rb
ChrisHuynh333: So this was my design decision and attempt at making the add and subtract metadata summary as generic as possible so that it can be used first by sample metadata updating as shown in this PR, and then also sample deletion/transfer and group/project transfer/deletion in the next PR. 
So anything regarding updating a sample's metadata, where we're dealing with actual sample metadata (ie add `metadatafield1: value1` to a sample) will have `update_by_one` as `true`, because you're right, default `hash value` will possibly not be an integer.
But then when we're dealing with any groups or projects for transfer/deletion, all we care about is their `metadata_summary` (ie: `metadatafield1: 5`), in which case, `update_by_one` is false, and the hash value will be an integer.

I actually initially had separate functions to handle these, but figured they're so similar that I put them together with the `update_by_one` being the varying factor. For readability and simplifying understanding in the future, we could definitely go back to having the functions separated.
Hope that all makes sense

app/models/namespace.rb
ChrisHuynh333: See above!

app/models/namespace.rb
ksierks: Ok. I got confused with update_by_one being false and setting the value. I was also thinking there might be a case were we don't want to add or subtract, but update. Sorry and thanks.

app/services/samples/metadata/update_service.rb
```diff
@@ -28,6 +28,8 @@ def execute
 
         @sample.save
 
+        update_metadata_summary
```
deepsidhu85, line 31: So after going through the code I think we have a small issue. Right now the way it stands this will run the update_metadata_summary regardless of if a metadata field was `added` or `deleted`. This will also run if only values were `updated` which I don't think we want.

A possible solution is add a flag to the update_status hash which will be set to `true` if an addition or deletion action happened. Then in the `update_metadata_summary` we can return right away if the flag is false.

app/models/namespaces/project_namespace.rb
```diff
@@ -34,5 +34,19 @@ def of_ancestors_and_self
...
+      namespaces = [self]
+      namespaces += parent.self_and_ancestors unless parent.type == 'User'
+      namespaces
+    end
+
+    def update_metadata_summary_by_update_service(metadata_to_subtract, metadata_to_add)
```
deepsidhu85, line 44: I think this method can just be called `update_metadata_summary`. Also I think we need to rename these variables as they seem to imply that metadata is being removed or added, instead of the counts

app/models/namespaces/project_namespace.rb
```diff
@@ -34,5 +34,19 @@ def of_ancestors_and_self
...
+      namespaces_to_update = self_and_parents
```
deepsidhu85, line 45: I don't think we need a separate method for self_and_parents. We already have a query to get the ancestors. You should be able to just call `namespaces_to_update = [self] + parent.self_and_ancestors`

app/services/samples/metadata/update_service.rb
```diff
@@ -97,6 +99,22 @@ def fields_not_updated(metadata_fields_update_status)
...
+      def update_metadata_summary
+        return unless @sample.previous_changes['metadata']
+
+        old_metadata = @sample.previous_changes['metadata'][0]
+        new_metadata = @sample.previous_changes['metadata'][1]
+
+        # Checks which keys are overlapping after metadata changes.
+        # Keys that are overlapping will not affect summary counts, therefore can be deleted from both hashes
+        old_metadata.each do |metadata_field, _v|
+          if new_metadata.key?(metadata_field)
```
deepsidhu85, line 112: This should only happen if only values were changed correct? Otherwise there should be no overlap? If so the comment above to only run this method if an `addition` or `deletion` are made, then this loop shouldn't be required anymore

deepsidhu85 (changes requested):
So testing this out in the rails console it appears to work as expected. However there are a few issues below that I commented on

app/models/namespaces/project_namespace.rb
ChrisHuynh333: So the reason its `...by_update_service` is because in the next PR, I have like 4 other `update_metadata_summary` functions (namespace_transfer, namespace_deletion, sample_deletion etc.). They're all different enough where I believe they require their own functions (definitely an area I would love review and a fresh set of eyes). So can we make note of this point but keep it until the next PR review?
Agreed on the argument var names, will update in next commit.

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Changes in 3ecf7d76649ad401bdaf20be265bd179130197ce with following notes:
- `update_status` now contains a key/value flags for `metadata_was_added` and `metadata_was_deleted`, with default value `false` and will update to `true` if those actions happen.
- `update_metadata_summary` will only occur if either of those are true (ie: skip if metadata was only `updated`)
- tests updated accordingly

app/services/samples/metadata/update_service.rb
ChrisHuynh333: This loop has now been removed in 7d1d6d693cca394dbacd24cdd02a4e1a8593c0e7
See comment about service overhaul.

ChrisHuynh333:
As per @deepsidhu85 comments, I made a decent overhaul to the `Metadata::UpdateService` which includes changes to how data is passed to update `update_metadata_summary`

- I got rid of the `previous_changes` check and `loop` deleting overlapping keys
- I've removed the `metadata_field_update_status` hash and now initiate a `metadata_changes` hash in the initializer. `metadata_changes` now contains empty lists for keys `added` `updated` `deleted` and `not_updated`. I now append all `metadata_fields` to the appropriate `metadata_changes` key based on the action, since I'm already iterating over all the `metadata_fields` passed to the `update_service` in `perform_metadata_update` . This allows, not only do we get rid of the terrible `metadata_fields_update_status` var name, but to also get rid of `previous_changes`, and needing to loop over the before/after changes and I can just pass the `added`/`deleted` key/values from `metadata_changes` to the `update_metadata_summary`.
- `not_updated` is still used to add errors for eventual flash msges, `updated`, with `added` and `deleted`, is still used for eventual success flash msges

app/services/samples/metadata/update_service.rb
ChrisHuynh333: Logic was changed in 7d1d6d693cca394dbacd24cdd02a4e1a8593c0e7, see comment below for additional context.

We now check if any `metadata_fields` were `added` or `deleted`

app/models/namespaces/project_namespace.rb
ChrisHuynh333: Thanks for this, much better. Changed in 398dc863c9111dce05cac4a08fa3f29ea205ad0e

app/models/namespaces/project_namespace.rb
ChrisHuynh333: Changed arguments to `deleted_metadata` and `added_metadata` and the functions that update the summary counts to `subtract_from_metadata_summary_count` and `add_to_metadata_summary_count` in 398dc863c9111dce05cac4a08fa3f29ea205ad0e

app/models/namespace.rb
```diff
@@ -225,4 +225,32 @@ def validate_nesting_level
   def restore_routes
     Route.restore(Route.only_deleted.find_by(source_id: id).id, recursive: true)
   end
+
+  def subtract_from_metadata_summary_count(namespaces, metadata, update_by_one)
```
JeffreyThiessen, line 229: Could you please add a function comment block to `subtract_from_metadata_summary_count` with an explanation of the different fields and how `update_by_one` works/is used. Your explanation to @ksierks 's question makes sense, but I think we should add that info to the function.

app/models/namespace.rb
```diff
@@ -225,4 +225,32 @@ def validate_nesting_level
...
+          namespace.metadata_summary.delete(metadata_field)
+        else
+          namespace.metadata_summary[metadata_field] -= value
+        end
+      end
+      namespace.save
+    end
+  end
+
+  def add_to_metadata_summary_count(namespaces, metadata, update_by_one)
```
JeffreyThiessen, line 243: This function too as above

app/services/samples/metadata/update_service.rb
```diff
@@ -58,44 +61,48 @@ def transform_metadata_keys
...
           @sample.metadata_provenance[key] =
             @analysis_id.nil? ? { source: 'user', id: current_user.id } : { source: 'analysis', id: @analysis_id }
           @sample.metadata[key] = value
-          true
         end
       end
 
-      def fields_not_updated(metadata_fields_update_status)
-        metadata_fields_not_updated = metadata_fields_update_status[:not_updated]
+      def assign_errors_to_not_updated_fields
```
JeffreyThiessen, line 92: Since this only gets called if there are `not_updated` keys found in the `assign_metadata_to_sample` function, I think this function should be renamed. It suggests that errors raised by the other validation functions (like sample not on project) would get added to the not updated list.

Something like `handle_not_updated_fields` would be good. And probably a good idea to add a function comment saying it only handles the not updated keys found by `assign_metadata_to_sample` function.

JeffreyThiessen (changes requested):
Overall look really good. I like the exhaustive test cases. Code looks to be working as intended from my testing.

Just a few comments on function  documentation.

app/models/namespace.rb
ChrisHuynh333: Added in 422c2be245bdf39d1f2475382065ee11a0c855e5. It's quite long since I wanted to ensure whenever we come back to it in the future, it's clear what the thought process was. Let me know if it's too much.

app/models/namespace.rb
ChrisHuynh333: Added in 422c2be245bdf39d1f2475382065ee11a0c855e5

app/services/samples/metadata/update_service.rb
ChrisHuynh333: I like this a lot better. Changed in 422c2be245bdf39d1f2475382065ee11a0c855e5

JeffreyThiessen (approved):
Looks good

app/controllers/projects/samples/metadata_controller.rb
```diff
@@ -9,7 +9,8 @@ def update # rubocop:disable Metrics/AbcSize
         respond_to do |format|
           metadata_fields = ::Samples::Metadata::UpdateService.new(@project, @sample, current_user,
                                                                    metadata_params).execute
-          if !metadata_fields.nil? && metadata_fields[:updated].count.positive?
+          changed_metadata_fields = metadata_fields[:added] + metadata_fields[:updated] + metadata_fields[:deleted]
```
deepsidhu85, line 12: Can we rename this variable to `modified_metadata`?

deepsidhu85 (changes requested):
This looks good to me. Just one more comment about the code. Still need to review the testing

test/fixtures/members.yml
```diff
@@ -368,6 +368,48 @@ group_eleven_member_user_26:
   namespace_id: <%= ActiveRecord::FixtureSet.identify(:group_eleven) %>
   access_level: <%= Member::AccessLevel::OWNER %>
 
+group_twelve_member_john_doe:
+  user_id: <%= ActiveRecord::FixtureSet.identify(:john_doe) %>
+  created_by_id: <%= ActiveRecord::FixtureSet.identify(:john_doe) %>
+  namespace_id: <%= ActiveRecord::FixtureSet.identify(:group_twelve) %>
+  access_level: <%= Member::AccessLevel::OWNER %>
+
+subgroup_twelve_a_member_john_doe:
```
deepsidhu85, line 377: Since the sub groups/projects under  group_twelve all have the same parent ancestor, you only need to have the john doe added as a member to the top level group and the subgroups and projects will inherit the membership. Only if you wanted to give the member a higher access level then they have in the parent group would you need to add a direct member to the subgroups/projects

deepsidhu85 (changes requested):
Just have one comment for the test fixtures

test/fixtures/members.yml
ChrisHuynh333: Makes sense! Fixed in 3a93671162a425316846ea3c51595790644eceff

app/controllers/projects/samples/metadata_controller.rb
ChrisHuynh333: Fixed in 97f16f77456fa6b0b755930801a23fc43a3670d6

deepsidhu85 (approved):
Looks good to me!

app/models/namespace.rb
```diff
@@ -225,4 +225,54 @@ def validate_nesting_level
...
+  # the namespace.metadata_summary, and update_by_one is false. With namespaces, we will pass metadata_summary
+  # (ie: project.namespace.metadata_summary = {metadatafield1: 5, metadatafield2: 10}), and we can iterate over the
+  # summary and utilize the hash values to update all parental namespaces, so setting value = 1 is unnecessary and
+  # we assign false to update_by_one (ie: all parental namespaces will have their metadata_summary updated by
+  # 5 for metadatafield1 and 10 for metadatafield2).
+  #
+  # Subtract will be called for namespace_transfer (original parents), namespace_deletion,
+  # sample_transfer (original parents), sample_deletion, and update_service when metadata is deleted from a sample.
+  def subtract_from_metadata_summary_count(namespaces, metadata, update_by_one)
+    namespaces.each do |namespace|
```
ericenns, line 250: **nitpick (blocking):** Can you add a `namespace.with_lock do` block inside so that a namespace is locked for update while this is happening so that we don't have conflicts from simultaneous updates. https://guides.rubyonrails.org/active_record_querying.html#pessimistic-locking

app/models/namespace.rb
```diff
@@ -225,4 +225,54 @@ def validate_nesting_level
...
+        end
+      end
+      namespace.save
+    end
+  end
+
+  # Add will be called for namespace_transfer (new parents), sample_transfer (new parents),
+  # and update_service when metadata is added to a sample.
+  def add_to_metadata_summary_count(namespaces, metadata, update_by_one)
+    namespaces.each do |namespace|
```
ericenns, line 266: **nitpick (blocking):** Same as comment above.

ericenns (changes requested):
This generally looks good, just a comment about database locking which is needed here.

app/models/namespace.rb
ChrisHuynh333: Added in c945fbabc9577fb847d931c897f6c42ed635763e

app/models/namespace.rb
ChrisHuynh333: Added in c945fbabc9577fb847d931c897f6c42ed635763e

ericenns (approved):
Looks good to me!

PR #358, opened 2024-01-08
Attachments listing update

ChrisHuynh333:
This PR prevents non-owners of samples to delete attachments by hiding the `Delete` button in the row under the `Action` header, as well as the `Delete Files` button alongside the `Upload/Concatenate Files` buttons, which is used to delete multiple attachments.

The buttons are hidden under the `sample destroy_attachment?` policy. The `Attachments::DestroyService` and `attachments deletions_controller` have been updated with the same policy.

Non-Owner:
image

Owner:
image

Login as both an owner and non-owner of a sample with attachments, navigate to the attachments page and you can either see or not see the Delete buttons based on the access level.

## PR acceptance checklist

### DIFF
--- a/app/policies/sample_policy.rb
+++ b/app/policies/sample_policy.rb
@@ -2,6 +2,15 @@
 
 # Policy samples authorization
 class SamplePolicy < ApplicationPolicy
+  def destroy_attachment?
+    project = Project.find(record.project_id)
+    return true if project.namespace.parent.user_namespace? && project.namespace.parent.owner == user
+    return true if Member.namespace_owners_include_user?(user, project.namespace) == true
+
+    details[:name] = record.name
+    false
+  end
+
   scope_for :relation, :group_samples do |relation, options|
     group = options[:group]
 
--- a/app/policies/sample_policy.rb
+++ b/app/policies/sample_policy.rb
@@ -2,10 +2,9 @@
 
 # Policy samples authorization
 class SamplePolicy < ApplicationPolicy
-  def destroy_attachment?
-    project = Project.find(record.project_id)
-    return true if project.namespace.parent.user_namespace? && project.namespace.parent.owner == user
-    return true if Member.namespace_owners_include_user?(user, project.namespace) == true
+  def destroy_attachment? # rubocop:disable Metrics/AbcSize
+    return true if record.project.namespace.parent.user_namespace? && record.project.namespace.parent.owner == user
+    return true if Member.namespace_owners_include_user?(user, record.project.namespace) == true
 
     details[:name] = record.name
     false
--- a/app/views/projects/samples/attachments/_attachment.html.erb
+++ b/app/views/projects/samples/attachments/_attachment.html.erb
@@ -2,7 +2,7 @@
   id="<%= dom_id(attachment) %>"
   class="bg-white border-b dark:bg-slate-800 dark:border-slate-700"
 >
-  <% if allowed_to?(:update?, @project) %>
+  <% if allowed_to?(:update_sample?, @project) %>
     <td class="px-6 py-4">
       <%= check_box_tag "attachment_ids[]",
       (
--- a/app/views/projects/samples/attachments/_attachment.html.erb
+++ b/app/views/projects/samples/attachments/_attachment.html.erb
@@ -99,7 +99,7 @@
   </td>
 
   <td class="px-6 py-4">
-    <% if allowed_to?(:destroy_sample?, @project) %>
+    <% if allowed_to?(:destroy_attachment?, @sample) %>
       <%= button_to(
         t(".delete"),
         namespace_project_sample_attachment_path(
--- a/app/views/projects/samples/attachments/_attachment.html.erb
+++ b/app/views/projects/samples/attachments/_attachment.html.erb
@@ -99,7 +99,7 @@
   </td>
 
   <td class="px-6 py-4">
-    <% if allowed_to?(:update_sample?, @project) %>
+    <% if allowed_to?(:destroy_sample?, @project) %>
       <%= button_to(
         t(".delete"),
         namespace_project_sample_attachment_path(
--- a/app/views/projects/samples/attachments/_attachment.html.erb
+++ b/app/views/projects/samples/attachments/_attachment.html.erb
@@ -2,25 +2,27 @@
   id="<%= dom_id(attachment) %>"
   class="bg-white border-b dark:bg-slate-800 dark:border-slate-700"
 >
-  <td class="px-6 py-4">
-    <%= check_box_tag "attachment_ids[]",
-    (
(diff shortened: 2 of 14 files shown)


app/policies/sample_policy.rb
```diff
@@ -2,6 +2,15 @@
 
 # Policy samples authorization
 class SamplePolicy < ApplicationPolicy
+  def destroy_attachment?
+    project = Project.find(record.project_id)
```
deepsidhu85, line 6: Since we are authorizing the `@sample` you can just do `record.project` in the calls below. We don't need to do a query here

test/system/projects/samples_test.rb
```diff
@@ -704,5 +704,21 @@ class SamplesTest < ApplicationSystemTestCase
       end
       assert_text I18n.t('projects.samples.attachments.deletions.destroy.success')
     end
+
+    test 'user should not see sample attachment delete buttons if they are non-owner' do
```
deepsidhu85, line 708: This test should be broken up into 2 different tests (one for when the user can see the delete buttons and one for when the user can't)

deepsidhu85 (changes requested):
Still need to test it out but looking good. Just a couple of comments below

app/views/projects/samples/show.html.erb
```diff
@@ -51,19 +51,21 @@
         },
```
deepsidhu85, line 51: One more thing, the checkboxes also should not display for non-owner

image

deepsidhu85 (changes requested):
One more comment below about the checkboxes

app/policies/sample_policy.rb
ChrisHuynh333: Removed in 8c4be64cd61a4a7fae8b642efa2a2c6c75d0ea7f

test/system/projects/samples_test.rb
ChrisHuynh333: Fixed in 8c4be64cd61a4a7fae8b642efa2a2c6c75d0ea7f

app/views/projects/samples/show.html.erb
ChrisHuynh333: As discussed on Teams, because maintainers will still need the checkboxes to concatenate, I've hidden the checkboxes for users below maintainer level using the same policy as concatenation `project update?` in 8c4be64cd61a4a7fae8b642efa2a2c6c75d0ea7f

app/views/projects/samples/attachments/_attachment.html.erb
```diff
@@ -2,25 +2,27 @@
...
-    id: dom_id(attachment.file, :file),
-    "aria-label": attachment.file.filename,
-    data: {
-      selection_target: "rowSelection",
-      action: "input->selection#toggle"
-    },
-    class:
-      "w-4 h-4 text-primary-600 bg-slate-100 border-slate-300 rounded focus:ring-primary-500 dark:focus:ring-primary-600 dark:ring-offset-slate-800 focus:ring-2 dark:bg-slate-700 dark:border-slate-600" %>
-  </td>
+  <% if allowed_to?(:update?, @project) %>
```
deepsidhu85, line 5: I think we should use the policy method `update_sample?` instead as the attachments belong to a sample

app/views/projects/samples/attachments/_table.html.erb
```diff
@@ -6,7 +6,9 @@
   >
     <thead class="text-slate-700 bg-slate-50 dark:bg-slate-900 dark:text-slate-400">
       <tr>
-        <th aria-hidden="true" class="px-6 py-3"></th>
+        <% if allowed_to?(:update?, @project) %>
```
deepsidhu85, line 9: Same comment as above to use `update_sample?`

test/system/projects/samples_test.rb
```diff
@@ -704,5 +704,24 @@ class SamplesTest < ApplicationSystemTestCase
...
+      login_as users(:ryan_doe)
+      visit namespace_project_sample_url(namespace_id: @namespace.path, project_id: @project.path, id: @sample1.id)
+      assert_text I18n.t('projects.samples.show.delete_files_button'), count: 0
+      within %(turbo-frame[id="attachments"]) do
+        assert_selector 'table #attachments-table-body tr', count: 2
+        assert_text I18n.t('projects.samples.attachments.attachment.delete'), count: 0
+      end
+    end
   end
```
deepsidhu85, line 726: Can you add some tests to make sure that the checkboxes only display for members with an access level >= Maintainer

app/views/projects/samples/show.html.erb
deepsidhu85: As mentioned in the other comment we should use the `update_sample?` policy method

deepsidhu85 (changes requested):
Looking good just a few more small comments

app/views/projects/samples/attachments/_attachment.html.erb
ChrisHuynh333: Updated in e75c67e5f4998e09d5e36342368a13419a860e91

app/views/projects/samples/attachments/_table.html.erb
ChrisHuynh333: Updated in e75c67e5f4998e09d5e36342368a13419a860e91

test/system/projects/samples_test.rb
ChrisHuynh333: Added in e75c67e5f4998e09d5e36342368a13419a860e91

deepsidhu85 (approved):
Looks good to me!

app/views/projects/samples/attachments/_attachment.html.erb
```diff
@@ -2,25 +2,27 @@
...
-    ),
-    nil,
-    id: dom_id(attachment.file, :file),
-    "aria-label": attachment.file.filename,
-    data: {
-      selection_target: "rowSelection",
-      action: "input->selection#toggle"
-    },
-    class:
-      "w-4 h-4 text-primary-600 bg-slate-100 border-slate-300 rounded focus:ring-primary-500 dark:focus:ring-primary-600 dark:ring-offset-slate-800 focus:ring-2 dark:bg-slate-700 dark:border-slate-600" %>
```
joshsadam, line 22: **question:** Do you think we should hide the `Action` table header, by this I mean the entire column, if they don't have permissions, it looks funny leaving the row there.

deepsidhu85: From what I understand, we are going to move the downloading of a file from the filename into an action

joshsadam: Makes sense.

joshsadam (approved):
This looks good to me.

PR #387, opened 2024-01-23
Edit metadata

ChrisHuynh333:
In this PR, the edit action has been added to the metadata listing ui. Users with access level >= Maintainer level will have access. The edit action will open a modal where users can edit the metadata key and/or value.

The metadata/edit_fields controller will initially receive the edit form param and validate it before calling the Samples::Metadata::UpdateService.

Note: The PR and changes take into account add and delete metadata which will be added afterwards. 
- `services/samples/metadata/fields` folder currently only contains `edit_service` but an `add_service` will be added.
- The table action ellipsis only contains edit (or nothing for analysis metadata), but delete will be added to the dropdown.

New edit action:
image

Edit modal:
image

Successful update:
image

Unchanged metadata error:
image

Key already exists error:
image

1. Set up a sample with metadata, preferably with fields from both a user and analysis source.
2. Navigate to the sample metadata page
3. Edit both keys and values to ensure they update accordingly.
4. Ensure two keys with the same name cannot exist (including different cases/capitilization).
5. Ensure any metadata added by an analysis does not have the edit action available in the action dropdown.
6. Ensure users with access < Maintainer do not see the action column or action dropdown ellipsis.

## PR acceptance checklist


ChrisHuynh333:
I think my process to figure out which field was edited is quite 'expensive' currently. All metadata gets passed to the back end, and I'm looping through the metadata until the edited field is found. I imagine there's probably a solution where we can isolate the edited field on the front end and only submit that, but figured I'd get reviewed and thoughts before pursuing.

app/controllers/projects/samples/metadata/fields_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        # Validates metadata edit params and builds the expected update_service param prior to calling
+        # MetadataController and the metadata update_service
+
+        # Param is received as:
+        # params: {sample: {edit_field: {key: {old_key: new_key}, value: {old_value: new_value}}}
+        # Fields that have not been changed will have equal old and new
+        def update # rubocop:disable Metrics/AbcSize
+          authorize! @project, to: :update_sample?
+          metadata_update_params = ::Samples::Metadata::Fields::EditService.new(@project, @sample, current_user,
```
ericenns, line 18: **nitpick (blocking):** Can you update the `EditService` to actually edit the fields, so after validation etc it should call the `Samples::Metadata::UpdateService`

app/controllers/projects/samples/metadata/fields_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+                                                                                edit_field_params).execute
+          if metadata_update_params.empty?
+            render status: :unprocessable_entity,
+                   locals: { key: edit_field_params['edit_field']['key'].values[0],
+                             value: edit_field_params['edit_field']['value'].values[0] }
+          else
+            request[:sample] = metadata_update_params
+            Projects::Samples::MetadataController.dispatch(:update, request, response)
```
ericenns, line 26: **nitpick (blocking):** As per comment above we don't wan't to call the other controller from this controller method.

app/services/samples/metadata/fields/edit_service.rb
```diff
@@ -0,0 +1,82 @@
...
+        end
+
+        def execute
+          authorize! @project, to: :update_sample?
+
+          validate_sample_in_project
+
+          validate_edit_fields
+
+          construct_metadata_update_params
```
ericenns, line 28: **nitpick (blocking):** after crafting the update params this service should call the `Samples::Metadata::UpdateService`

ericenns (changes requested):
Looking better, but still would like some more refactor as per comments below

app/controllers/projects/samples/metadata/fields_controller.rb
ChrisHuynh333: Updated in 5df4af3ff5328969250bf3de5a14d6bde116f993

app/services/samples/metadata/fields/edit_service.rb
ChrisHuynh333: Updated in 5df4af3ff5328969250bf3de5a14d6bde116f993

app/controllers/projects/samples/metadata/fields_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+            render_params = get_render_status_and_message(updated_metadata_field)
+            render status: render_params[:status], locals: { type: render_params[:message][:type],
+                                                             message: render_params[:message][:message] }
+          end
+        end
+
+        private
+
+        def edit_field_params
+          params.require(:sample).permit(edit_field: {})
```
ericenns, line 31: **nitpick (blocking):** You should be able to update this params method to be specific, i.e. use the keys present in comment above `update` method. So just incorporating `key` and `value` as being a hash should provide more context and then if bad keys were given it wouldn't try to pass them.

ericenns (changes requested):
This is looking a lot better, just one minor comment.

app/controllers/projects/samples/metadata/fields_controller.rb
ChrisHuynh333: Hopefully I didn't misinterpret what you meant - Added in a0e2712c46389d63c119a6251e2dab1bc003c610

app/views/projects/samples/metadata/_edit_form.html.erb
```diff
@@ -0,0 +1,37 @@
+<%= form_with(model: @sample, url: namespace_project_sample_metadata_field_path(sample_id: @sample.id), method: :patch) do |form| %>
+  <%= turbo_frame_tag 'update-alert' %>
+  <div class="grid gap-4">
+  <%= form.fields_for :edit_field do |field| %>
+    <%= field.fields_for :key do |metadata_key| %>
+      <div class="form-field">
+        <%= metadata_key.label t("projects.samples.show.metadata.edit.key"),
+                  class:
+                    "block mb-2 text-sm font-medium text-slate-900 dark:text-white" %>
+        <%= metadata_key.text_field key,
```
joshsadam, line 10: **suggestion:** Can you try to set autofocus on this when the modal opens

app/views/projects/samples/metadata/_edit_form.html.erb
```diff
@@ -0,0 +1,37 @@
...
+      </div>
+    <% end %>
+     <%= field.fields_for :value do |metadata_value| %>
+      <div class="form-field">
+        <%= metadata_value.label t("projects.samples.show.metadata.edit.value"),
+                  class:
+                    "block mb-2 text-sm font-medium text-slate-900 dark:text-white" %>
+        <%= metadata_value.text_field value,
+                        value: value,
+                        required: true,
```
joshsadam, line 24: **question:** @ericenns how do you feel about a user wanting to remove a value from this, i.e. clearing the input?  Or are we just going to have a delete option?

joshsadam (commented):
This looks really good to me.  I just have a couple questions.

app/views/projects/samples/metadata/_edit_form.html.erb
ChrisHuynh333: Fixed in f4e48c28724d8724dd996660540c38875bf81aef

config/locales/en.yml
```diff
@@ -875,6 +884,11 @@ en:
           error: "The following list of samples failed to transfer:"
           no_samples_transferred_error: "Samples were not transferred for the following reasons:"
       metadata:
+        edit:
+          edit: Edit
```
joshsadam, line 888: **suggestion:** Can you change the label of the button to open the modal, and the button to save the updated metadata to `Update`.

ChrisHuynh333: Updated in 37288a651a4b8797332102de5ea22a1996b006d4

app/services/samples/metadata/fields/update_service.rb
```diff
@@ -0,0 +1,83 @@
...
+            end
+
+            @metadata_update_params['metadata'][@key_update.keys[0]] = ''
+          end
+
+          @metadata_update_params['metadata'][@key_update.values[0]] = @value_update.values[0]
+        end
+
+        # Checks if the new key already exists within the @sample.metadata
+        def validate_new_key
```
deepsidhu85, line 70: I don't think this method and loop is required. You should be able to just call `@sample.metadata.key?(@key_update.values[0].downcase)` above in the `construct_metadata_update_params` method

test/controllers/projects/samples/metadata/fields/fields_controller_test.rb
```diff
@@ -0,0 +1,86 @@
...
+          test 'cannot update sample if not a member with access to the project' do
+            sign_in users(:david_doe)
+            patch namespace_project_sample_metadata_field_path(@namespace, @project29, @sample32),
+                  params: { 'sample' => { 'update_field' => {
+                    'key' => { 'metadatafield1' => 'metadatafield3' },
+                    'value' => { 'value1' => 'value3' }
+                  } }, format: :turbo_stream }
+            assert_response :unauthorized
+          end
+        end
```
deepsidhu85, line 82: Please add a test for a user not being able to update analysis provided metadata

test/services/samples/metadata/fields/update_service_test.rb
```diff
@@ -0,0 +1,157 @@
...
+          params = { 'update_field' => { 'key' => { 'metadatafield1' => 'metadatafield2' },
+                                         'value' => { 'value1' => 'value1' } } }
+
+          Samples::Metadata::Fields::UpdateService.new(@project29, @sample32, @user, params).execute
+
+          assert @sample32.errors.full_messages.include?(
+            I18n.t('services.samples.metadata.update_fields.key_exists', key: 'metadatafield2')
+          )
+        end
+      end
```
deepsidhu85, line 154: Please add a test for a user not being able to update analysis provided metadata

deepsidhu85 (changes requested):
This looks great just have a couple of comments below

app/services/samples/metadata/fields/update_service.rb
ChrisHuynh333: Updated in 2a79e5b24b98f16b80cf676792daa31801249850

test/controllers/projects/samples/metadata/fields/fields_controller_test.rb
ChrisHuynh333: Updated in 2a79e5b24b98f16b80cf676792daa31801249850

test/services/samples/metadata/fields/update_service_test.rb
ChrisHuynh333: Updated in 2a79e5b24b98f16b80cf676792daa31801249850

deepsidhu85 (approved):
Looks good to me!

PR #405, opened 2024-02-02
Delete metadata

ChrisHuynh333:
This PR adds the ability to delete metadata through the website UI. Users with access can delete metadata either by the action dropdown, or by using the checkboxes and clicking the `Delete Metadata` button.

Delete by action dropdown:
[Screencast from 2024-02-14 08:09:23 AM.webm](https://github.com/phac-nml/irida-next/assets/82407232/8246cb8a-9e1f-45cd-b674-bf318c00bbeb)

Delete multiple fields with checkboxes and modal:
[Screencast from 2024-02-14 08:09:51 AM.webm](https://github.com/phac-nml/irida-next/assets/82407232/0904e84d-b207-44fd-b116-522fe88d6882)

Metadata page for user < Maintainer access:
image

1. Navigate to a sample's metadata page
2. Ensure you are able to delete metadata by both the action dropdown and with the checkboxes and `Delete Metadata` button
3. Logic for attachment checkboxes was slightly altered due to the implementation of the metadata checkboxes, so ensure the following:
a) Ensure checking off files and metadata does not enable the other's action buttons (ie: checking off metadata does not enable concatenate/delete files buttons, checking off files does not enable the delete metadata button)
b) Ensure file concatenation behaves as expected
c) Check both files and metadata for deletion. Ensure deleting one set does not affect the other set (ie: deleting files does not affect any metadata)
4. Ensure a user with access < Maintainer does not see metadata checkboxes, action dropdown, or the `Delete Metadata` button.

## PR acceptance checklist


app/javascript/controllers/projects/samples/metadata/delete_listing_controller.js
```diff
@@ -0,0 +1,26 @@
...
+    static targets = ["table"];
+
+    connect() {
+        const body = document.getElementById("metadata-table-body");
+
+        for (let row of body.rows) {
+            const isChecked = row.cells[0].children[0].checked;
+
+            if (isChecked) {
+                const newRow = this.tableTarget.insertRow(-1);
```
joshsadam, line 14: You are inserting theis row into the `thead`.  Add a `tbody` to your table and update the target to that.

app/views/projects/samples/metadata/deletions/_modal.html.erb
```diff
@@ -0,0 +1,37 @@
...
+    data: {
+      controller: "projects--samples--metadata--destroy",
+      'data-projects--samples--metadata--destroy-storage-key-value':namespace_project_sample_url(id: @sample.id),
+      action: 'turbo:submit-end->projects--samples--metadata--destroy#clear'
+    }, method: :delete) do |form| %>
+
+      <div data-projects--samples--metadata--destroy-target="field">
+      </div>
+
+      <div class="grid gap-4">
```
joshsadam, line 29: **nitpick:** You don't need this wrapper `div.grid.gap-4` or even the `flex` on the inner div.

joshsadam (changes requested):
Just a few quick comments on some things I saw while looking through this. It functions great though, thanks.

ChrisHuynh333:
Because the files and metadata now share a url, and therefore would share checkbox logic, I've had to specify the query path in order for the checkboxes to not overlap with one another (ie: attachment checkboxes not enabling delete metadata and vice versa).

This has led to a cumbersome implementation where I've had to add `tab: 'files'` to all redirects and navigation to 'namespace_project_sample_url` (edit sample, clicking sample from groups samples page, etc.) so that the query `?=files` is in the URL. Just wondering if this is fine or there's a better solution for this. Maybe a before_action checking if the `param[:tab].nil?` and then add the files query to the URL if nil? params[:tab] is not nil when clicking between files/metadata tab on the sample show page, but would be nil from anywhere else.

Let me know if this doesn't make sense!

app/javascript/controllers/projects/samples/metadata/delete_listing_controller.js
ChrisHuynh333: Great catch, fixed in b9073c70ebecdaae80f23d5cbe4a5a050279ee7e

app/views/projects/samples/metadata/deletions/_modal.html.erb
ChrisHuynh333: Fixed in b9073c70ebecdaae80f23d5cbe4a5a050279ee7e

ChrisHuynh333:
> Because the files and metadata now share a url, and therefore would share checkbox logic, I've had to specify the query path in order for the checkboxes to not overlap with one another (ie: attachment checkboxes not enabling delete metadata and vice versa).
> 
> This has led to a cumbersome implementation where I've had to add `tab: 'files'` to all redirects and navigation to `namespace_project_sample_url` (edit sample, clicking sample from groups samples page, etc.) so that the query `?tab=files` is in the URL. Just wondering if this is fine or there's a better solution for this. Maybe a before_action checking if the `param[:tab].nil?` and then add the files query to the URL if nil? params[:tab] is not nil when clicking between files/metadata tab on the sample show page, but would be nil from anywhere else.
> 
> Let me know if this doesn't make sense!

I came up with a workaround:
- All routing is back to normal - initial click onto a sample#show will not include any queries in the URL, but clicking between files/metadata and soon to be history tabs will have `?tab=<TAB NAME>`
- Added an empty div in attachments/_table which will contain the sample ID
- in stimulus controller `selection_controller`, I check for this empty div, if it is present, we used a hard coded path for storageKey that contains ?tab=files. This will allow both the initial routing and routing which includes ?tab=files to have the same 'set' of checkboxes stored.
- In stimulus controller `attachments/selected_attachments`, I've had to set the default storageKey to have a hard coded `?tab=files` as well.

I think this solution is the most flexible... Let me know if you think of an alternative idea.

config/locales/en.yml
```diff
@@ -844,6 +845,8 @@ en:
           actions:
             dropdown:
               update: Update
+              delete: Delete
+            delete_confirm: Are you sure you want to remove this metadata field from the sample?
```
joshsadam, line 849: **question:** Do you think we should at least say the `key` that is going to be deleted?  I can see a user clicking the remove button, seeing the confirmation, and then getting nervous that they might of clicked the wrong one.

app/javascript/controllers/selection_controller.js
```diff
@@ -9,10 +9,20 @@ export default class extends Controller {
   static outlets = ["action-link"];
 
   connect() {
+    // When you are first routed to a sample#show page, the path will not contain any query information about the
```
joshsadam, line 12: **suggestion:** I think you might have other thought this a little too much.  What I think would be better is to revert this file back to the original then just set the `data-storage-key` attribute (which actually should be re-written as a stimulus `value` I think).  That will pass in what ever key you want to use for storage without having to hack this.  This selection should be good for any table that has the capability to  select items.

joshsadam (changes requested):
A couple small things for you to update

config/locales/en.yml
ChrisHuynh333: Agreed, fixed in bb0755d9fe743251a1552eb46bb0e92b16c8441d

app/javascript/controllers/selection_controller.js
ChrisHuynh333: Thanks for this, much simpler. Changed in 47a3ba31d176c4f5e799487e1606f6263374f109

joshsadam (approved):
This looks great to me, thanks for making all those changes

app/controllers/projects/samples/metadata/deletions_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+module Projects
+  module Samples
+    module Metadata
+      # Controller actions for Project Samples Metadata Deletions Controller
+      class DeletionsController < Projects::Samples::ApplicationController
+        respond_to :turbo_stream
+
+        def new
+          authorize! @sample, to: :destroy_attachment?
```
deepsidhu85, line 11: We probably want to authorize the project and to `update_sample?` since maintainer access level and higher should be allowed to delete metadata. `destroy_attachment?` doesn't make sense here since we are modifying metadata not the attachment

app/controllers/projects/samples/metadata_controller.rb
```diff
@@ -22,6 +22,30 @@ def edit
                                                    value: params[:value]
                                                  }), status: :ok
       end
+
+      def destroy
+        authorize! @project, to: :update_sample?
```
deepsidhu85, line 27: You can remove this authorization here since the `Samples::Metadata::UpdateService` handles that in the `execute`

deepsidhu85 (changes requested):
This looks great just have a few comments below before I give it a final manual test

app/controllers/projects/samples/metadata/deletions_controller.rb
ChrisHuynh333: Fixed in 969c18a281013c8d4311e2b0ddb355e3066185f6

app/controllers/projects/samples/metadata_controller.rb
ChrisHuynh333: Removed in 969c18a281013c8d4311e2b0ddb355e3066185f6

deepsidhu85 (approved):
Looks good to me!

PR #419, opened 2024-02-15
Workflow: Pipeline configuration

deepsidhu85:

This PR adds in a pipeline configuration which will register the available pipelines that can be launched within IRIDA Next. The pipeline configuration is stored in `config/pipelines/pipeline.json`. For each pipeline and version a `nextflow_schema` and `schema_input` file are retrieved from the github repo for the pipeline and are stored under `private/pipelines/REPO_ORGANIZATION_AND_NAME/VERSION/`. Along with these files a `status.json` file is generated and stored which has the `etag` header values for both files, which is used to determine if a newer schema file needs to be downloaded when the server is started up.

image

**Version 1.0.2**

image

**Version 1.0.0**

image

1. Verify that you don't have a `private/pipelines/` directory
2. Start the IRIDA Next server
3. You should now see that there is a `private/pipelines/phac-nml/iridanextexample` directory with sub directories for the 3 versions in the pipelines.json file (1.0.0, 1.0.1, 1.0.2)
4. Verify in each version directory there are 3 files: `nextflow_schema.json`, `schema_input.json`, and `status.json`
5. Shutdown the server
6. To test that an existing file with the same etag as the file on the server is not downloaded again, open `lib/irida/pipelines.rb` and after the line `return if resource_etag_exists(schema_file_url, pipeline_schema_files_path, type)` in the `write_schema_file` method, put a print statement such as `puts "WRITING FILE" and save the file
7. Start the IRIDA Next server
8. Watch the console to see that there are no "WRITING FILE" lines
9. Login as any user
10. Create a project and sample
11. Select the sample from the project `Samples` page and click the rocket icon
12. Verify 3 pipelines are displayed (Versions 1.0.2, 1.0.1, 1.0.0)
13. Click through the different pipelines (Version 1.0.2 and 1.0.0) should have different parameter forms

**Note: In the dev environment the Register Pipelines is called once for each process (web, css, and job)**

## PR acceptance checklist

### DIFF
--- a/app/controllers/workflow_executions/submissions_controller.rb
+++ b/app/controllers/workflow_executions/submissions_controller.rb
@@ -30,11 +30,7 @@ module WorkflowExecutions
       workflow_name = params[:workflow_name]
       workflow_version = params[:workflow_version]
 
-      workflow_index = @workflows.index do |workflow|
-        (workflow.name == workflow_name) && (workflow.version == workflow_version)
-      end
-
-      @workflow = @workflows[workflow_index]
+      @workflow = @workflows[find_pipeline_by(workflow_name, workflow_version)]
     end
 
     def samples
--- a/app/controllers/workflow_executions/submissions_controller.rb
+++ b/app/controllers/workflow_executions/submissions_controller.rb
@@ -9,6 +9,7 @@ module WorkflowExecutions
     respond_to :turbo_stream
     before_action :workflows, only: %i[pipeline_selection new]
     before_action :samples, only: %i[new]
+    before_action :workflow, only: %i[new]
     before_action :workflow_schema, only: %i[new]
 
     def pipeline_selection
@@ -16,7 +17,6 @@ module WorkflowExecutions
     end
 
     def new
-      @workflow = @workflows[1]
       render status: :ok
     end
 
@@ -26,16 +26,19 @@ module WorkflowExecutions
       @workflows = available_workflows
     end
 
+    def workflow
+      workflow_index = @workflows.index { |workflow| workflow.id == params['workflow_id'].to_i }
+      @workflow = @workflows[workflow_index]
+    end
+
     def samples
       sample_ids = params[:sample_ids]
       @samples = Sample.where(id: sample_ids)
     end
 
     def workflow_schema
-      schema_loc = @workflows[0].schema_loc
-
       # Need to get a schema file path from the workflow
-      @workflow_schema = JSON.parse(Rails.root.join(schema_loc).read)
+      @workflow_schema = JSON.parse(Rails.root.join(workflow.schema_loc).read)
     end
   end
 end
--- a/app/controllers/workflow_executions/submissions_controller.rb
+++ b/app/controllers/workflow_executions/submissions_controller.rb
@@ -4,10 +4,10 @@ module WorkflowExecutions
   # Workflow submission controller
   class SubmissionsController < ApplicationController
     respond_to :turbo_stream
-    before_action :workflows, only: %i[pipeline_selection]
+    # before_action :workflows, only: %i[pipeline_selection]
     before_action :samples, only: %i[new]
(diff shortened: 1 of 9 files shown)


app/controllers/workflow_executions/submissions_controller.rb
```diff
@@ -3,11 +3,14 @@
 module WorkflowExecutions
   # Workflow submission controller
   class SubmissionsController < ApplicationController
+    include Irida::Pipelines
+    cattr_reader :available_pipelines
```
ericenns, line 7: **nitpick (blocking):** I think you accidentally added this line e.g. `cattr_reader :available_pipelines` as there is no available_pipelines here and instead it is defined in `Irida::Pipelines`

lib/irida/pipeline.rb
```diff
@@ -0,0 +1,25 @@
...
+  class Pipeline
+    attr_accessor :name, :description, :metadata, :type, :type_version,
+                  :engine, :engine_version, :url, :execute_loc,
+                  :version, :schema_loc, :schema_input_loc
+
+    def initialize(entry, version, schema_loc, schema_input_loc)
+      @name = entry['name']
+      @description = entry['description']
+      @metadata = { workflow_name: name, workflow_version: version }
+      @type = nil
```
ericenns, line 14: **nitpick (blocking):** Can you set these `nil` fields to the values that were hardcoded previously in the submissions controller as those are the defaults and should be set here. Otherwise submission will fail on WES.

lib/irida/pipelines.rb
```diff
@@ -0,0 +1,129 @@
...
+require 'net/http'
+require 'irida/pipeline'
+
+module Irida
+  # Module that reads a workflow config file and registers the available pipelines
+  module Pipelines
+    mattr_accessor :register_pipelines, :find_pipeline_by
+    cattr_accessor :available_pipelines, :pipeline_config_dir, :pipeline_schema_file_dir
+
+    @@pipeline_config_dir = 'config/pipelines' # rubocop:disable Style/ClassVars
```
ericenns, line 15: **nitpick (blocking):** can you change these `@@variable` to normal UPPER_CASE_VARIABLE except for `@@available_pipelines = []` which should be changed to `@available_pipelines = []`

lib/irida/pipelines.rb
```diff
@@ -0,0 +1,129 @@
...
+require 'json'
+require 'open-uri'
+require 'uri'
+require 'net/http'
+require 'irida/pipeline'
+
+module Irida
+  # Module that reads a workflow config file and registers the available pipelines
+  module Pipelines
+    mattr_accessor :register_pipelines, :find_pipeline_by
```
ericenns, line 12: **praise:** the methods should be accesible without using `mattr_accessor` it just means you might have to call them with `Irida::Pipelines.available_pipelines` but I think that is nicer than using the `include` that you did in SubmissionsController.

ericenns (changes requested):
Just a few styling and code formatting comments

app/controllers/workflow_executions/submissions_controller.rb
deepsidhu85: Yup definitely was added accidentally. Updated to call Irida::Pipelines.* directly and removed the include and cattr_accessor in d08fd2d05586261bfb12aafc47e7e971a61f711f

lib/irida/pipelines.rb
deepsidhu85: Updated variables to get rid of the rubocop errors caused by the use of @@ ind08fd2d05586261bfb12aafc47e7e971a61f711f

lib/irida/pipelines.rb
deepsidhu85: Definitely alot nicer. Cleaned up in d08fd2d05586261bfb12aafc47e7e971a61f711f

ericenns (approved):
Looks good to me!

PR #425, opened 2024-02-19
Clone Samples UI

### LINKED ISSUE #404: Samples - Clone UI
# Summary
As a user I want to be able to clone data or a subset of data that has been shared with me, into my own project so that I can perform analyses and edit the data without affecting the original source data.

# Acceptance Criteria
- User with correct permissions (`Maintainer`) should be able to select samples from a project that has been shared with them
- `Clone` button that is enabled when >= 1 samples are selected
- Modal which has a dropdown for the user to select the project which they want to clone the data to
- Should make use of the `CloneController` from #403

ksierks:
The front-end for cloning samples.
Fixes #404.

The clone dialog:
Screenshot from 2024-02-19 15-37-50
A success message:
notification
A validation error:
Screenshot from 2024-02-27 10-50-31
Sample errors:
Screenshot from 2024-02-29 10-33-50

1. Navigate to a project that the logged-in user is at least a maintainer of.
1. Hit `Samples` on the left side menu.
1. Check the samples you wish to clone.
1. Click the `Clone samples` button.
1. A dialog should appear. Select the project you wish to clone the samples to.
1. Verify the samples were cloned successfully.

Note: The samples that could not be cloned should remain checked within the samples table.

## PR acceptance checklist


app/javascript/application.js
```diff
@@ -1,8 +1,9 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "@hotwired/turbo-rails";
 import "controllers";
-import * as ActiveStorage from "@rails/activestorage"
-ActiveStorage.start()
+import "./turbo_streams";
```
joshsadam, line 4: **thought:** Can you update this functionality to be the same as you did for the transfer samples?  That did not require any custom turbo actions.

app/javascript/controllers/projects/samples/transfers_controller.js
```diff
@@ -16,13 +16,15 @@ export default class extends Controller {
       sessionStorage.getItem(this.storageKeyValue)
     );
 
-    for (const storageValue of storageValues) {
```
joshsadam, line 19: **nitpick:** This controller should probably be renamed as it is being used by multiple functionality now

app/javascript/turbo_streams/index.js
```diff
@@ -0,0 +1,13 @@
+function updateSessionStorageItem() {
```
joshsadam, line 1: **nitpick:** You won't need this anymore

joshsadam (changes requested):
Hi Katherine,  Looking through your code I think the custom turbo action is an overkill (interesting feature though I did not know about it).  Can you refactor this code to work similar to how you did for the samples transfer functionality?  Also can you change from using alert element to using the notification system, we are trying to remove all alerts.

app/views/projects/samples/clones/create.turbo_stream.erb
```diff
@@ -0,0 +1,26 @@
+<% @cloned_sample_ids.keys.each do |cloned_sample_id| %>
+  <%= turbo_stream.update_session_storage_item(
+    namespace_project_samples_url,
+    cloned_sample_id
+  ) %>
+<% end %>
+<% if @project.errors.messages_for(:base).empty? %>
```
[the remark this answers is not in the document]
joshsadam: If there are a lot of errors, you could return a new open dialog with them listed within.

deepsidhu85:
Just leaving a general comment pertaining to the modal for the cloning samples. Currently it does not display which samples were selected to be cloned (I know the same is true for the transfer modal). We probably should have the samples selected (atleast the sample names, maybe PUIDs?) displayed in the modal. Since this is also the case for the transfer modal I think it would be better to do this in a follow up pr where both modals can be addressed

app/views/projects/samples/clones/create.turbo_stream.erb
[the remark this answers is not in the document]
deepsidhu85: Agree with this, if there are a lot we should have them inside a modal with a scrollable section so that the modal doesn't become humongous either

app/javascript/application.js
ksierks: I don't know if I can effectively and thoroughly explain the benefits of the custom turbo_stream action in this PR. My preference is to do a Teams call to do a quick demo and talk it over. I'll try...

Here is an example to test the new custom turbo_stream action functionality:
1. Login to irida-next.
1. Go to a project that the user is at least a maintainer of.
1. Click the samples tab.
1. Select some samples you wish to clone to another project.
1. Click the `clone samples` button.
1. Select a project from the drop down.
1. Click the `submit` button.
1. Select the same samples as before and some other samples.
1. Click the `clone samples` button.
1. Select the same project as before from the drop down.
1. Verify the successful samples are unchecked and the unsuccessful samples are checked.

Transfer samples does not uncheck successful transferred samples during a partial response. Currently, it clears the entire key from storage. This functionality is missing from transfer samples. I tried implementing it in the transfer samples PR, but I went down the wrong path using the form turbo:submit-end event. I want to be able to update transfer samples to include the custom turbo_stream action too.

If a custom turbo_stream is overkill, maybe we could use the storage actions from turbo_power, https://github.com/marcoroth/turbo_power? I didn't want to install a new gem and they don't have an action with this exact functionality.

Lastly, we send the cloned sample ids back in the response, but we don't use the returned sample ids. Now we can use the sample ids to persist the state after cloning/transferring is completed.

app/javascript/turbo_streams/index.js
ksierks: Please see above comment.

app/views/projects/samples/clones/create.turbo_stream.erb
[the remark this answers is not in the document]
ksierks: I'm a little confused as to when we want clone samples to be the same as transfer samples and when we want clones samples to be different than transfer samples. Should I be replacing the alerts with notifications for **both** transfer and clone samples **within this PR**?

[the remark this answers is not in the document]
deepsidhu85: We can just update the one for clone samples in this PR. The transfer one can be updated in a separate pr since it's for different functionality

app/javascript/application.js
deepsidhu85: I definitely understand the benefits of the custom turbo stream action, but I guess my question is, why are we keeping samples selected that failed to clone? The user receives an error message with the samples that failed to clone so I don't see the benefit of keeping them selected. If the idea is that the user can then try to clone those samples to a different project, well in that case the user should have to make a conscious choice about selecting which samples they wanted to clone. 

I feel like we had the same discussion about the transfers as well where originally they weren't being cleared but we updated it to just delete the storage key from session storage

ksierks: Recap from Thursday meeting:
- A user could select 100+ samples and try to clone them to a project but accidentally select the wrong project.
It might be tedious for the user to reselect the same samples if the selected project was, for example, one character off.
 ex.) User selects project "/yersinia/yersinia-pseudotuberculosis/outbreak-2023", but meant to select "/yersinia/yersinia-pseudotuberculosis/outbreak-2022".
- Going to wait for @ericenns to be back to have a discussion.

app/javascript/controllers/projects/samples/transfers_controller.js
ksierks: a35631dccb7ac197f67e8b2843fc9dd990bc4bab

app/views/projects/samples/clones/create.turbo_stream.erb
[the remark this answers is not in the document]
ksierks: I'm currently working on this requested change, but I just wanted to remind there is an `overflow-y-auto` set on the alert to allow scrolling.

Screenshot from 2024-02-22 12-07-07

app/javascript/application.js
ericenns: Without using a custom turbo stream action, one could just update their turbo stream response to append an HTML element to the page with data attributes for samples that need to selected or need to be unselected. Then the selection controller could process the data attributes to select or unselect samples.

This achieves the same goal as the custom turbo stream action, but has the benefit of keeping all the selection specific javascript code within a single hotwire controller. This would lead to easier maintenance and understanding.

I'm not sure where I stand yet on the idea of maintaining a partial selection on errors yet, I'll have to think about this.

ksierks: Removed in 6394b9c3ce8973dc00802ab5468aa437769834ce.

app/views/projects/samples/clones/create.turbo_stream.erb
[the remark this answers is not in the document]
ksierks: There are 2 kinds of errors, validation and sample. Validation checks for empty project and sample ids and if the source and destination project ids match. With only enabling the clone button after a sample has been selected and removing the current project from the manageable project drop down, if everything if working as expected, a user should not see a validation error.
I left the validation error as an alert, as there will only ever return one. I moved the success message inside a flash message. (I think that's what you meant by notification.)
There could be many sample errors, so I put them all inside a scrollable dialog, like requested. 
This is what I have so far, but could you please have a look to let me know if I'm on the right track? 1fa9394ae828a4c655fe81da6d552db01cba6d37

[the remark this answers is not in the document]
joshsadam: We are trying to remove all alerts being put on the page.  The primary reason behind this is it causes a shift in the page content.  I like how you have the other errors displaying in the dialog, is there a reason we cannot do that for the validation error?

app/views/projects/samples/clones/_errors.html.erb
```diff
@@ -0,0 +1,16 @@
+<div
+  data-controller="projects--samples--metadata--complete"
+  data-projects--samples--metadata--complete-filters-outlet=".filters"
+>
+  <p class="mb-4"><%= t(".description") %></p>
+
+  <% errors.each do |error| %>
+    <%= viral_alert(type:, message: error, classes: "mb-4") %>
```
joshsadam, line 8: **suggestion:** I think this should just be one alert with a title like `The following errors occurred` (or something better :)), and then a list of each that failed.  It would take up a lot less vertical space.

app/views/projects/samples/clones/create.turbo_stream.erb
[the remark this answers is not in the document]
ksierks: Oh. I thought the problem was about errors taking up too much space, not they shouldn't be on the page at all. After asking errors be put inside the dialog, I copied what we did for for the file metadata import. Screenshots with errors are updated. Please see 8a17ed77f999c15976b7de67eaeab8a6717c8957.

app/views/projects/samples/clones/_errors.html.erb
ksierks: I just want the wording to be clear that no cloning occurred for validation errors and partial cloning occurs for sample errors. Screenshots with errors are updated. Please see 8a17ed77f999c15976b7de67eaeab8a6717c8957.

ericenns (changes requested):
This is working pretty well, my only comments so far are that the error messages are not clear. See image below:

image

It appears to me that the `Sample.id` is being shown in the error message. I notice in transfer samples it uses the `Sample.name`. I think for cloning it would be better to show a message like below:

`Sample 'sample_name (sample_puid)` could not be cloned as Sample 'sample_name' exists in the target project.`

That way we are showing which sample could not be cloned and the reason why, e.g. a Sample already exists with the same name in the target project.

joshsadam (approved):
This looks good to me now, thanks for all the changes.

ksierks:
> This is working pretty well, my only comments so far are that the error messages are not clear. See image below:
> 
> image
> 
> It appears to me that the `Sample.id` is being shown in the error message. I notice in transfer samples it uses the `Sample.name`. I think for cloning it would be better to show a message like below:
> 
> `Sample 'sample_name (sample_puid)' could not be cloned as Sample 'sample_name' exists in the target project.`
> 
> That way we are showing which sample could not be cloned and the reason why, e.g. a Sample already exists with the same name in the target project.

d0cd28201579fae7d55f0ba377657bc7f34dacb9

ericenns (approved):
Looks good to me!

joshsadam (approved):

PR #439, opened 2024-02-22
Workflow: Completion Service 

### LINKED ISSUE #253: Workflow: Completion Service
## Summary
When a workflow has finished running, the Workflow StatusService marks the state as `completed`

Add a CompletionService which 
1. checks that WorkflowExecution is completed
2. handles adding the workflows output files to the Workflow Execution object.
3. sets WorkflowExecution to `finalized`

For this PR, only the main workflow files will be added. Adding per sample results/metadata will be handled in #255 

WorkfowExecution model should have `:outputs` attachments

The Ga4ghWesApi ApiConnection should hold the service path

The output filename needs to be joined with the service path.

An input key is generated from joining the output filenames(with service path) and output keys with `generate_input_key`

An input key + filename is combined and used to copy the file into a new blob, which will then be what is attached to the WorkflowExecution

## References
ActiveSupport::Gzip.decompress(ActiveStorage::Blob.service.download(output_json))

https://github.com/phac-nml/pipeline-standards?tab=readme-ov-file#422-complete-example
"output/irida.output.json.gz"

https://www.rubydoc.info/github/rails/rails/ActiveStorage%2FBlob.compose
This function accepts the input key + filename, it copies the file and makes a new file/blob,
    which we can then attach to the model
    - filename: should be the filename that is shown in the output.json.gz

JeffreyThiessen:
Fixes #253

* Create a new service WorkflowExecutions:CompletionService
  * expects a `completed` WorkflowExecution
  * downloads the `iridanext.output.json.gz` file from the blob service. This file is made by a workflow
  * uses the above file to get a list of output files to attach to the WorkflowExecution
  * downloads the output files into tempfiles, then uploads them into new independent blobs
  * attaches the new blobs to the Workflow Execution
  * sets WorkflowExecution to the new state `finalized`
* Adds `blob_run_directory` column to the workflow_executions table.
  * this column is responsible for holding the blob service key to which the workflow writes files to
* Removes the foreign key constraint on the attachments table.
  * This was unneeded as is breaks non sample attachables, and the polymorphic tag handles this for us

#### Note: When testing with ga4gh connection + local DB for blob storage, the ga4gh_wes connection will not write files back to your db, so you must manually place the files where they are expected to be. If you are using Azure for blob storage with ga4gh_wes hosted on Azure, this will be handled for you like in prod.

1. Run a workflow execution via GUI
- you should upload a paired end to a new sample and use that sample as your workflow input
image

2. wait for state to become `completed`
image

3. Get your WorkflowExecution in rails console
```bash
rails c
```
```ruby
# Workflow Execution
wfe = WorkflowExecution.last
```

4. Local DB step: get your output directory
```ruby
# This is where the workflow execution data should be
wfe.workflow_params['--outdir']

# example:
# "/home/jthiessen/git/irida-next-core/storage/6e/zc/6ezc87o51ym3y4xtljonlucelzuz/output/"
```
You will need to create the `/output/` sub directory to place files in

Use the files in `test/fixtures/files/blob_outputs/normal` or make your own
first `cd` command should match above without the `output/`
```bash
cd ~/git/irida-next-core/storage/aa/bb/genkey/
mkdir output
cd output
cp ~/git/irida-next-core/test/fixtures/files/blob_outputs/normal/iridanext.output.json.gz .
cp ~/git/irida-next-core/test/fixtures/files/blob_outputs/normal/summary.txt .
```

5. Run the Completion Service
```ruby
WorkflowExecutions::CompletionService.new(wfe).execute
```

6. Check that the summary file is attached to the WorkflowExecution
```ruby
wfe.outputs

=> [#<Attachment:0x00007f3e166653e0
  metadata: {"format"=>"unknown", "compression"=>"none"},
  deleted_at: nil,
  attachable_type: "WorkflowExecution",
  created_at: Wed, 20 Mar 2024 16:22:28.327634000 UTC +00:00,
  updated_at: Wed, 20 Mar 2024 16:22:28.343108000 UTC +00:00,
  id: "bde67db3-1e4c-4198-b7a8-6a29d28ffbdf",
  attachable_id: "5c6496f1-4897-4d42-9f4f-5ece446ac441",
  puid: "INXT_ATT_AYDUBZSEKO",
  log_data: nil>]

wfe.outputs[0].filename

=> #<ActiveStorage::Filename:0x00007f3e16a7d270 @filename="summary.txt">
```

## PR acceptance checklist


app/models/workflow_execution.rb
```diff
@@ -11,6 +11,7 @@ class WorkflowExecution < ApplicationRecord
 
   has_many :samples_workflow_executions, dependent: :destroy
   has_many :samples, through: :samples_workflow_executions
+  has_many :outputs, dependent: :nullify, class_name: 'Attachment', as: :attachable
```
ericenns, line 14: **nitpick (blocking):** I think we can remove the `dependent: :nullify` since if the `WorkflowExecution` is destroyed it should destroy the `Attachment` instances.

app/services/workflow_executions/completion_service.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+module WorkflowExecutions
+  # Service used to complete a WorkflowExecution
+  class CompletionService < BaseService
+    def initialize(workflow_execution, user = nil, params = {})
```
ericenns, line 6: **nitpick (blocking):** Instead of passing `user` which I don't think will be passed from the `WorkflowExecutionsStatusJob` which is likely where this service will be called from we can just use the `user` from the `WorfklowExecution` which can be accessed via `@workflow_execution.submitter` and passed to `Attachments::CreateService`.

ericenns (changes requested):
This is looking good, just a few comments

app/models/workflow_execution.rb
JeffreyThiessen: fixed in https://github.com/phac-nml/irida-next/pull/439/commits/4aecb110d87bcdc9df22480b43862175567669ad

app/services/workflow_executions/completion_service.rb
JeffreyThiessen: fixed in https://github.com/phac-nml/irida-next/pull/439/commits/4aecb110d87bcdc9df22480b43862175567669ad

ericenns (approved):
I had to update the workflow execution to be in state `completed` and then I was able to run the `completion_service` and it is working as expected.

deepsidhu85 (approved):
This looks good to me! Working as expected (just had to change the state to completed using rails console as we weren't testing with a wes connection)

PR #444, opened 2024-02-26
Data Export - Create Service

ChrisHuynh333:
This PR creates the `DataExports` `CreateService`. It will accept and validate parameters required to start a `DataExport` model instance before calling the `CreateExportJob` (future PR) to do the actual export creation.

We will use existing seeded data to test so `bin/rails db:drop/create/migrate/seed` if needed.

1. Set `user = User.find_by(email: 'user7@email.com')`
2. Set `params_1 = {'export_type' => 'sample', 'export_parameters' => {'ids' => [Sample.find_by(name: 'Yersinia pestis/Outbreak 2022 Sample 10').id]}, 'email_notification' => true, 'name' => 'test export'}`
3. Create a new `DataExport` with `DataExports::CreateService.new(user, params_1).execute`
4. Check the `export` exists under `user.data_exports.last`. 
5. Ensure  status was automatically set to `processing`

6. Repeat the above except with `user = User.find_by(email: 'user8@email.com')` (we want user8). An export should not generate with a user of access level guest for the sample (expect access level >= Analysis).

## PR acceptance checklist


ksierks (commented):
Hi Chris. I was playing around with the export create service. While experimenting with different parameters, I noticed the following: 
- when not including a name, true is returned, but no export was created
- when including a sample id that does not exist, an error occurs
- changing the export type to something other than sample or analyst, true is returned but no export was created

I'm not sure if any we want to handle any of these cases. They're just my findings.

ChrisHuynh333:
> Hi Chris. I was playing around with the export create service. While experimenting with different parameters, I noticed the following:
> 
> * when not including a name, true is returned, but no export was created
> * when including a sample id that does not exist, an error occurs
> * changing the export type to something other than sample or analyst, true is returned but no export was created
> 
> I'm not sure if any we want to handle any of these cases. They're just my findings.

I will look into these. Thanks!

app/services/data_exports/create_service.rb
```diff
@@ -0,0 +1,58 @@
...
+    # to export the chosen samples' data
+    def validate_sample_export
+      project_ids = []
+
+      params['export_parameters']['ids'].each do |sample_id|
+        sample = Sample.find_by(id: sample_id)
+        raise DataExportCreateError, I18n.t('services.data_exports.create.invalid_sample_id') if sample.nil?
+
+        project_id = sample.project_id
+        project_ids << project_id unless project_ids.include?(project_id)
```
ksierks, line 50: This is more of a question and asking confirmation about the permissions...
I think you might be the first to add checkboxes to the group samples table. 
For the project samples table, you only need to lookup the project once as all the samples belong to the project. We check the direct membership of the project. 
I would have thought that we would want to do the same for group samples. Check the direct membership of the group for the group samples table and add that functionaly to group_policy. Instead, I see you're iterating through all the sample projects. 
If we were to add the checkboxes to the group samples table, we would have to authorize each row individually. 
I am also working on a PR that requires group sample selection. I'm selfishly wondering if @deepsidhu85 or @ericenns can confirm this is how they want to see group samples permissions? If so, I need to update my PR like this one. Sorry and thanks!

ericenns: The `DataExports::CreateService` doesn't have a concept of `Group` or `Project` for selection, so it needs to verify for each sample that the user has the correct permissions. This is independent of displaying checkboxes on the respective samples tables, as this service could be used by the API for example so we have to revalidate permissions. Otherwise a bad user could craft an HTTP request and bypass permissions if we didn't have checking also at the service level.

For the Group Samples table, we don't have to check permissions for each `Sample` we can just check the users permission at the `Group` level.

ksierks: Oh ok. I think that makes sense. I will try to do that same for the group samples workflow submission.

ericenns:
> @ksierks
> 
> > * when not including a name, true is returned, but no export was created
> 
> * I was unable to recreate this. I get a valid export with `name = nil`. If you're able to recreate this, let me know what params are used!
> 
> > * when including a sample id that does not exist, an error occurs
> 
> * Service now validates each sample ID and will raise an error if it is not found. Also added accompanying test. [9e13a52](https://github.com/phac-nml/irida-next/commit/9e13a52f198994345d07fbfbbeb5c378e5786599)
> 
> > * changing the export type to something other than sample or analyst, true is returned but no export was created
> 
> * Service now checks if the export is `valid?` and will return false if invalid. This will cover if type is incorrect. And then I imagine we'll build upon the false return once the controller is set up. [17698be](https://github.com/phac-nml/irida-next/commit/17698beb8d0b82796ef3722fc14cde4529514cbb)

@ChrisHuynh333 what Katherine was saying is that the service just returns a boolean and not the `@data_export`, I think I agree with her that the service should be returning a `@data_export` regardless of success or failure as we can check the resulting object to know if it is valid and was actually saved. This means that a UI could present errors to the user for example if it failed.

ChrisHuynh333:
> > @ksierks
> > > * when not including a name, true is returned, but no export was created
> > 
> > 
> > 
> > * I was unable to recreate this. I get a valid export with `name = nil`. If you're able to recreate this, let me know what params are used!
> > 
> > > * when including a sample id that does not exist, an error occurs
> > 
> > 
> > 
> > * Service now validates each sample ID and will raise an error if it is not found. Also added accompanying test. [9e13a52](https://github.com/phac-nml/irida-next/commit/9e13a52f198994345d07fbfbbeb5c378e5786599)
> > 
> > > * changing the export type to something other than sample or analyst, true is returned but no export was created
> > 
> > 
> > 
> > * Service now checks if the export is `valid?` and will return false if invalid. This will cover if type is incorrect. And then I imagine we'll build upon the false return once the controller is set up. [17698be](https://github.com/phac-nml/irida-next/commit/17698beb8d0b82796ef3722fc14cde4529514cbb)
> 
> @ChrisHuynh333 what Katherine was saying is that the service just returns a boolean and not the `@data_export`, I think I agree with her that the service should be returning a `@data_export` regardless of success or failure as we can check the resulting object to know if it is valid and was actually saved. This means that a UI could present errors to the user for example if it failed.

Apologies for misinterpreting! Sounds good I've returned the `@data_export` from the service. d81591d552f77d9b432e1791bc72f78dfa4615da

ksierks (approved):
Thanks for the changes!

ericenns (approved):
Looks good to me!

PR #461, opened 2024-03-11
Attachments: PUID

ericenns:

This updates the Attachment model to include PUIDs, for `illumina_pe` and `pe` the forward and reverse individual attachments will have the same PUID.

image

image

image

First ensure that migrations succeed with existing database that has both `pe` and `illumina_pe` files.

Next ensure that uploading PE files have the same PUID and still works as expected.

## PR acceptance checklist

### DIFF
--- a/app/views/projects/samples/attachments/_table.html.erb
+++ b/app/views/projects/samples/attachments/_table.html.erb
@@ -10,6 +10,7 @@
         <% if allowed_to?(:update_sample?, @project) %>
           <th aria-hidden="true" class="px-6 py-3"></th>
         <% end %>
+        <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.puid") %></th>
         <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.filename") %></th>
         <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.format") %></th>
         <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.type") %></th>
--- a/test/test_helpers/axe_helpers.rb
+++ b/test/test_helpers/axe_helpers.rb
@@ -98,6 +98,8 @@ module AxeHelpers
   def visit(path, **attributes)
     super
 
+    page.driver.wait_for_network_idle
+
     assert_accessible
   end
 end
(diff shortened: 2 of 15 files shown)


app/views/projects/samples/attachments/_table.html.erb
```diff
@@ -10,6 +10,7 @@
         <% if allowed_to?(:update_sample?, @project) %>
           <th aria-hidden="true" class="px-6 py-3"></th>
         <% end %>
+        <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.puid") %></th>
         <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.filename") %></th>
         <th scope="col" class="px-6 py-3"><%= t("projects.samples.show.table_header.format") %></th>
```
ChrisHuynh333, line 15: Maybe a question for @joshsadam as well. Just wondering about consistency in styling, if we wanted to match the darkened background of the PUID column on attachments like it exists for samples. As well as if we want the PUID clickable to download like the name.

ChrisHuynh333 (changes requested):
Looks great and works as described. One question regarding the concatenate/delete modals, if we wanted to keep the attachment names in the popup, as it is now contains the PUIDs:
image

app/views/projects/samples/attachments/_table.html.erb
joshsadam: 1.  The dark background on the samples page is just there to handle the horizontal scrolling and to create a difference.  We could update all tables to have their primary column dark, but I don't think that is part of this PR.
2.  Clicking the name to download is something that we should not have had.  It needs to be moved (not in this PR) to where the delete button is.  The delete button should not be an actual button, but just a text link like on the [Flobite page](https://flowbite.com/docs/components/tables/#striped-rows).

joshsadam (approved):
This looks good to me, migrations work.

ChrisHuynh333 (approved):
Looks great to me!

test/test_helpers/axe_helpers.rb
```diff
@@ -98,6 +98,8 @@ def click_button(locator = nil, **kwargs)
   def visit(path, **attributes)
     super
 
+    page.driver.wait_for_network_idle
```
joshsadam, line 101: **praise:** This is an interesting little bit. Might prevent a lot of our issues.  Thanks for looking at these failures

joshsadam (approved):
Thanks for fixing (hopefully) the issue with those tests.  This looks great.

PR #462, opened 2024-03-11
Sample files updated column

deepsidhu85:

This PR adds a new column to the samples table `attachments_updated_at` which gets updated whenever files are added or removed to/from a sample.

image

**No attachments**

image

**Uploaded attachments  - samples table at project level**

image

**Destroyed attachments - samples table at project level**

image

**Samples table at group level**

image

1. Switch to the main branch
2. Reset your database `bin/rails db:drop db:create db:migrate db:seed`
3. Switch to this branch
4. Run `bin/rails db:migrate`
5. Open `rails console` in the terminal
6. Run the query `Sample.first` and make note of the new `attachments_updated_at` column and its timestamp
7. Run the query `Sample.last`. The timestamps for the `attachments_updated_at` columns should be different
8. Login to IRIDA Next 
9. Create a project under a group, add a sample, upload some files to the sample
10. View the project samples by clicking the `Samples` link in the left sidebar
11. Table should display timestamp for `Files Last Updated`
12. View the samples for the group (the group the project was created under in step 9)
13. You should see the table has a column for `Files Last Updated`

**Note:** The sample `updated_at` column is updated alongside the `attachments_updated_at` column when attachments are uploaded/removed for a sample. This is the behaviour of the `touch` option for `belongs_to`. 

## PR acceptance checklist

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -202,6 +202,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_03_13_142828) do
     t.datetime "updated_at", null: false
     t.uuid "workflow_execution_id"
     t.uuid "sample_id"
+    t.datetime "attachments_updated_at"
     t.index ["created_at"], name: "index_samples_workflow_executions_on_created_at"
     t.index ["sample_id"], name: "index_samples_workflow_executions_on_sample_id"
     t.index ["workflow_execution_id"], name: "index_samples_workflow_executions_on_workflow_execution_id"
@@ -244,6 +245,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_03_13_142828) do
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.uuid "submitter_id", null: false
+    t.datetime "attachments_updated_at"
     t.index ["created_at"], name: "index_workflow_executions_on_created_at"
     t.index ["submitter_id"], name: "index_workflow_executions_on_submitter_id"
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -186,7 +186,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_03_13_142828) do
     t.jsonb "metadata_provenance", default: {}, null: false
     t.string "puid", null: false
     t.uuid "project_id", null: false
-    t.datetime "files_updated_at"
+    t.datetime "attachments_updated_at"
     t.index ["created_at"], name: "index_samples_on_created_at"
     t.index ["deleted_at"], name: "index_samples_on_deleted_at"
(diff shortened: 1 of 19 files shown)


ksierks:
Hi Deep. I have a quick question. If I look at a sample with no files, the `files last updated` column is blank. I uploaded and immediately deleted an attachment to/from the sample. The sample has no files, but the`last updated` and `files last updated` columns have the same timestamp now. Is that intentional? Does this inform the user there were attachments on the sample? If so, is there a way to see that in the history like with metadata? I only ask, because if the sample updated_at timestamp is changed should the history capture what the change is?

deepsidhu85:
> Hi Deep. I have a quick question. If I look at a sample with no files, the `files last updated` column is blank. I uploaded and immediately deleted an attachment to/from the sample. The sample has no files, but the`last updated` and `files last updated` columns have the same timestamp now. Is that intentional? Does this inform the user there were attachments on the sample? If so, is there a way to see that in the history like with metadata? I only ask, because if the sample updated_at timestamp is changed should the history capture what the change is?

Hi Katherine,

That is the correct behaviour. A request was made to be able to distinguish when files were last modified for a sample (added or removed). For the history stuff we only display record level changes for a group, project, and sample. We will be adding in an activity feed in the future which will capture the changes to the sample attachments etc

db/schema.rb
```diff
@@ -203,6 +204,7 @@
     t.datetime "updated_at", null: false
     t.uuid "workflow_execution_id"
     t.uuid "sample_id"
+    t.datetime "attachments_updated_at"
```
ksierks, line 207: I don't know a lot about data exports, but does it make any sense to add the new column "attachments_updated_at" to it? That's the only other model I can think of that has attachments.

deepsidhu85: I don't think we need it for the data exports. It's moreso for the tables where we want to show when the files were last updated for a sample

ksierks (approved):
I think this is great! Everything works as described (except sorting the files last updated table column). I wasn't able to test workflow attachments. Thanks!

db/schema.rb
ericenns: It isn't needed for data exports as that uses ActiveStorage::Attachment and not our Attachment model as it is just storing a single zip file which we don't need extended metadata for. The column is only needed for models which have `has_many :attachments, as: attachable...`

ksierks: Does that mean the column shouldn't be on workflow executions?

ericenns: Currently workflow_executions doesn't have that setup but I asked Deep to add it here so that Jeff can utilize it in the workflow completion service.

ericenns (approved):
Looks good to me, Sorting is now working as I would expect

PR #463, opened 2024-03-12
Group Project filter by list

joshsadam:

## PR acceptance checklist


ericenns (approved):
I tested this out and it is working as expected!

config/locales/en.yml
```diff
@@ -330,6 +330,12 @@ en:
       previous_version: "Previous (Version %{version})"
       current_version: "Current (Version %{version})"
       keys_deleted: "Keys deleted:"
+    list_filter:
+      title: Filter by sample names or Id
+      clear: Remove all tags
+      description: 'Paste a list of <kbd>,</kbd> seperated sample names or identifiers'
```
deepsidhu85, line 336: Can you change `identifiers` to `persistent unique identifiers (puid)`. or something similar. Just for clarity for the user since samples have ids as well

app/components/list_filter_component.html.erb
```diff
@@ -0,0 +1,150 @@
+<div
+  data-controller="list-filter"
+  data-list-filter-selection-outlet="#samples-table"
+>
+  <%= viral_dialog(closable: false) do |dialog| %>
```
deepsidhu85, line 5: Any specific reason for setting the dialog to be non closable? A user can accidentally click the button and it launches the dialog but there's no way for the user to close it. You can close it using the `Apply Filter` button with an empty text field but a user would have no way of knowing that. I think we should add a `X` to close the dialog or a `Cancel` button

deepsidhu85 (changes requested):
This is looking /working good. Just a couple of comments below

config/locales/en.yml
joshsadam: The table header has it as `ID` and I don't know if the user would know or care about the implementation of it as a PUID.

app/components/list_filter_component.html.erb
joshsadam: This was decided by @ericenns and myself when we did the project samples table. The  reason for this is that the `tags` are actually the form  `inputs` if a user removes one and then closes the dialog without the table update then the next time the table changes (sort, paging, etc...) the list to filter by will be different, loading the samples into the table.

config/locales/en.yml
deepsidhu85: I guess you're right...

app/components/list_filter_component.html.erb
deepsidhu85: That makes sense but I still think we need to revisit this and see if we can solve it another way. To me it's just a bad user experience to have to apply the filter when they might have just accidentally launched the modal. I'll approve for now and we can revisit whenever we have a refactor phase

deepsidhu85 (approved):
Looks good to me!

PR #523, opened 2024-04-08
Bot accounts model service setup

deepsidhu85:

This PR adds in the db, model, and service layer for namespace bot accounts.

1. Create a new account for IRIDA next from the UI, login, and create a project
2. Open rails console in the terminal
3. Set the following `user = User.last`, `proj_namespace = Namespaces::ProjectNamespace.last`, `params = {"token_name": "Uploader", "scopes":["read_api", "api"], "access_level":20}`
4. Run `Bots::CreateService.new(user, proj_namespace, params).execute`
5. Verify that a new user was created with a user_type of `1` which is a `project_bot`
6. Verify that the user is linked to the namespace via NamespaceBot
7. Set the last `NamespaceBot` to a variable
8. Run `Bots::DestroyService.new(VARIABLE_FROM_STEP_7, proj_namespace, user)
9. Verify that the NamespaceBot was destroyed and the underlying user from the `users` table. Also verify that the namespace is not deleted.

## PR acceptance checklist

### DIFF
--- a/app/models/namespace_bot.rb
+++ b/app/models/namespace_bot.rb
@@ -13,7 +13,7 @@ class NamespaceBot < ApplicationRecord
   validate :validate_bot_user_type
 
   def validate_bot_user_type
-    return unless user.user_type == User.user_types[:human]
+    return unless User.user_types[user.user_type] == User.user_types[:human]
 
     errors.add(:base, 'Human user cannot be set as a bot user for a namespace')
   end
--- a/app/models/namespace_bot.rb
+++ b/app/models/namespace_bot.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# entity class for NamespaceBot
+class NamespaceBot < ApplicationRecord
+  has_logidze
+  acts_as_paranoid
+
+  belongs_to :user, class_name: 'User'
+  belongs_to :namespace, class_name: 'Namespace'
+
+  validates :user_id, uniqueness: { scope: [:namespace_id] }
+end
--- a/app/models/namespace_bot.rb
+++ b/app/models/namespace_bot.rb
@@ -5,8 +5,16 @@ class NamespaceBot < ApplicationRecord
   has_logidze
   acts_as_paranoid
 
-  belongs_to :user, class_name: 'User'
+  belongs_to :user, class_name: 'User', dependent: :destroy
   belongs_to :namespace, class_name: 'Namespace'
 
   validates :user_id, uniqueness: { scope: [:namespace_id] }
+
+  validate :validate_bot_user_type
+
+  def validate_bot_user_type
+    return unless user.user_type == User.user_types[:human]
+
+    errors.add(:base, 'Human user cannot be set as a bot user for a namespace')
+  end
 end
--- a/app/models/namespace_bot.rb
+++ b/app/models/namespace_bot.rb
@@ -13,7 +13,7 @@ class NamespaceBot < ApplicationRecord
   validate :validate_bot_user_type
 
   def validate_bot_user_type
-    return unless User.user_types[user.user_type] == User.user_types[:human]
+    return unless user.human?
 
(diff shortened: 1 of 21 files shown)


app/models/user.rb
```diff
@@ -36,6 +41,8 @@ class User < ApplicationRecord
 
   delegate :full_path, to: :namespace
 
+  attribute :user_type, :integer, default: -> { User.user_types[:human] }
```
ericenns, line 44: **nitpick (blocking):** Can you move this line into the `has_user_type` model concern.  I believe it should work in the `included do` block.

app/services/bots/destroy_service.rb
ericenns, line 1: **nitpick (blocking):** Don't we also need to destroy the `NamespaceBot` entity.

app/models/user.rb
```diff
@@ -23,6 +27,7 @@ class User < ApplicationRecord
 
   has_many :personal_access_tokens, dependent: :destroy
   has_many :data_exports, dependent: :destroy
+  has_many :namespace_bots, class_name: 'NamespaceBot', dependent: :destroy
```
ericenns, line 30: **nitpick (blocking):** Do you want to add a `has_many :bots, through: :namespace_bots ...` so that you can get the bots directly? Or do you just want to pass the `NamespaceBot` to the destroy service potentially so that it and the bot can be deleted?

ericenns (changes requested):
Looking good, a few comments.

But also I think you need to add NamespaceBot fixtures.

app/services/bots/destroy_service.rb
deepsidhu85: Yup the plan is to pass in the namespace_bot and destroy that which will also destroy the underlying bot (user) and its associations

app/models/user.rb
deepsidhu85: Updated in adfe4001c792f2935730fcb01ed3c9c93771ea9a

app/services/bots/destroy_service.rb
deepsidhu85: Updated the destroy service to accept a namespace_bot in adfe4001c792f2935730fcb01ed3c9c93771ea9a

app/models/user.rb
deepsidhu85: After going through it, it makes sense to pass the namespacebot to the destroy service which will destroy the entity and the underlying user and it's associations (PAT, Member). Updated in adfe4001c792f2935730fcb01ed3c9c93771ea9a. I also added a has_many in this commit to the project_namespace but removed it in the commit after as I don't think we will be needing it (if we do we can always add it back)

app/models/namespace_bot.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  belongs_to :user, class_name: 'User', dependent: :destroy
+  belongs_to :namespace, class_name: 'Namespace'
+
+  validates :user_id, uniqueness: { scope: [:namespace_id] }
+
+  validate :validate_bot_user_type
+
+  def validate_bot_user_type
+    return unless User.user_types[user.user_type] == User.user_types[:human]
```
ericenns, line 16: **nitpick (blocking):** I think you can just do `return unless user.human?` since we are using enums

app/models/user.rb
```diff
@@ -104,11 +109,19 @@ def build_namespace_path
   end
 
   def ensure_namespace
+    return unless User.user_types[user_type] == User.user_types[:human]
```
ericenns, line 112: **nitpick (blocking):** Same as above, I think this can be `return unless user.human?`

ericenns (changes requested):
Looking good, just another couple comments

app/models/user.rb
deepsidhu85: Updated in 8929d48319a2c018f27c667dec548413e9eb6af2

ericenns (approved):
Looks good to me!

PR #546, opened 2024-04-17
Workflow executions: Requeue jobs

### LINKED ISSUE #513: WorkflowExecution: Requeue jobs when ga4gh server is down
When the job server responds with a 400 code, requeue the job with a delay.
* on submission of a job
* on getting the status of a submitted job
* can use https://github.com/bensheldon/good_job?tab=readme-ov-file#retries

For non 400 code errors, keep track of number of times a job has been requeued, after a set number, set an error_code

Add an error_code column to WorkflowExecutions 
Make error_code an enum with translations for human readable error descriptions

JeffreyThiessen:
Fixes #513

1. Adds retry and error rescue to jobs that connect to ga4gh wes. Details below.

2. Fixes a race condition bug where a job that a user cancelled sometimes finishes and completes(writes to samples) instead of cancelling.

3. Created a new set of helper functions `ActiveJobTestHelpers` to allow mocking Faraday based API integrations for active job tests.

You will need to have `Sapporo` running and connected to test

The following 3 jobs reach out to ga4gh wes.

1. WorkflowExecutionStatusJob
2. WorkflowExecutionSubmissionJob
3. WorkflowExecutionCancelationJob

If ga4gh wes is unreachable or returns an error code, it must be handled so that the workflow execution doesn't get stuck in whatever state it was in when the error occurred.

#### For ConnectionError type
This error type should always be reattempted.

You can simply turn off your Sapporo instance at the different stages of the workflow execution.

Verify that the job is reattempted continually.

#### For APIExceptionError types 
These error types should be attempted 5 times before being put into the error state

You can make workflow_execution services throwto test the job re-queuing by adding a line to throw the exception. For example:

```ruby
# app/services/workflow_executions/submission_service.rb#execute
def execute
      return false unless @workflow_execution.prepared?

      raise Integrations::ApiExceptions::BadRequestError, 'my thrown exception'

      run = @wes_client.run_workflow(**@workflow_execution.as_wes_params)
...
```

#### WorkflowExecutionCancelationJob

WorkflowExecutionCancelationJob is a special case because ga4gh wes can return 401 and/or 403 errors when a cancel request is sent to a completed run (Sapporo always replies with 200). 

In this case, we want to mark those runs as `canceled` and not move on to completion steps. Other errors should still behave as above.

## PR acceptance checklist

### DIFF
--- a/app/jobs/workflow_execution_cancelation_job.rb
+++ b/app/jobs/workflow_execution_cancelation_job.rb
@@ -4,6 +4,14 @@
 class WorkflowExecutionCancelationJob < ApplicationJob
   queue_as :default
 
+  # When server is unreachable, continually retry
+  retry_on Integrations::ApiExceptions::ConnectionError, wait: :exponentially_longer, attempts: Float::INFINITY
+  # TODO: retry on 401
+  # TODO: retry on 403
+  # TODO: edge case where canceling a workflow that is actually completed (delay caused)
+  # expects 401/403. Sapporo responds with 200
+  # Need to check run status in retry block
+
   def perform(workflow_execution, user)
     wes_connection = Integrations::Ga4ghWesApi::V1::ApiConnection.new.conn
     WorkflowExecutions::CancelationService.new(workflow_execution, wes_connection, user).execute
--- a/app/jobs/workflow_execution_cancelation_job.rb
+++ b/app/jobs/workflow_execution_cancelation_job.rb
@@ -7,6 +7,16 @@ class WorkflowExecutionCancelationJob < ApplicationJob
   # When server is unreachable, continually retry
   retry_on Integrations::ApiExceptions::ConnectionError, wait: :exponentially_longer, attempts: Float::INFINITY
 
+  # 401
+  rescue_from Integrations::ApiExceptions::UnauthorizedError do |job, exception|
+    handle_completed_run_errors(job, exception)
+  end
+
+  # 403
+  rescue_from Integrations::ApiExceptions::ForbiddenError do |job, exception|
+    handle_completed_run_errors(job, exception)
+  end
+
   # Puts workflow execution into error state and records the error code
   retry_on Integrations::ApiExceptions::APIExceptionError, wait: :exponentially_longer, attempts: 5 do |job, exception|
     workflow_execution = job.arguments[0]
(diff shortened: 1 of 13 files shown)


app/jobs/workflow_execution_cancelation_job.rb
```diff
@@ -4,6 +4,33 @@
 class WorkflowExecutionCancelationJob < ApplicationJob
   queue_as :default
 
+  # When server is unreachable, continually retry
+  retry_on Integrations::ApiExceptions::ConnectionError, wait: :polynomially_longer, attempts: Float::INFINITY
+
+  # Puts workflow execution into error state and records the error code
+  retry_on Integrations::ApiExceptions::APIExceptionError, attempts: 3 do |job, exception|
```
ericenns, line 11: **nitpick:** Can we set the `wait` for this as by default it appears to be `3.seconds` (https://github.com/phac-nml/irida-next/pull/546/files)

I'm not sure is `15.seconds` is a good choice or if we set it to `:polynomially_longer` with `attempts: 3`. Since that would be (first wait ~3s, then ~18s, then ~83s, etc).

db/migrate/20240418180337_add_error_code_to_workflow_execution.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Migration to add error_code column to workflow_executions table
+class AddErrorCodeToWorkflowExecution < ActiveRecord::Migration[7.1]
+  def change
+    add_column :workflow_executions, :error_code, :integer
```
ericenns, line 6: **nitpick (blocking):** can we change this to `http_error_code` since we may want to add an `error_code` in the future that is an enum like `CommunicationError` etc that is not entirely related to `http_error_code`.

ericenns (changes requested):
This looks pretty good, just a few comments.

db/migrate/20240418180337_add_error_code_to_workflow_execution.rb
JeffreyThiessen: Changed in https://github.com/phac-nml/irida-next/pull/546/commits/ebfe278e975461c66edf27a7560aad0a453916eb

app/jobs/workflow_execution_cancelation_job.rb
JeffreyThiessen: Changed in https://github.com/phac-nml/irida-next/pull/546/commits/ebfe278e975461c66edf27a7560aad0a453916eb

I opted for `:polynomially_longer`

ericenns (approved):
Looks good to me!

PR #577, opened 2024-04-26
Workflow Executions: Update Model Tests

ChrisHuynh333:
This PR updates testing the workflow execution model to increase coverage

1. Ensure all tests pass locally

## PR acceptance checklist

### DIFF
--- a/test/models/workflow_execution_test.rb
+++ b/test/models/workflow_execution_test.rb
@@ -200,43 +200,40 @@ class WorkflowExecutionTest < ActiveSupport::TestCase
     workflow_execution_example.state = :submitted
     assert_enqueued_emails 0 do
       workflow_execution_example.send_email
-      assert_enqueued_email_with PipelineMailer, :complete_email, args: [workflow_execution_example]
     end
 
     workflow_execution_example.state = :running
     assert_enqueued_emails 0 do
       workflow_execution_example.send_email
-      assert_enqueued_email_with PipelineMailer, :complete_email, args: [workflow_execution_example]
     end
 
     workflow_execution_example.state = :completing
     assert_enqueued_emails 0 do
       workflow_execution_example.send_email
-      assert_enqueued_email_with PipelineMailer, :complete_email, args: [workflow_execution_example]
     end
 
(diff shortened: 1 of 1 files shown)


test/models/workflow_execution_test.rb
```diff
@@ -94,4 +94,173 @@ def setup
...
+    assert @workflow_execution_valid.sent_to_ga4gh?
+
+    @workflow_execution_valid.state = :prepared
+    assert_not @workflow_execution_valid.sent_to_ga4gh?
+
+    @workflow_execution_valid.state = :initial
+    assert_not @workflow_execution_valid.sent_to_ga4gh?
+  end
+
+  test 'send_email' do
```
ericenns, line 185: **nitpick (blocking):** If I run this test case it reports no assertions have been made. If I refactor lines 190 - 194 with the following I actually see assertions.

```ruby
assert_enqueued_emails 1 do
      workflow_execution_completed.send_email
      assert_enqueued_email_with PipelineMailer, :complete_email, args: [workflow_execution_completed]
    end
```

Note: this example will fail as the fixture you are using for this test does not have `email_notification: true`.

Can you refactor all of these case in your test. The issue is that `workflow_execution_completed.send_email do` is not valid syntax. That would only work if the method accepted a block but then again you wouldn't pass test assertions to a block.

ericenns (changes requested):
This is looking better except that the tests for `send_email` are not working as you intended.

test/models/workflow_execution_test.rb
ChrisHuynh333: Thanks for catching this. Fixed in 9eaca37fe764f2a38bec728a104ef44e632e25b5

ChrisHuynh333: I realized it doesn't make much sense to have `assert_enqueued_email_with PipelineMailer, :complete_email, args: [workflow_execution_example]` in blocks that have 0 enqueued emails. Removed in 00276fea505da70d2136e5c087deb906a3f2f6bd

ericenns (approved):
Looks good to me!

PR #612, opened 2024-05-21
Fix: PAT Copy to clipboard on Firefox (DFCT0010063)

ericenns:

This PR fixes the clipboard component so that it functions in browsers other than chrome when value is hidden.

Previously the copy to clipboard component used javascript to verify that the javascript is first allowed to write to the clipboard before attempting to write to the clipboard. This permission verification only works on Chrome, no other browsers respond to this. But since this is a user driven action we do not need to verify permission as it is automatically granted. The fallback behavior was to select the input and then initiate copy, but since the input stores *********** when hidden we cannot use that as it only works after the user clicks the eye icon to show it on screen. Since all the other browsers support `Clipboard.writeText()` we can just call the method directly and remove the fallback behavior after removing browser permission verification.

1. Update asdf ruby plugin
```bash
asdf plugin update ruby
```
2. Ensure asdf dependencies updated
```bash
asdf install
```
3. Install gems
```bash
bundle
```
4. Migrate database
```bash
bin/rails db:migrate
```
5. Launch server
```bash
bin/dev
```
6. Launch firefox and generate a new personal access token, and click copy to clipboard. Paste into a blank document and ensure content is not `********************`.

## PR acceptance checklist


joshsadam (approved):
Seems to work now in both Firefox and Chrome.  Thanks for the fix.

PR #620, opened 2024-05-29
GraphQL: Add projectSample query [DFCT0010065]

ericenns:
## What does this PR do and why?
_Describe in detail what your merge request does and why._

This PR adds in a new query named `projectSample` which returns a `Sample` if found within a `Project`. The query find the `Sample` by either its puid or its name.

## Screenshots or screen recordings
_Screenshots are required for UI changes, and strongly recommended for all other pull requests._

## How to set up and validate locally
_Numbered steps to set up and validate the change are strongly suggested._

1. Launch server and login as your preferred user
2. Navigate to `GraphiQL` endpoint (e.g. `http://localhost:3000/graphiql`)
3. Try the `projectSample` query using sample puid
```graphql
{
  projectSample(projectPuid: "PROJECT_PUID", samplePuid: "SAMPLE_PUID") {
    name
    puid
  }
}
```
4. Try the `projectSample` query using sample name
```graphql
{
  projectSample(projectPuid: "PROJECT_PUID", sampleName: "SAMPLE NAME") {
    name
    puid
  }
}
```
5. Try querying for a Sample which belongs to a different project and ensure no result

## PR acceptance checklist
This checklist encourages us to confirm any changes have been analyzed to reduce risks in quality, performance, reliability, security, and maintainability.

- [x] I have evaluated the [PR acceptance checklist](https://phac-nml.github.io/irida-next/docs/development/development_processes/code_review#acceptance-checklist) for this PR.

PR #627, opened 2024-06-03
Uploader policy update (DFCT0010069)

deepsidhu85:

This PR updates policy methods for a user with uploader access level to be able to create a sample, modify a sample (create/remove files, update sample metadata), and read projects

image

image

1. Start up irida next
2. Login as any user
3. Create a project and a group
4. Add bot accounts to both with an `UPLOADER` access level. Copy down the tokens
5. Generate new tokens for both bot accounts and set the Expiration to an earlier date. Copy down the token
6. Use `Postman` to test out the graphql queries
7. In Postman click the icon/button next to `Untitled Request` and select `Graphql`
8. In the `URL` input, type in `http://localhost:3000/api/graphql`
9. Click on the `authorization` tab. Here you can run the query/mutation with the bot accounts using their email and the tokens as the password. Make sure to try out all the queries/mutations using the valid and expired tokens for each bot account. Verify the bot account with the `UPLOADER` access level can query/mutate if they have a valid token, otherwise they should not be able to.
10. Take a look in the `test/graphql/` directory for all the queries and mutations and to see how the data is passed in to each query and mutation. 

## PR acceptance checklist

### DIFF
--- a/app/graphql/irida_schema.rb
+++ b/app/graphql/irida_schema.rb
@@ -38,6 +38,11 @@ class IridaSchema < GraphQL::Schema # rubocop:disable GraphQL/ObjectDescription
   # Stop validating when it encounters this many errors:
   validate_max_errors 100
 
+  def self.execute(query_str = nil, **kwargs)
+    Current.token = kwargs[:context][:token]
+    super
+  end
+
   # Relay-style Object Identification:
 
   # Return a string UUID for `object`
--- a/app/graphql/mutations/attach_files_to_sample.rb
+++ b/app/graphql/mutations/attach_files_to_sample.rb
@@ -23,7 +23,8 @@ module Mutations
                else
                  Sample.find_by(puid: args[:sample_puid])
                end
-      files_attached = Attachments::CreateService.new(current_user, sample, { files: args[:files] }).execute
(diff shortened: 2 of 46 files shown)


JeffreyThiessen (changes requested):
Overall looks great!

Can you make a separate error message for attempting to use an expired token?
Right now the following message is returned `Unable to access object while accessing the API in guest mode`, which doesn't make much sense for this case.

ericenns:
> Overall looks great!
> 
> Can you make a separate error message for attempting to use an expired token? Right now the error returned is `Unable to access object while accessing the API in guest mode`, which doesn't make much sense for this case.

I added that error message in PR [620](https://github.com/phac-nml/irida-next/pull/620). At that stage, the token would have already failed authentication and when that happens the User is placed in guest mode. The same thing happens if the User attempts to use the API without a token. So when the error is thrown there is no context of having an expired token as it wont be set. So long ways around to saying that we can't change this error message, as it is can be thrown when accessing specific API endpoints that require authorization while attempting to use an expired token or while not providing any token at all.

JeffreyThiessen (approved):
:+1:

ericenns (approved):
Looks good to me!

PR #657, opened 2024-06-27
[ENHC0010037] Add member table sorting and searching

### LINKED ISSUE #641: Members: Add searching and sorting to table (ENHC0010037)
## Summary

Currently, the members table for both groups and projects does not have sorting or searching capability. We want to update the table to add this functionality to allow the user to sort and search

## Acceptance Criteria

- Since the members listing pages are pretty much identical, convert these over to be a component which will be shared by both the project and groups member index templates
- Search functionality added ( searches username ~~and access level~~)
- Sort functionality added for all columns except `Access Granted` and `Actions`

Note: Take a look at the files within `app/components/samples` for an example of how this should work.

ksierks:
Added sorting and searching to group & project member tables.
Fixes #641.

image
image

1. Navigate to a project or group members page.
2. Use the search bar to filter by username.
3. Click on a table header to sort the table.

## PR acceptance checklist


joshsadam:
There is something wrong with the way the `Access Level` is being sorted, can you see if you can figure out what is causing it?

ksierks:
> There is something wrong with the way the `Access Level` is being sorted, can you see if you can figure out what is causing it?
> 
> 

Maybe I'm misunderstanding, but it's sorting the access level by enum value. Guest (10), Analyst (30), Maintainer (40), and Owner (50) is the sorted order.

joshsadam (approved):
This looks good to me.

ChrisHuynh333 (approved):
Looks great!

app/controllers/concerns/membership_actions.rb
```diff
@@ -16,9 +16,13 @@ module MembershipActions # rubocop:disable Metrics/ModuleLength
   def index
     authorize! @namespace, to: :member_listing?
     respond_to do |format|
```
deepsidhu85, line 18: We can set the @q, default sort, and the @pagy, @members stuff before the respond_to block and then within it all you would require is format.html and format.turbo_stream. When setting @q you can just use `@q = load_members.ransack(params[:q])`

The `load_members` method also needs to be updated to convert the array into an activerecord collection so that it will work with ransack. This can be accomplished with `MyModel.where(id: arr.map(&:id))`

app/views/groups/members/_access_level.html.erb
```diff
@@ -9,8 +9,8 @@
         id: "member-#{member.id}-access-level-select",
         value: member.access_level,
         onchange: "this.form.requestSubmit();",
-        "aria-label": "Access Level"
-      }
+        "aria-label": "Access Level",
```
deepsidhu85, line 12: We should localize the aria-label text here and the projects one as well

app/views/groups/members/index.html.erb
```diff
@@ -23,10 +26,29 @@
   <% end %>
 <% end %>
 
-<div class="bg-white dark:bg-slate-800">
-  <%= turbo_frame_tag "new_member_modal" %>
-  <%= turbo_frame_tag "member-update-alert" %>
+<div class="flex flex-row-reverse">
+  <%= search_form_for @q, url: group_members_url(**request.query_parameters), html: {"data-controller": "filters"} do |f| %>
```
deepsidhu85, line 30: The search form breaks if you are on the `Groups` tab and try to do a search. If you clear the text box the table for members is displayed but the `Groups` tab is selected.

I think it may be good to create a members search component which will be used both by the members and groups tabs for both groups and projects

deepsidhu85 (changes requested):
This is looking good. Just have a few comments below

deepsidhu85:
Also looks like this spacing needs to be adjusted (between the alert and the title)

image

deepsidhu85 (changes requested):
We also need to add some testing for the new component. Take a look under `test/components/` to see examples on how the testing would look

app/controllers/concerns/membership_actions.rb
deepsidhu85: As discussed we don't require this and can just update the member_policy scope

app/controllers/concerns/membership_actions.rb
```diff
@@ -34,13 +38,15 @@ def new
...
 
-  def create # rubocop:disable Metrics/MethodLength
+  def create # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
     @new_member = Members::CreateService.new(current_user, @namespace, member_params, true).execute
 
     if @new_member.persisted?
       respond_to do |format|
         format.turbo_stream do
-          @pagy, @members = pagy(load_members)
+          @q = Member.where(id: load_members.pluck(:id)).ransack(params[:q])
```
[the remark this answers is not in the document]
ksierks: ba42991e71a3f2570b9af131a11fbacb8a9adc5b

app/controllers/concerns/membership_actions.rb
ksierks: 53a6024210cf0bc7d16da260fed1a23a5f3a4a1c

app/views/groups/members/_access_level.html.erb
ksierks: 5f7e635939d0be815d43b61f272ea197c105c9e0

ksierks:
> Also looks like this spacing needs to be adjusted (between the alert and the title)
> 
> image

Like this, 9bf3635f3712dfbf67ab52a042c04f73a34aa1fe?

deepsidhu85:
> > Also looks like this spacing needs to be adjusted (between the alert and the title)
> > image
> 
> Like this, [9bf3635](https://github.com/phac-nml/irida-next/commit/9bf3635f3712dfbf67ab52a042c04f73a34aa1fe)?

Looks like the image link is broken

ksierks:
> We also need to add some testing for the new component. Take a look under `test/components/` to see examples on how the testing would look

9cf18c33ec56c53ff22bc70171f1e79f558a3c5a

app/views/groups/members/_current_user_is_member.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= link_to(
   t(:"groups.members.index.leave_group"),
-  group_member_path(id: member.id),
+  group_member_path(namespace, id: member.id),
```
deepsidhu85, line 3: You shouldn't need to put `id: member.id` and can instead just replace the second argument with `member`. Same applies to `current_user_is_not_member.html.erb`

test/components/members/table_component_test.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module Members
+  class TableComponentTest < ViewComponent::TestCase
+    test 'Should render a table of members' do
+      namespace = groups(:group_one)
+      user = users(:john_doe)
+      members = Member.for_namespace_and_ancestors(namespace).not_expired
+
+      with_request_url '/-/groups/group-1/-/members' do
+        q = Member.ransack({ s: 'user_email asc', user_email_cont: 'adm' })
```
deepsidhu85: I think this should be done in a separate test case since you are filtering the listing. This test should just display all the members and assert that the correct info is displayed

app/views/groups/members/_current_user_is_member.html.erb
deepsidhu85: Same will also apply to the same partials in project members

deepsidhu85 (changes requested):
Looking good just a few comments below

app/controllers/concerns/membership_actions.rb
```diff
@@ -38,9 +39,11 @@ def create # rubocop:disable Metrics/MethodLength
     @new_member = Members::CreateService.new(current_user, @namespace, member_params, true).execute
 
     if @new_member.persisted?
+      # @q = load_members.ransack(params[:q])
```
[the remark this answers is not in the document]
ksierks: 283bcd6c7392c000eeedad5757814e82c5f83b49

app/views/groups/members/_current_user_is_member.html.erb
ksierks: 20515370abc9a4fdf4d55d8a6b7a8fbe5a43aed9

test/components/members/table_component_test.rb
ksierks: I wasn't trying to test searching and sorting here. That has been done within the system tests already. This component renders the members that are passed to it. If I pass q.result as the members list to the component, I would just be checking that the passed list is the same list that is being rendered and I'm already doing that. Maybe I should remove the q parameters instead?

deepsidhu85 (approved):
Looks good to me!

PR #658, opened 2024-06-27
[DFCT0010103] Fix blob deletion bug and allow cron job configuration

JeffreyThiessen:

1. Fixes a bug where when a `WorkflowExecution` in the `new` state is canceled by the user, `nil` would be passed into the blob service `delete_prefixed`, resulting in an `rm *` effect deleting all the blobs.

Now new runs that are canceled will not schedule a cleanup job or a cancelation job (which included a cleanup step).
Additionally a safety check has been added to the cleanup job such that `nil` and `""` cannot be passed into the function.

2. We want to be able to enable and disable the automatic cron cleanup jobs.

Adds ENV variable `ENABLE_CRON`, when `"true"` cron jobs will run, when `"false"` cron jobs will not run.
Default is `"true"`

Adds ENV variable `CRON_CLEANUP_AFTER_DAYS`. This determines how many days old an attachment or sample must be before it is cleaned by the cron jobs.
Default is `7` days

## PR acceptance checklist


Takadonet (dismissed):

test/services/workflow_executions/cleanup_service_test.rb
```diff
@@ -47,15 +54,95 @@ def setup # rubocop:disable Metrics/MethodLength
...
+      end
+
+      @workflow_execution = WorkflowExecutions::CleanupService.new(@workflow_execution, @user, {}).execute
+
+      assert_nothing_raised do
+        @output_file_blob.download
+        @input_file_blob.download
+        @samplesheet_file_blob.download
+        @unrelated_file_blob.download
+      end
```
deepsidhu85, line 87: For this test and the ones below, we should assert that the workflow execution is marked as `cleaned` after the cleanup service is called

deepsidhu85 (changes requested):
The code is looking good just a few comments below

ChrisHuynh333 (dismissed):
Works as described! Nothing to add, I will test again if the changes Deep suggested are made.

test/services/workflow_executions/cleanup_service_test.rb
JeffreyThiessen: Added in https://github.com/phac-nml/irida-next/pull/658/commits/f271c7de42405ee78590a231b86496f64f6ad346

deepsidhu85 (approved):
Looks good to me!

PR #664, opened 2024-07-10
Linelist Export [STRY0011998]

joshsadam:

## PR acceptance checklist


ksierks (dismissed):
I did a quick once over... I think this looks great!

app/javascript/controllers/sortable_list_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from "@hotwired/stimulus"
```
deepsidhu85, line 1: There's some strange behaviour when trying to sort a list which is scrollable. When reordering element `A` from the top to the bottom of the list, the scrolling only works if you center the list item correctly in the div and then try to move it. Looks like we should be able to get this working correctly by using Autoscroll. See https://github.com/SortableJS/Sortable/blob/master/plugins/AutoScroll/README.md

app/javascript/controllers/data_exports/metadata_selection_controller.js
```diff
@@ -0,0 +1,69 @@
...
+    this.fullListItems = this.availableList.querySelectorAll("li")
+    this.selectedList.addEventListener("mouseover", () => { this.#checkSubmitState() })
+  }
+
+  addAll() {
+    for (const item of this.fullListItems) {
+      this.selectedList.append(item)
+    }
+
+    this.#setSubmitButtonDisableState(false)
```
deepsidhu85, line 29: When a user adds all items to the selected list, we should disable the link for `Add All` link as all the fields have been added. Same applies if there is no items in the selected list, the `Remove All` link should be disabled

deepsidhu85 (changes requested):
This is looking good just have a couple of comments below

app/javascript/controllers/sortable_list_controller.js
ChrisHuynh333: Great catch, added in 2c820683116e72604aea8b4d070622519399006a.
Let me know if the scroll still doesn't feel good, we can adjust the speed and how far from the div edge the element has to be before scrolling begins

app/javascript/controllers/sortable_list_controller.js
```diff
@@ -0,0 +1,27 @@
...
+    },
+    scrollSpeed: {
+      type: Number,
+      default: 8,
+    }
+  }
+  connect() {
+    this.sortable = new Sortable(this.element, {
+      scroll: true,
+      forceFallback: true,
```
deepsidhu85, line 19: Can you remove this forceFallback as its causing some weird behaviour where the elements the list item that is selected is hovered over get highlighted. Removing this fixes the issue

app/components/viral/sortable_list/list_component.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        id="<%= list_item %>"
+        class="
+        block
+        w-full
+        px-4
+        py-2
+        border-b
+        border-slate-200
+        cursor-move
+        hover:bg-slate-100
```
deepsidhu85, line 21: Can you remove the hover: and dark:hover: classes as this is also causing some strange behaviour when dragging a list item from the available to selected columns. The next list item gets highlighted as you are dragging the previous item over

deepsidhu85 (changes requested):
This is almost good to go, couple of small comments

app/javascript/controllers/sortable_list_controller.js
deepsidhu85: This is working great!

app/javascript/controllers/sortable_list_controller.js
ChrisHuynh333: Removed in eed8e25f89304aabec0116cd5d92acc7eab71a95

app/components/viral/sortable_list/list_component.html.erb
ChrisHuynh333: Removed in eed8e25f89304aabec0116cd5d92acc7eab71a95

deepsidhu85 (approved):

joshsadam (commented):
This looks really good, I just have two suggestions/questions

app/components/viral/sortable_list/list_component.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="<%= @system_arguments[:container_classes] %>">
+  <%= title %>
+  <ul
+    id="<%= id %>"
+    data-controller="sortable-list"
+    data-sortable-list-group-name-value="<%= group %>"
+    class="<%= @system_arguments[:list_classes] %>"
+    >
+    <% unless list_items.empty? %>
```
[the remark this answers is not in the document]
ChrisHuynh333: We would want the possibility of an empty list, much like the lists in this PR where the selected list starts empty. My unless statement was to avoid doing an `.each` loop on an empty list, but it looks like it works fine on an empty array, so I could take out the `unless` statement.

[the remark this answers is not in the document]
ChrisHuynh333: Ended up removing the `unless` statement 287ef3ce83906309846f4ec9d91b4430c3a826fb

deepsidhu85 (approved):

PR #668, opened 2024-07-17
[ENHC0010037] Member search component

ksierks:
Moved the members search box into a new component.
No changes to the group & project member table sorting and searching functionality.
Depends on #657.

image
image

1. Navigate to a project or group members page.
2. Use the search bar to filter by username.

See http://localhost:3000/rails/lookbook/inspect/search/default for preview.

## PR acceptance checklist


app/views/groups/members/index.html.erb
```diff
@@ -26,28 +26,7 @@
...
-      <%= turbo_frame_tag "member_hidden_values" %>
-      <%= f.search_field :user_email_cont,
-                     "data-action": "filters#submit",
-                     class:
-                       "block w-full p-2.5 pl-10 text-sm text-slate-900 border border-slate-300 rounded-lg bg-slate-50 focus:ring-primary-500 focus:border-primary-500 dark:bg-slate-700 dark:border-slate-600 dark:placeholder-slate-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500",
-                     placeholder: t(".search.placeholder") %>
-    </div>
-  <% end %>
-</div>
+<%= render Members::SearchComponent.new(@q, @tab, @namespace) %>
```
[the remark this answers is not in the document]
deepsidhu85: Then in the `groups/members/index.turbo_stream.erb` file you can update the turbo frame tag in the `groups/members/index.html.erb` file to render the search component with the attributes:

  q: @q,
  url: group_members_path,
  search_attribute: "user_email_cont",
  tab: @tab,
  namespace: @namespace,
  placeholder: t(".search.placeholder"),

app/views/projects/members/index.html.erb
```diff
@@ -34,28 +34,7 @@
...
-      <%= turbo_frame_tag "member_hidden_values" %>
-      <%= f.search_field :user_email_cont,
-                     "data-action": "filters#submit",
-                     class:
-                       "block w-full p-2.5 pl-10 text-sm text-slate-900 border border-slate-300 rounded-lg bg-slate-50 focus:ring-primary-500 focus:border-primary-500 dark:bg-slate-700 dark:border-slate-600 dark:placeholder-slate-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500",
-                     placeholder: t(".search.placeholder") %>
-    </div>
-  <% end %>
-</div>
+<%= render Members::SearchComponent.new(@q, @tab, @namespace) %>
```
deepsidhu85, line 37: Same comments from `groups/members/index.html.erb` above apply to the project members. We will just need to pass in the `namespace_project_group_links_path` for the url instead

config/locales/en.yml
```diff
@@ -1735,3 +1735,6 @@ en:
       aria_labels:
         expires_at: Member expiration
       inherited_from: Inherited from
+    search_component:
```
deepsidhu85, line 1738: Since we will pass the placeholder into the search component and will differ depending on the page its being used for, this search placeholder can be moved to groups -> members -> index in the locale file

deepsidhu85 (changes requested):
This works great however in it's current form it would only work for the group and project members pages but wouldn't for the group links. I provided some comments below on how we can update it so that it will work between both pages. We should actually be able to use it across the whole application wherever we use a search_form but we can update those pages later

ksierks:
> This works great however in it's current form it would only work for the group and project members pages but wouldn't for the group links. I provided some comments below on how we can update it so that it will work between both pages. We should actually be able to use it across the whole application wherever we use a search_form but we can update those pages later

6e430fb6fa0215f395541c3d8f5f0671e61ee56a should address all of the comments. If this component is to be used anywhere we use a search_form, I am wondering if it lives in the wrong location. Maybe I should it move it from `app/components/members` to `app/components`?

app/views/groups/members/index.html.erb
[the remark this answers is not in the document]
ksierks: I made the change, but I noticed the cursor within the search box moves in-front of the word now.

image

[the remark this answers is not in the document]
joshsadam: I'm going to take a quick stab at that for you @ksierks , I think the issue is that you are returning it in the turbo stream when you don't need to, there is no need to update that.  Just return the updated table in the stream.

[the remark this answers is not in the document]
ksierks: Hopefully, this is what you mean. 1e634e06b6601d3eec5ce03ab290360def295cdd

app/components/search_component.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <%= f.label @search_attribute, @placeholder, class: "sr-only" %>
+  <div class="relative lg:w-72">
+    <div
+      class="
+        absolute inset-y-0 left-0 flex items-center pl-3 pointer-events-none
+      "
+    >
+      <%= viral_icon(name: "magnifying_glass", classes: "h-5 w-5") %>
+    </div>
+    <%= helpers.turbo_frame_tag "member_hidden_values" %>
```
deepsidhu85, line 13: I think we should update this turbo_frame_tag to `ransack_hidden_values` and then update wherever we use `member_hidden_values` since it's just adding ransack hidden sort values

deepsidhu85 (changes requested):
This is working great just have one last comment

config/locales/en.yml
ksierks: 3dbb985ad983a1d56580a16133a6aa71cb67ed2c

app/components/search_component.html.erb
ksierks: 3dbb985ad983a1d56580a16133a6aa71cb67ed2c

deepsidhu85 (approved):

deepsidhu85 (approved):
Looks good to me!

PR #679, opened 2024-07-30
Samples table enhancements

ericenns:

This PR updates the Project and Group Samples table to preserve scroll when sorting, searching, or paging. This also updates the column header names to use `Sample ID` and `Sample Name` for the first two columns instead of `ID` and `Name`. In addition horizontal scrolling of  the table does not result in column headers overlapping with the sticky columns.

image

1. Run `bundle` to make sure your gems match the ones set in this PR
2. Launch the server
3. Login as any user
4. Navigate to a Project Samples page. Verify that clicking any of the sort links or pagination links preserve scroll location on page. And ensure that all samples actions on that page respond correctly and update the table accordingly.
5. Navigate to a Group Samples page. Verify that clicking any of the sort links or pagination links preserve scroll location on page. And ensure that all samples actions on that page respond correctly and update the table accordingly.

## PR acceptance checklist


ChrisHuynh333 (dismissed):
Looks good to me!

ksierks (approved):

app/views/projects/samples/deletions/destroy.turbo_stream.erb
```diff
@@ -10,9 +10,4 @@
       open: false,
```
deepsidhu85, line 10: I think we should only need to append the `flashes` above if the sample was not deleted, since when it is deleted we set the flash and message in the controller

test/system/projects/samples_test.rb
```diff
@@ -1298,19 +1299,6 @@ class SamplesTest < ApplicationSystemTestCase
         assert_selector 'tr:last-child td:nth-child(2)', text: @sample2.name
       end
 
-      click_on 'metadatafield2'
```
deepsidhu85, line 1301: Any specific reason of why this section was removed?

test/system/groups/samples_test.rb
```diff
@@ -122,21 +122,21 @@ def retrieve_puids
       # they are ordered as expected against one another.
       assert_selector 'table tbody tr', count: 20
 
-      click_on I18n.t('groups.samples.table.puid')
+      click_on 'Sample ID'
```
deepsidhu85, line 125: I think these should be internationalized

deepsidhu85 (changes requested):
This is working great and is looking good. Just have a few comments below

app/views/projects/samples/deletions/destroy.turbo_stream.erb
ericenns: Fixed in https://github.com/phac-nml/irida-next/pull/679/commits/bf16319308574e206fac01e46fc034837f02d228

test/system/projects/samples_test.rb
ericenns: Because it was redundant for the test and was causing issues due to scrolling and sticky columns. If you can sort one column you can sort any column.

test/system/groups/samples_test.rb
ericenns: I attempted to do this, but was unable to as the translations are stored with the component and not accessible during testing.

deepsidhu85: I noticed that with another PR as well. We will probably need to figure something out eventually

deepsidhu85 (approved):

Issue #680, opened 2024-07-31
Pipelines Configuration: Executable


ericenns:
## Problem

As an administrator of IRIDA Next I would like to be able to specify which versions of pipelines are not executable.

## Summary

* Update pipeline json config at `config/pipelines/pipelines.json` to support `executable` in `version` config.
  ```json
  [
    {
      "url": "https://github.com/phac-nml/iridanextexample",
      "name": "phac-nml/iridanextexample",
      "description": "IRIDA Next Example Pipeline",
      "versions": [
        {
          "name": "1.0.2",
          "executable": false, # new option
        },
      ],
    },
  ]
  ```
* Update `lib/irida/pipeline.rb` to include `executable` attribute, which is set based off of `version['executable']` if present otherwise it defaults to true
* Update `lib/irida/pipelines.rb` to include an `executable_pipelines` function, which returns all the pipelines which are `executable`
* Update `automatable_pipelines` in `lib/irida/pipelines.rb` to return all the pipelines which are `executable` and `automatable`
* Update pipeline selection dialog to use `executable_pipelines` instead of `available_pipelines`, only use `available_pipelines` for workflow execution show page (all other instance should use `executable_pipelines` execept for cases where `automatable_pipelines` is required

PR #698, opened 2024-08-13
Pipeline config/executable

malchua:

This PR allows an administrator of IRIDA Next to specify which versions of pipelines are not executable.
This PR relates to issue #680.

1. Sign in
2. Select Projects on the sidebar and click on a project that is listed
3. Select Samples in the side bar and select one or two of the samples
4. Click the rocket-ship button to start creating a workflow execution
5. Compare the pipelines that are listed with the pipelines listed in `config/pipelines/pipelines.json`
6. Check to see that only the pipelines that do not have the executable option set to `false` are listed
7. Select Settings in the side bar and select Automated Workflow Executions from the drop-down menu
8. Click on New automated workflow execution button
9. Compare the pipelines that are listed with the pipelines listed in `config/pipelines/pipelines.json`
10. Check to see that only the pipelines with `automatable: true' and not `executable: false` are listed

## PR acceptance checklist

### DIFF
--- a/app/services/workflow_executions/preparation_service.rb
+++ b/app/services/workflow_executions/preparation_service.rb
@@ -16,14 +16,9 @@ module WorkflowExecutions
       @storage_service = ActiveStorage::Blob.service
     end
 
-    def execute # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+    def execute # rubocop:disable Metrics/MethodLength
       # confirm pipeline found
-      if @pipeline.nil?
-        @workflow_execution.state = :error
-        @workflow_execution.cleaned = true
-        @workflow_execution.save
-        return false
-      end
+      return false unless validate_pipeline
 
       @samplesheet_headers = @pipeline.samplesheet_headers
 
@@ -65,6 +60,15 @@ module WorkflowExecutions
     def find_pipeline
       Irida::Pipelines.instance.find_pipeline_by(@workflow_execution.metadata['workflow_name'],
                                                  @workflow_execution.metadata['workflow_version'])
+    end
+
+    def validate_pipeline
+      return true unless @pipeline.nil?
+
+      @workflow_execution.state = :error
+      @workflow_execution.cleaned = true
+      @workflow_execution.save
+      false
     end
 
     def parse_attachments_from_samplesheet(samplesheet)
--- a/app/services/workflow_executions/preparation_service.rb
+++ b/app/services/workflow_executions/preparation_service.rb
@@ -11,17 +11,22 @@ module WorkflowExecutions
     def initialize(workflow_execution, user = nil, params = {})
       super(user, params)
       @workflow_execution = workflow_execution
-      @samplesheet_headers = parse_samplesheet_headers
+      @pipeline = find_pipeline
       @samplesheet_rows = []
       @storage_service = ActiveStorage::Blob.service
-    rescue NoMethodError => e
-      @workflow_execution.errors.add(:base, "#{e.message}: workflow execution not found in executable pipelines")
-      @workflow_execution.state = :error
-      @workflow_execution.cleaned = true
-      @workflow_execution.save
     end
 
-    def execute # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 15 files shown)


JeffreyThiessen (approved):
Look great. Tested and working as intended. :+1:

ChrisHuynh333 (changes requested):
If I have an automated workflow execution set up (when `executable` was `true`, and then its selected pipeline's `executable` is set to `false`, an error occurs if I go to edit the automated workflow execution.

ChrisHuynh333 (changes requested):
Last one from me.
If a workflow execution is started, and the selected pipeline's `executable` is set to `false` while the workflow execution is in state `new/initial`, it will hang and stay in this state because the pipeline is no longer able to be found. In talking to Eric, the state should be updated to `error` in these cases.

You will want to add error handling to the `preparation_service` like we have for other services and add an appropriate message in case in the future, we want to display the error msg back to the user. You will also want to set `.cleaned = true` in order to for the WE to be deletable. We do not need to actually go through the `cleanup_service` as there is nothing that requires cleaning

ericenns:
> I changed it so that even if the automated workflow execution does not have an executable pipeline, they can still see it and can edit it. Should I instead change that to so they won't be able to see it if it's not executable?

I was thinking that we need some way to inform the user that the `automated_workflow_execution` will not work since it  is either no longer executable or automatable. For now I think your solution is sufficient, but I think it should be addressed in a follow up PR.

app/services/workflow_executions/preparation_service.rb
```diff
@@ -11,12 +11,22 @@ class PreparationService < BaseService
...
 
-    def execute # rubocop:disable Metrics/MethodLength
+    def execute # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+      # confirm pipeline found
+      if @pipeline.nil?
+        @workflow_execution.state = :error
+        @workflow_execution.cleaned = true
+        @workflow_execution.save
+        return false
+      end
```
ChrisHuynh333, line 26: Due to now needing an additional rubocop disable, I think we can just move checking `@pipeline.nil?`  into its own method.

Can you replace this with `return false unless validate_pipeline`

And then under `private`:
```
def validate_pipeline
      return true unless @pipeline.nil?

      @workflow_execution.state = :error
      @workflow_execution.cleaned = true
      @workflow_execution.save
      false
    end
```

ChrisHuynh333 (changes requested):
Works great! Just one last refactor from me

ChrisHuynh333 (approved):
Looks good!

config/pipelines/pipelines.json
```diff
@@ -4,6 +4,10 @@
     "name": "phac-nml/iridanextexample",
     "description": "IRIDA Next Example Pipeline",
     "versions": [
+      {
```
deepsidhu85, line 7: Can we remove the `"executable": false` from this pipeline and change one of the earlier versions to include it instead? For testing purposes we probably want the newest version to be executable

app/services/workflow_executions/preparation_service.rb
```diff
@@ -52,10 +57,18 @@ def execute # rubocop:disable Metrics/MethodLength
...
-    def parse_samplesheet_headers
-      workflow = Irida::Pipelines.instance.find_pipeline_by(@workflow_execution.metadata['workflow_name'],
-                                                            @workflow_execution.metadata['workflow_version'])
-      workflow.samplesheet_headers
+    def find_pipeline
+      Irida::Pipelines.instance.find_pipeline_by(@workflow_execution.metadata['workflow_name'],
+                                                 @workflow_execution.metadata['workflow_version'])
+    end
+
+    def validate_pipeline
```
deepsidhu85, line 65: We probably want to update the job to also set the state to error as the pipeline can be marked as non-executable after submission and before the preparation job actually runs

deepsidhu85 (changes requested):
This is looking good just a few comments below

deepsidhu85 (approved):
Looks good to me!

PR #699, opened 2024-08-13
Enhancement: Attachment Checksum Validation

ChrisHuynh333:
This PR changes the Attachment checksum validation. Previously, samples could not have two attachments with the same checksum attached. With the new validation, samples can have attachments with the same checksum attached, as long as they have different filenames

Attachments with same checksum but different filenames:
Old Validation:
[Screencast from 2024-08-13 03:53:17 PM.webm](https://github.com/user-attachments/assets/076e033a-00e3-47c2-8cac-0eb23e303aac)

New Validation:
[Screencast from 2024-08-13 03:51:56 PM.webm](https://github.com/user-attachments/assets/9377163f-3a41-4154-a013-660e3e19c052)

1. Create two files that are a copy of one another (same checksum) except with different filenames
2. Navigate to any sample files page
3. Upload the above files. They should upload normally.
4. With the above files still attached to the sample, try attaching the files again. They should all fail uploading as the checksum and filenames match already existing attachments.

## PR acceptance checklist

### DIFF
--- a/app/validators/attachment_checksum_validator.rb
+++ b/app/validators/attachment_checksum_validator.rb
@@ -2,7 +2,7 @@
 
 # Validator for Attachment checksum
 class AttachmentChecksumValidator < ActiveModel::Validator
-  def validate(record) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+  def validate(record)
     return if record.file.checksum.blank?
 
     klass = record.class
--- a/app/validators/attachment_checksum_validator.rb
+++ b/app/validators/attachment_checksum_validator.rb
@@ -8,21 +8,13 @@ class AttachmentChecksumValidator < ActiveModel::Validator
     klass = record.class
 
     # add error if another Attachment associated with the Attachable exists with the same checksum
-
-    # checks if file has same checksum and name (no error if same contents but different name)
+    # no error occurs if the files have same name but different checksum or same checksum but different filenames
     if klass.joins(:file_blob)
             .where.not(id: record.id)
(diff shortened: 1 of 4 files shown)


app/validators/attachment_checksum_validator.rb
```diff
@@ -2,18 +2,28 @@
...
     # add error if another Attachment associated with the Attachable exists with the same checksum
+
+    # checks if file has same checksum and name (no error if same contents but different name)
     if klass.joins(:file_blob)
             .where.not(id: record.id)
             .exists?(attachable_id: record.attachable_id,
                      attachable_type: record.attachable_type,
                      deleted_at: nil,
-                     file_blob: { checksum: record.file.checksum })
+                     file_blob: { checksum: record.file.checksum, filename: record.file.filename.to_s }) ||
```
ericenns, line 18: **nitpick (blocking):** No need to have a `||` since we only want to add the error when an existing attachment exists with the same filename and checksum. If a file already exists with the same name but different checksum we can allow attachment as that is existing behaviour.

ericenns (changes requested):
Just a minor comment, validation was failing when it should have passed

app/validators/attachment_checksum_validator.rb
ChrisHuynh333: Fixed in 163e8b5fe23ffeb5377469c8f6b1f57dddf182a8

ericenns (approved):
Looks good to me!

ericenns (approved):
Looks good to me!

PR #715, opened 2024-08-21
Add viral search component and controller

joshsadam:
This pull request adds a new viral search component and controller to the application. The search component is a UI element that allows users to search for items and displays the search results in a dropdown. The controller handles the search functionality and navigation within the dropdown. This feature enhances the user experience by providing a convenient and efficient way to search for items.


app/views/groups/group_links/_invite_group_modal.html.erb
```diff
@@ -15,16 +15,14 @@
...
-                        new_group_link.group_id
-                      ),
-                      {
-                        prompt: I18n.t("groups.group_links.new.placeholder.select_group")
-                      },
-                      required: true,
-                      "aria-label": t(:"groups.group_links.index.aria_labels.group_id") %>
+          <%= viral_select2(form:, name: :group_id) do |select| %>
+            <% @namespace_linkable_groups.each do |group| %>
+              <%= select.with_option(primary: group.name, secondary: group.id, value: group.id) %>
```
ChrisHuynh333, line 20: question: Would it make more sense to display the PUIDs rather than our uuids? I just assumed users generally referenced our PUIDs when wanting to use ID as an identifier

ChrisHuynh333 (commented):
This is great! One comment/question from me so far, will continue to look in more detail.

app/views/groups/group_links/_invite_group_modal.html.erb
ericenns: Yeah we want to show PUID not the ID, which is a UUID that the users never see in the application.

joshsadam: Fixed in 76f7580

ericenns (changes requested):
This looks pretty good, just have a few comments:

* Add support for a `placeholoder`, which would display when nothing is select
* Make it look more like the existing select dropdown
  * Background color of input
  * Add `caret` to indicate that the user can click it select a value.

Otherwise the dropdown portion of it looks good.

app/assets/stylesheets/application.tailwind.css
```diff
@@ -84,8 +84,23 @@ body {
...
+    background-repeat: no-repeat;
+    background-size: 0.75em 0.75em;
+    padding-right: 2.5rem;
+    -webkit-print-color-adjust: exact;
+    print-color-adjust: exact;
+  }
+
+  .form-field select,
+  .form-field select option {
+    @apply border-slate-300  text-slate-900 sm:text-sm rounded-md focus:ring-primary-500 focus:border-primary-500 block w-full p-2.5 dark:bg-slate-800 dark:border-slate-600 dark:placeholder-slate-400 dark:text-slate-50 dark:text-opacity-60 dark:focus:ring-primary-500 dark:focus:border-primary-500;
```
ericenns, line 99: **nitpick (blocking):** Please replace `dark:text-slate-50 dark:text-opacity-60` with `dark:text-white`

ericenns (changes requested):
Just a minor issue with select dropdown in dark mode, see comments.

app/assets/stylesheets/application.tailwind.css
joshsadam: Fixed in 69faf8dc793d99ab18897c3f68c5ee1d2d4be6dd

ericenns (approved):
Looks good to me!

Issue #736, opened 2024-09-05
Samples Table: Preserve page size when searching


ericenns:
## Problem
When a user changes the number of samples to display per page and then initiates a search via search box or filter by sample puids/names, the page size resets to the default of 20 items.

## Acceptance criteria
* `limit` param is included when POSTing to samples search endpoint
* `limit` param if present is included in `redirect_to` within `search` action of samples controllers
* `page` param should not be preserved when searching

Issue #75, opened 2023-04-20
Group Settings Update


ericenns:
### Summary
Update the Group Settings page to separate updates to `path` and `parent` namespace into the `Advanced` section.

### Implementation Guide
* Remove `path` and `parent namespace` updates from `Name and description` section.
* Add `Change group URL` under `Advanced` section and only allow changing the `path`
  * Make POST request to group_path (normal update endpoint)
* Add `Transfer group` under `Advanced` section and allow user to select any group, subgroup, or personal namespace they have access to
  * Add `groups/transfer_service.rb` (can use `projects/transfer_service.rb` as an example).
  * Add `transfer` method to `groups_controller.rb` and call the `groups/transfer_service.rb`
  * Add `transfer` endpoint to `routes/groups.rb` using `PUT` method

ksierks:
@ericenns @josh I have a small question. It looks like at one point the accordion was used to organise the group advanced settings. I also see that the card is being used in the [WIP: Project settings update](https://github.com/phac-nml/irida-next/pull/92/files#top) #92 PR. Do I continue to use the accordion or use the card for consistency? Thanks.

ksierks:
Done.

PR #748, opened 2024-09-10
Samples Table: Fixes preserving page size & number when toggling metadata

### LINKED ISSUE #737: Samples table: Preserve page size and page number when toggling metadata
## Problem
When a user changes the number of samples to display per page and then initiates a search via search box or filter by sample puids/names, the page size resets to the default of 20 items.

## Acceptance criteria
* `limit` and `page` params are included when POSTing to samples search endpoint
* `limit` and `page` params if present are included in `redirect_to` within `search` action of samples controllers

ksierks:
Persisting the page size & number during metadata toggling.
Fixes #737.

image
image

1. Navigate to a group or project samples page.
2. Select a page size other than the default.
3. Toggle the metadata button.
4. Verify the page size & number remain unchanged.

## PR acceptance checklist


config/initializers/pagy.rb
```diff
@@ -145,8 +145,8 @@
 
 # Overflow extra: Allow for easy handling of overflowing pages
 # See https://ddnexus.github.io/pagy/docs/extras/overflow
-# require 'pagy/extras/overflow'
-# Pagy::DEFAULT[:overflow] = :empty_page    # default  (other options: :last_page and :exception)
+require 'pagy/extras/overflow'
```
joshsadam, line 148: **question:** Just curious on why you had to uncomment this, I could not see it being used.

ericenns (changes requested):
This is not working as expected. Now `page` is preserved when searching which breaks last acceptance criteria for #736 and results in a weird url in the address bar.

image

I think that the metadata toggling will need to be moved into a separate form so that `page` param is not including in the search form.

ksierks:
> > > This is not working as expected.
> > 
> > 
> > What is not working as expected? From the end user point of view, I think it's doing what it's suppose to be doing. Your screenshot shows page 17 in the URL, because you were likely on page 17 before filtering. Aren't the results within the table correct? Once you clear the filter, the user sees page 1 again. I'm confused.
> 
> > > This is not working as expected.
> > 
> > 
> > What is not working as expected? From the end user point of view, I think it's doing what it's suppose to be doing. Your screenshot shows page 17 in the URL, because you were likely on page 17 before filtering. Aren't the results within the table correct? Once you clear the filter, the user sees page 1 again. I'm confused.
> 
> In issue #736, it specifies that `page` param should not be preserved when searching/filtering (e.g. search box or filter by samples names or Id.), in your previous pr the `page` param was not sent in the search form, but in this PR you send that param. By setting up the overflow, now the address bar `page` param does not match the currently displayed `page`.
> 
> The intention with not sending the `page` param while searching was so that the user would always be directed to the first page of results. With this current setup the `page` param is sent in the search form and the user is sent to the `last` page of results if the current page they are on overflows the pages after searching.

I think overflow should go to the last page. The reason the error is happening is because the user is currently on a page other than the first page. 

The user acceptance criteria is designed to come from the user's perspective. I think you meant to label it `Implementation Guide`, like in https://github.com/phac-nml/irida-next/issues/75. I tried re-reading the user acceptance criteria from the issue and there was no mention of moving the metadata toggle into it's own form. Only to add the page param to the current search form. 
image.

I will try to re-implement it with this new information and move the metadata toggling into it's own form. Thanks.

config/initializers/pagy.rb
ericenns, line 1: **nitpick (blocking):** Could you revert this file to what it looks like in main branch

ericenns (changes requested):
This is now working as expected. Just one minor comment from me, otherwise this looks great

config/initializers/pagy.rb
ksierks: 31e8a941a9b797f0d4cad557051e5dda81f9c40a

ericenns (approved):
Looks good to me!

joshsadam (approved):
This works.  I think that is a pretty good solution.

PR #760, opened 2024-09-16
Public activity: Groups

deepsidhu85:

This PR adds in an `Activity` page to a **group** which will list the following activities:

| Type  | Action |
| ------------- | ------------- |
| Group  | Created, Updated, Transferred (in as subgroup, out), Subgroup Removed  |
| Member  | Added, Updated, Removed, Leaves Project  |
| Shared Group  | Shared, Unshared  |
| Project  | Created, Removed, Transferred (in and out)  |

image

image

1. Create a `group` and then edit it's attributes (eg. `description`)
2. View the `activity` page for the `group` and there should be 2 entries, one for creating of the group and another for group attributes being updated
3. From the `members` page, Add a couple of members, one as an owner, and another with any access level, and then update the second user (the one with `any` access level`) to another access level
4. View the `activity` page for the project and you should see 3 entries, two for adding new users, and one for updating a members attributes
5. Now remove the user above, then also leave the group with the current user. Keep note of the remaining user who has an owner access level and then sign in with them
6. View the group's `activity` page and you should see entries for a user being removed, and a user removed themselves from the group
7. Share the `group` via the members page to any other group, then `unlink` the shared `group`
8. View the `activity` page and you should see two entries, one for sharing and one for unsharing the group
9. Create a `project` in the `group` and view the `activity` page. There should be an entry for the project creation in the `group` with a clickable link to the `project` in the activity entry
10. Now remove this `project` and go back to the `activity` page. Verify that there is an entry for removal of the `project`. Also verify that the `PUID` is not clickable in the removal or the creation entries for the `project`
11. Create another `group` and within it create a `project`. Transfer this `project` to the `group` in step 1
12. View the `activity` for the `group` and there should be an entry for the `project` being transferred into the `group`
13. Now transfer the above `project` to another `group` and view the `activity` page for the `group` from step 1. There should be an additional entry for the `project` being transferred to another namespace
14. Create 2 new groups, and 1 new subgroup under the second group that was created
15. Transfer the second group into the namespace for the first group from step 14
16. View the activity page for the first group in step 14 and there should be an entry about the second group being transferred in, and view the activity page for the second group which should also  have an entry for the group being transferred under another namespace
17. Delete the subgroup from step 14 and then view the activity for the second group from step 14
18. View the activity page and verify that there is an entry for the subgroup that was removed

## PR acceptance checklist


app/services/groups/transfer_service.rb
```diff
@@ -47,5 +49,69 @@ def validate(new_namespace)
       raise TransferError,
             I18n.t('services.groups.transfer.same_group_and_namespace')
     end
+
+    def create_activities(old_namespace, new_namespace) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
```
ChrisHuynh333, line 53: When transferring a group, the transferred group has two entries for the same action:
image

ChrisHuynh333 (changes requested):
Looks great! Couple things from me just playing around to start:

First, thoughts on making the PUIDs that are links stand out differently (maybe underlined or slightly different color, not sure), I wouldn't have thought they were clickable, and then after knowing they're clickable, no way to distinguish between clickable and non-clickable PUIDs

app/services/groups/transfer_service.rb
deepsidhu85: Fixed in 093ab7b91119f065634ca4b057aba6dbd631abef

deepsidhu85:
> Looks great! Couple things from me just playing around to start:
> 
> First, thoughts on making the PUIDs that are links stand out differently (maybe underlined or slightly different color, not sure), I wouldn't have thought they were clickable, and then after knowing they're clickable, no way to distinguish between clickable and non-clickable PUIDs

Since we are doing the same exact with our other links I think we will leave this as is for now.

app/models/concerns/track_activity.rb
```diff
@@ -55,6 +57,31 @@ def project_activity(activity)
...
+      project_puid: activity.parameters[:project_puid],
+      name: activity_trackable.name,
+      type: 'Namespace',
+      action: activity.parameters.key?(:action) ? activity.parameters[:action] : 'default'
+    }
+
+    return base_params if activity.parameters[:action].blank?
+
+    transfer_activity_parameters(base_params, activity)
+  end
```
joshsadam, line 83: **suggestion (non-blocking):** There is a lot going on here, and we always keep adding the `rubocop:disabled` comment.  Can you break this up a bit like:

```ruby
def group_activity(activity)
  activity_trackable = activity_trackable(activity, Group)

  base_params = build_base_params(activity, activity_trackable)
  additional_params = build_additional_params(activity)

  params = base_params.merge(additional_params)
  
  return params if activity.parameters[:action].blank?

  transfer_activity_parameters(params, activity)
end

private

def build_base_params(activity, activity_trackable)
  {
    created_at: format_created_at(activity.created_at),
    key: "activity.#{activity.key}_html",
    user: activity_creator(activity),
    group: activity_trackable,
    name: activity_trackable.name,
    type: 'Namespace',
    action: activity.parameters.fetch(:action, 'default')
  }
end

def build_additional_params(activity)
  {
    created_group: get_object_by_id(activity.parameters[:created_group_id], Group),
    transferred_group: get_object_by_id(activity.parameters[:transferred_group_id], Group),
    removed_group_puid: activity.parameters[:removed_group_puid],
    project: get_object_by_id(activity.parameters[:project_id], Project),
    project_puid: activity.parameters[:project_puid]
  }
end

def format_created_at(created_at)
  created_at.strftime(I18n.t('time.formats.abbreviated'))
end
```

ericenns (dismissed):
This is looking like a good first pass at this, it adds the bare minimum activities and bare minimum rendering.

ChrisHuynh333 (approved):
Looks good!

app/components/activities/group_activity_component.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Activities
+  # Component for rendering an activity of type Namespace for Projects
+  class GroupActivityComponent < Component
+    def initialize(activity: nil, **system_arguments)
+      @activity = activity
+
+      @system_arguments = system_arguments
```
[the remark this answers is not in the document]
deepsidhu85: Removed in 40309cfa83dfb1820548f4e73e858e3d2a05d3dc

app/components/activities/member_activity_component.rb
```diff
@@ -8,5 +8,16 @@ def initialize(activity: nil, **system_arguments)
 
       @system_arguments = system_arguments
```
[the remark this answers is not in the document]
deepsidhu85: Removed in 40309cfa83dfb1820548f4e73e858e3d2a05d3dc

app/models/concerns/track_activity.rb
deepsidhu85: Refactored in 40309cfa83dfb1820548f4e73e858e3d2a05d3dc to remove most of the rubocop disables

joshsadam (approved):
Looks good to me.  Thanks for making those changes.

PR #776, opened 2024-09-20
Namespace attachments: Index attachment query change

ChrisHuynh333:
This PR alters the `namespace.attachments` query in `index` to exclude paired-end (PE) reverse attachments. We are doing this because we render the PE attachments together within a single table row, but ransack will count both attachments, so 5 PE attachments would take up 10 'slots' within the attachments table, which is not intuitive with the limit selection.

In addition, when using ransack filter, the list of attachments will render each attachment individually (ie: PE attachments will be separated).

Currently on main (note limit: 10, but only 6 entries on the table):
image

Now:
image

Filtering PE:
image

In both project and group attachments:
1. Upload multiple paired-end files
2. Verify that each paired-end entry counts as a single attachment.
3. Test the new filter by filtering by something (`fastq.gz`) that will render both of the PE attachments. Verify they render individually.

## PR acceptance checklist

### DIFF
--- a/app/components/attachments/table_component.html.erb
+++ b/app/components/attachments/table_component.html.erb
@@ -116,10 +116,16 @@
                 <% if column == :id %>
                   <%= attachment.puid %>
                 <% elsif column == :filename %>
-                  <% if attachment.associated_attachment %>
+                  <% if @render_individual_attachments %>
                     <div>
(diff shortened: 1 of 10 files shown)


ericenns (approved):
Looks good to me!

PR #860, opened 2024-12-03
Data Table Bot Accounts: Remove PAT panel on destroy

ChrisHuynh333:
This PR is a followup to [#849](https://github.com/phac-nml/irida-next/pull/849), and adds logic to remove the new PAT frame upon bot deletion or revoked PAT. This is because we wouldn't want the new PAT panel to persist upon the potential removal of the PAT itself or its associated bot. Due to the changes in response logic in #849, the PAT frame now uses `data-turbo-permanent` attribute to persist through response refreshes, and requires explicit logic to remove/hide it.

Before:
[Screencast from 2024-12-19 01:17:14 PM.webm](https://github.com/user-attachments/assets/2c2be1e5-9343-4834-bfab-5676c7b36e05)

Now:
[Screencast from 2024-12-19 01:17:53 PM.webm](https://github.com/user-attachments/assets/76816d1b-f443-4c6d-9107-314c34d4bdfa)

Repeat for both project and group bots:
1. Navigate to the bot accounts page
2. Create a new bot or generate a new token for an existing bot. Afterwards, the new PAT frame should render.
3. Delete a bot. The new PAT frame should no longer appear.
4. Repeat step 2, and afterwards, revoke a PAT from a bot. Once again, the new PAT frame should no longer appear.

## PR acceptance checklist


joshsadam (changes requested):
I really like this and your code looks great.  Just a couple comments.

app/controllers/concerns/bot_actions.rb
```diff
@@ -53,6 +53,15 @@ def create # rubocop:disable Metrics/MethodLength
     end
   end
 
+  def new_destroy
```
deepsidhu85, line 56: I feel like this should be named something like `destroy_confirmation` as this brings up the confirmation modal when we click on the `Delete` for a bot account. Thoughts @ChrisHuynh333  @joshsadam

app/controllers/concerns/bot_personal_access_token_actions.rb
```diff
@@ -54,9 +54,18 @@ def create # rubocop:disable Metrics/MethodLength
     end
   end
 
-  def revoke
+  def new_revoke
```
deepsidhu85, line 57: Same type of comment as above. This should be named something like `revoke_confirmation` as this brings up the confirmation modal when we click on `Revoke` for a bot account token.

config/routes/group.rb
```diff
@@ -18,8 +18,10 @@
     resources :members, only: %i[create destroy index new update]
 
     resources :bots, only: %i[create destroy index new] do
+      get :new_destroy
```
deepsidhu85, line 21: As per my comments above this and the new_revoke routes would need to be changed

config/routes/project.rb
```diff
@@ -23,9 +23,11 @@
       end
 
       resources :bots, only: %i[create destroy index new] do
+        get :new_destroy
```
deepsidhu85, line 26: As per my comments above this and the new_revoke routes would need to be changed

deepsidhu85 (changes requested):
This is working great. Just left a few comments about possibly changing the route/method names

app/controllers/concerns/bot_actions.rb
joshsadam: Agreed, that name definitely works better

app/javascript/controllers/token_controller.js
```diff
@@ -35,4 +35,11 @@ export default class extends Controller {
     }
     this.visible = !this.visible;
   }
+
+  removeTokenPanel() {
+    let panel = document.getElementById("access-token-section");
```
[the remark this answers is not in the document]
ChrisHuynh333: This DOM element is outside of the current controller:
image

I could move the controller call further 'out' but figured this would be a simpler solution without messing too much with the token controller. Thoughts?

test/controllers/concerns/bot_actions_concern_test.rb
```diff
@@ -152,4 +152,28 @@ class BotActionsConcernTest < ActionDispatch::IntegrationTest
...
+    namespace = groups(:group_one)
+    namespace_bot = namespace_bots(:group1_bot0)
+
+    get group_bot_new_destroy_path(namespace, bot_id: namespace_bot.id)
+
+    assert_response :success
+  end
+
+  test 'new_destroy in project' do
+    sign_in users(:john_doe)
```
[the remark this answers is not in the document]
ChrisHuynh333: I've added unauthorized test cases (thanks for catching, I forgot this part) 1fb24d4a6a573c2fc31244cd61c54bcb6b806416

That will cover 'failure' as it's just launching the modal.

app/controllers/concerns/bot_actions.rb
ChrisHuynh333: Thanks, my head got stuck in how rails has `new` for create dialogs, but agreed I like this better as well.
Changed in 44216f0f3d059fa28a6e3abbd5f78ff3419ceaf8

config/routes/group.rb
ChrisHuynh333: Changed in 44216f0f3d059fa28a6e3abbd5f78ff3419ceaf8

config/routes/project.rb
ChrisHuynh333: Changed in 44216f0f3d059fa28a6e3abbd5f78ff3419ceaf8

app/controllers/concerns/bot_personal_access_token_actions.rb
ChrisHuynh333: Changed in 44216f0f3d059fa28a6e3abbd5f78ff3419ceaf8

deepsidhu85 (approved):

joshsadam (approved):
Looks good to me

Issue #884, opened 2025-01-08
Share Workflow Execution with Namespace


ericenns:
## Summary
As a User in IRIDA Next, I want the results of a pipeline I ran to be available to another user in my group, so that they can access my results and proceed with the workflow.

## Acceptance Criteria
* While creating a Workflow Execution the user should see a checkbox at the bottom which says something like `Share with Project` or `Share with Group`, depending on the namespace type that the user selected samples from
  * If the checkbox is selected this should set a `shared_with_namespace` boolean to true in the `workflow_execution` model
* While viewing Workflow Executions from a Project, workflow executions with `shared_with_namespace` set to true where `namespace_id` matches the project namespaces id should also appear in the table
  * These workflow executions that are shared should not be deletable, but could be exportable
* A user with `view_workflow_executions?` permission in the namespace that the workflow execution was shared with should be able to view the workflow execution, but they should not be able to modify it in anyway.

PR #895, opened 2025-01-13
Workflow Executions: Share results

malchua:

This PR relates to STRY0016470, where we want to be able to share our workflow execution results to the project/group so that members are able to see the results. This PR covers issue #884

image
image

1. Login as a user that is a member of a group.
2. Select a project within that group.
3. In the samples page for that project, create a workflow execution and check the box asking if you want to share the results with the project members.
4. After creating the workflow execution, click on the Workflow Executions section on the sidebar of that project.
5. Confirm that the workflow execution is present.
6. Log out, then login as a different member that has access to the same group (the access level must be at least an Analyst)
7. Go to the Workflow Executions page for the same project.
8. Confirm that this member is also able to see the same workflow execution.
9. Confirm that this member is not allowed to Cancel, Delete, or Edit this workflow execution.

## PR acceptance checklist


test/system/workflow_executions_test.rb
```diff
@@ -3,7 +3,7 @@
 require 'application_system_test_case'
```
ChrisHuynh333, line 3: Can you add a UI test verifying the expected absence of the shared workflow row actions

app/policies/workflow_execution_policy.rb
```diff
@@ -109,4 +109,11 @@ def update?
 
     relation.where(submitter_id: user.id)
   end
+
+  scope_for :relation, :automated_and_shared do |relation, options|
+    project = options[:project]
+
+    relation.where(submitter: project.namespace.automation_bot)
+            .or(relation.where(namespace_id: project.namespace.id, shared_with_namespace: true))
+  end
```
ChrisHuynh333, line 118: Do we need to share automated workflow executions? They are naturally shared to those with proper permissions

ChrisHuynh333 (commented):
Some comments below to start and also some clarification questions (maybe geared more towards @ericenns :
1. I would think we need a Workflow Execution listing for groups now? I don't see anywhere to view a workflow that's been shared with the group
2. Should the user that created a shared workflow still be able to cancel/delete/modify the workflow from their own personal workflow listing?
3. Listing the user that created the shared workflow on the workflow summary tab?


Also, could the description header for the project workflow executions page be updated as it no longer pertains to only automated workflows:
image

app/components/nextflow_component.rb
```diff
@@ -7,14 +7,16 @@ class NextflowComponent < Component
   attr_reader :schema, :url, :workflow, :metadata_fields, :samples, :namespace_id, :instance
 
   # rubocop:disable Metrics/ParameterLists
-  def initialize(url:, samples:, workflow:, fields:, namespace_id:, allowed_to_update_samples: true, instance: nil)
+  def initialize(url:, samples:, workflow:, fields:, namespace_id:, allowed_to_update_samples: true, instance: nil,
+                 namespace_type: nil)
```
ericenns, line 11: **nitpick (blocking):** Instead of passing in `namespace_type` can you instead get the type from the passed in `namespace_id`.

app/components/nextflow_component.html.erb
```diff
@@ -110,6 +110,21 @@
...
+      <div class="flex items-center h-5 mb-4">
+        <%= workflow.check_box :shared_with_namespace,
+                           {
+                             checked:
+                               instance.present? && instance["shared_with_namespace"],
+                             class:
+                               "w-4 h-4 mr-2.5 text-primary-600 bg-slate-100 border-slate-300 rounded focus:ring-primary-500 dark:focus:ring-primary-600 dark:ring-offset-slate-800 focus:ring-2 dark:bg-slate-700 dark:border-slate-600",
+                           } %>
+        <%= workflow.label :shared_with_namespace,
+                       t(
```
ericenns, line 122: **nitpick (blocking):** Instead of passing in the `namespace_type` to the translation can you instead nest the translations so that there is a specific one for group and a specific one for project.

e.g.
```yaml
en:
  components:
    nextflow:
      shared_with:
        project: 'Share sreults with Group members`
        group: `Share results with Project members`
```

and access it like so:

```ruby
t(:"components.nextflow.shared_with.#{@namespace_type.lower}")
```

app/views/projects/automated_workflow_executions/_edit_dialog.html.erb
```diff
@@ -1,15 +1,28 @@
...
+          @project,
+          @automated_workflow_execution,
+        ),
       samples: Sample.none,
       allowed_to_update_samples: true,
       namespace_id: @project.namespace.id,
       fields: nil,
-      instance: @automated_workflow_execution
+      instance: @automated_workflow_execution,
+      namespace_type: @project.namespace.type,
```
ericenns, line 25: **nitpick (blocking):** based off of comment above, this does not need to be passed in.

app/views/projects/automated_workflow_executions/_new_automated_workflow_execution_modal.html.erb
```diff
@@ -7,7 +7,8 @@
       samples: Sample.none,
       allowed_to_update_samples: true,
       namespace_id: @project.namespace.id,
-      fields: nil
+      fields: nil,
+      namespace_type: @project.namespace.type,
```
ericenns, line 11: **nitpick (blocking):** based off of comment above, this does not need to be passed in.

app/views/projects/workflow_executions/show.html.erb
```diff
@@ -41,7 +41,7 @@
             "button button--size-default button--state-default pointer-events-none cursor-not-allowed bg-slate-100 text-slate-600 mr-2 dark:bg-slate-600 dark:text-slate-300" %>
         <% end %>
       <% end %>
-      <% if @workflow_execution.cancellable? && allowed_to?(:cancel?, @workflow_execution) %>
+      <% if @workflow_execution.cancellable? && allowed_to?(:cancel?, @workflow_execution) && !@workflow_execution.shared_with_namespace %>
```
ericenns, line 44: **nitpick (blocking):** Why not just update the policy so that it would return false for allowed to instead of adding more logic to the view.

app/views/projects/workflow_executions/show.html.erb
```diff
@@ -57,7 +57,7 @@
           class: "button button--state-default button--size-default mr-1",
         ) %>
       <% end %>
-      <% if allowed_to?(:update?, @workflow_execution) %>
+      <% if allowed_to?(:update?, @workflow_execution) && !@workflow_execution.shared_with_namespace %>
```
ericenns, line 60: **nitpick (blocking):** Same as comment above, about just making sure the policy method returns true when it is shared_with_namespace.

app/views/projects/workflow_executions/show.html.erb
```diff
@@ -71,7 +71,7 @@
           class: "button button--state-default button--size-default mr-1",
         ) %>
       <% end %>
-      <% if @workflow_execution.deletable? && allowed_to?(:destroy?, @workflow_execution) %>
+      <% if @workflow_execution.deletable? && allowed_to?(:destroy?, @workflow_execution) && !@workflow_execution.shared_with_namespace %>
```
ericenns, line 74: **nitpick (blocking):** Same as comment above, about just making sure the policy method returns true when it is shared_with_namespace.

app/views/workflow_executions/submissions/create.turbo_stream.erb
```diff
@@ -10,6 +10,7 @@
         allowed_to_update_samples: @allowed_to_update_samples,
         namespace_id: @namespace_id,
         fields: @fields,
+        namespace_type: @namespace.nil? ? nil : @namespace.type,
```
ericenns, line 13: **nitpick (blocking):** Please remove this as comment way above makes this unnecessary.

ericenns (changes requested):
Left a few comments on implementation. I have not tried creating an automated workflow execution but we should ensure that the `shared_with_namespace` checkbox is not shown when creating an automated workflow execution as that would be confusing to the users.

app/components/nextflow_component.html.erb
```diff
@@ -110,6 +110,22 @@
                        t(:"components.nextflow.email_notification"),
                        class: "mr-2 text-sm font-medium text-slate-900 dark:text-white" %>
       </div>
+      <div class="flex items-center h-5 mb-4">
+        <% if !@workflow.automatable && @namespace_type %>
```
ChrisHuynh333, line 114: Can you flip these (and their closing tags) so that you don't render the empty div

test/system/projects/workflow_executions_test.rb
```diff
@@ -44,49 +46,78 @@ class WorkflowExecutionsTest < ApplicationSystemTestCase
...
+        assert_no_link 'Cancel'
+        assert_no_link 'Destroy'
+      end
+
+      tr = find('a', text: workflow_execution2.id).ancestor('tr')
+      within tr do
+        assert_no_link 'Cancel'
+        assert_no_link 'Destroy'
+      end
+    end
```
ChrisHuynh333, line 111: Most of our table components leverage the item's `ids` to identify their respective rows so I think a cleaner way to write this would be:
```
assert_selector "tr[id='#{workflow_execution1.id}']" // make sure the row is rendered to avoid flakes
within("tr[id='#{workflow_execution1.id}'] td:last-child") do // specify the cell
  assertions
end
```
Also use translations as much as possible

test/system/workflow_executions_test.rb
```diff
@@ -109,6 +111,28 @@ class WorkflowExecutionsTest < ApplicationSystemTestCase
...
+    visit workflow_executions_path
+
+    assert_selector 'h1', text: I18n.t(:'workflow_executions.index.title')
+
+    tr = find('a', text: workflow_execution.id).ancestor('tr')
+
+    within tr do
+      assert_link 'Cancel'
+    end
+  end
```
ChrisHuynh333, line 126: Same comment as above

test/system/workflow_executions_test.rb
```diff
@@ -109,6 +111,28 @@ class WorkflowExecutionsTest < ApplicationSystemTestCase
...
+
+  test 'should not include a shared workflow in the workflow executions when the submitter is not the current user' do
+    workflow_execution = workflow_executions(:workflow_execution_shared2)
+
+    visit workflow_executions_path
+
+    assert_no_selector 'a', text: workflow_execution.id
+  end
```
ChrisHuynh333, line 134: Can you add `assert_selector '#workflow-executions-table'` to verify the table has rendered prior to asserting that the workflow does not exist

app/components/workflow_executions/table_component.html.erb
```diff
@@ -140,7 +140,7 @@
                 <% end %>
               <% end %>
             <% end %>
-            <% if @renders_row_actions %>
+            <% if @renders_row_actions && (!workflow_execution.shared_with_namespace || @namespace.nil?) %>
```
ChrisHuynh333, line 143: Can you move the new conditionals under the `render_cell` call so that we still render the cell but not the actions. This way it prevents weird overlapping like so:
image

ChrisHuynh333 (changes requested):
One more comment while playing around

app/views/projects/workflow_executions/show.html.erb
ericenns: I guess if you left out the `&&` part then it would not show the actions for a workflow execution shared to a project when logged in as Jane. But if you logged in as John (where John launched the workflow execution and shared it to the project), would it show the actions for John? Whereas with the `&&` part it is not showing for `John`? If so I would be ok for it to show the actions to John in the project view for code simplicity reasons.

app/components/workflow_executions/table_component.html.erb
ChrisHuynh333: You want to leave `<% if @renders_row_actions %>` above the render cell call because if there's no row actions, we won't render the column at all (table header follows the same logic)

app/components/nextflow_component.html.erb
ChrisHuynh333: Building on this, could you fix the path within the export preview to navigate to the workflow within the project if it's shared:
image

Currently it will navigate to the workflow as if it's owned by the user (which would be inaccessible), but if shared, we want to access it through the project (hopefully that makes sense).
Please also add associated UI tests for creating export from a shared workflow and clicking a shared link navigates to the expected page. There are similar tests in `test/system/data_exports_test` that you can draw from.

ChrisHuynh333 (approved):
Looks good!

ericenns (approved):
Looks good to me!

PR #917, opened 2025-01-27
Advanced Search: Adding new 'exists' and 'not_exists' operators

ksierks:
The PR adds the ability to search for metadata fields that exists and not exists.

Exists:
image
image

Not exists:
image
image

1. Navigate to a group or project samples page.
1. Click the Advanced Search button. 
1. Verify the Advanced Search dialog opened.
1. Select a field and `exists` or `not_exists` operator to filter by.
1. Click the Apply Filter button.
1. Verify the table displays the correct search results.

## PR acceptance checklist

### DIFF
--- a/app/components/advanced_search/value.html.erb
+++ b/app/components/advanced_search/value.html.erb
@@ -1,5 +1,7 @@
 <% invalid_value = @condition.errors.include?(:value) %>
-<% operator_exists = @condition.operator.blank? %>
+<% operator_exists =
+  @condition.operator.blank? ||
+    %w[exists not_exists].include?(@condition.operator) %>
 <div
   class="
     form-field w-6/12 value <%= 'invalid' if invalid_value %>
--- a/app/components/advanced_search/value.html.erb
+++ b/app/components/advanced_search/value.html.erb
@@ -1,9 +1,9 @@
 <% invalid_value = @condition.errors.include?(:value) %>
-<% exists_operator = %w[exists not_exists].include?(@condition.operator) %>
+<% operator_exists = @condition.operator.blank? %>
 <div
   class="
     form-field w-6/12 value <%= 'invalid' if invalid_value %>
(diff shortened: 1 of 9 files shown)


app/components/advanced_search/value.html.erb
```diff
@@ -1,5 +1,11 @@
 <% invalid_value = @condition.errors.include?(:value) %>
-<div class="form-field w-6/12 value <%= 'invalid' if invalid_value %>">
+<% exists_operator = %w[exists not_exists].include?(@condition.operator) %>
+<div
+  class="
+    form-field w-6/12 value <%= 'invalid' if invalid_value %>
+    <%= 'invisible' if exists_operator %>
```
ericenns, line 6: **nitpick (blocking):** Could we change this to be `invisible` if existance operator or operator is not set. That way if the user loads the advanced search dialog without any filters it wouldn't display the value input until an operator is selected.

This change would also solve the problem where a user inputs a value before selecting an operator which currently clears the value.

app/javascript/controllers/advanced_search_controller.js
```diff
@@ -125,11 +125,15 @@ export default class extends Controller {
       ),
     ].indexOf(condition);
 
-    if (["in", "not_in"].includes(operator)) {
+    if (["exists", "not_exists"].includes(operator)) {
+      value.classList.add("invisible");
```
ericenns, line 129: **nitpick (blocking):** I think you also need to clear the input value, I know it is ignored on the server side. But in case that behaviour changes I think it should clear it here as well.

ericenns (changes requested):
I like that the value input is hidden when selecting `exists` or `not_exists` operator, but I think we should by default hide the value input until an operator has been selected.

deepsidhu85 (approved):
Looks good to me!

ericenns (approved):
Looks good to me!

PR #923, opened 2025-01-31
UI: Create Metadata Template

deepsidhu85:
## What does this PR do and why?
_Describe in detail what your merge request does and why._

This PR adds in the UI to create a new metadata template for projects and groups.

## Screenshots or screen recordings
_Screenshots are required for UI changes, and strongly recommended for all other pull requests._

image

Screenshot from 2025-01-31 18-04-39

**Duplicate name:**

image

**New template dialog only enables the submit button when fields are added to the selected list and the name has a required attribute..however if a user tries to bypass these via inspect element the errors are caught and displayed in an alert:**

image

image

## How to set up and validate locally
_Numbered steps to set up and validate the change are strongly suggested._

1. Login as admin@email.com
2. Go to project `http://localhost:3000/staphylococcus/staphylococcus-saprophyticus/outbreak-2024/-/metadata_templates` 
3. Click the `New Template` button
4. Move some fields from the `Available` list to the `Selected` list
5. Provide a name for the template and optional description
6. Click `Submit`
7. From the rails console run `MetadataTemplate.last`
8. Verify the template was created with the fields, name, and optional description
9. Repeat steps 3 though 7 for group `http://localhost:3000/-/groups/staphylococcus/-/metadata_templates`

## PR acceptance checklist
This checklist encourages us to confirm any changes have been analyzed to reduce risks in quality, performance, reliability, security, and maintainability.

- [x] I have evaluated the [PR acceptance checklist](https://phac-nml.github.io/irida-next/docs/development/development_processes/code_review#acceptance-checklist) for this PR.

PR #924, opened 2025-02-04
UI: Edit Metadata Template 

deepsidhu85:

This PR adds in the UI to edit metadata templates for projects and groups.

image

image

image

image

**Edit template dialog only enables the submit button when fields are added to the selected list and the name has a required attribute..however if a user tries to bypass these via inspect element the errors are caught and displayed in an alert:**

image

image

1. Login as `admin@email.com`
2. Follow the instructions in https://github.com/phac-nml/irida-next/pull/923 to create a template
3. From the metadata templates listing, click `Edit` for the template
4. Verify that the `available` list doesn't include the `selected` items
5. Update the `fields `(remove and add some) and update the `name` (make note of which fields are in the `selected` list)
6. Verify in the table listing the template `name` has updated
7. Click `Edit` for the template again
8. Verify that the `available` list doesn't include the `selected` items and that the `selected` items now includes the items from the selected list in `step 5`, as well as the `name`

## PR acceptance checklist

### DIFF
--- a/app/controllers/concerns/metadata_template_actions.rb
+++ b/app/controllers/concerns/metadata_template_actions.rb
@@ -122,7 +122,7 @@ module MetadataTemplateActions # rubocop:disable Metrics/ModuleLength
 
   def metadata_template_fields
     @current_template_fields = if params.key?(:metadata_template) && metadata_template_params.key?(:fields)
-                                 metadata_template_params[:fields].uniq
+                                 metadata_template_params[:fields]
                                else
                                  @metadata_template.nil? ? [] : @metadata_template.fields
                                end
--- a/app/controllers/concerns/metadata_template_actions.rb
+++ b/app/controllers/concerns/metadata_template_actions.rb
@@ -7,7 +7,11 @@ module MetadataTemplateActions # rubocop:disable Metrics/ModuleLength
   included do
     before_action proc { namespace }
     before_action proc { metadata_template }, only: %i[destroy edit show update]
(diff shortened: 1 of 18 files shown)


joshsadam:
My only comment is that when you complete the edit, you do a full page reload, which I am assuming is because it is the same form as the create, but the table does not need to be updated so that is kind of a waste of bandwidth.

app/controllers/concerns/metadata_template_actions.rb
```diff
@@ -117,7 +122,7 @@ def metadata_template
 
   def metadata_template_fields
     @current_template_fields = if params.key?(:metadata_template) && metadata_template_params.key?(:fields)
-                                 metadata_template_params[:fields]
+                                 metadata_template_params[:fields].uniq
```
ericenns, line 125: **nitpick (blocking):** Did you possibly add this `uniq` here before I got you to fix the duplication of fields bug in the `create`? If so I think this can be removed here.

ericenns (changes requested):
Just one minor comment, otherwise this looks good

app/controllers/concerns/metadata_template_actions.rb
deepsidhu85: Thanks completely forgot to remove it from here but yeah it was added before the fix in create. Removed `.uniq` in f5b222532a8259d50bbe773d6924345c530a2abf

ericenns (approved):
Looks good to me!

PR #946, opened 2025-02-14
Workflow Executions: Workflow and Samplesheet validation

ericenns:

This updates the WorkflowExecution model to validate that the workflow exists as specified by `metadata['workflow_name']` and `metadata['workflow_version']`, additionally this refactors and reenables the SamplesWorkflowExecution `samplesheet_params` validation.

These validation errors should really only be producible by using the submitWorkflowExecution mutation from the graphql API.

Submissions through the UI should not be able to produce these errors as we automatically set most of the params, or restrict input to valid items (e.g. attachments specific to the sample and regex).

Submitting a workflow execution via graphql with missing `fastq_1` for first sample and mismatched `puid` for second sample, gives the following:

image

Also updated submission to handle rendering errors from controller when validation failed
image

To induce this run the following in your javascript console, after selecting a valid pipeline version.
```javascript
document.querySelector("input[name='workflow_execution[metadata][workflow_version]']").value = "invalid"
```

1. Launch the server (if already running then please stop it and start new)
2. Submit a workflow execution via the graphql api following directions [here](https://phac-nml.github.io/irida-next/docs/extend/graphql_workflow_submission) with the following notes:
   * Make an error in the samplesheet params
   * Omit params
   * incorrect PUID
   * wrong attachment gids
3. Verify that the errors are caught and returned

## PR acceptance checklist

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1861,6 +1861,8 @@ en:
       add_all: Add all
       remove_all: Remove all
   workflow_executions:
+    create:
+      error_message: 'The following errors prevented the Workflow Execution from being created:'
     edit_dialog:
       cancel_button: Cancel
       description: You are editing workflow execution '%{workflow_execution_id}'
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1836,6 +1836,13 @@ en:
       group_error: There group has errors
       numeric_operator_error: can only be used on numbers
       uniqueness_error: must be unique
+    workflow_execution_samplesheet_params_validator:
+      attachment_format_error: property '%{property}' must match the expected file format
+      attachment_gid_error: property '%{property}' must be a valid Attachment GID
+      blank_error: property '%{property}' is required and cannot be blank
+      sample_attachment_error: property '%{property}' must belong to the Sample
+      sample_name_error: property '%{property}' must match Sample Name
+      sample_puid_error: property '%{property}' must match Sample PUID
   viral:
     pagy:
       limit_component:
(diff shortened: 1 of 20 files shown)


config/locales/en.yml
```diff
@@ -1836,6 +1839,13 @@ en:
       group_error: There group has errors
       numeric_operator_error: can only be used on numbers
       uniqueness_error: must be unique
+    workflow_execution_samplesheet_params_validator:
+      attachment_format_error: property '%{property}' must match the expected file format
```
JeffreyThiessen, line 1843: We should include the filename and expected value to these errors, it is hard to deduce which file is causing the problem when we have many samples/files.

JeffreyThiessen (changes requested):
Looks good and working as expected. Just needs a change to the error messages to make it easier to fix problems in a failed submission.

ChrisHuynh333 (approved):
Looks good!

config/locales/en.yml
deepsidhu85: Agreed that we should include the file name and expected values to these errors. Otherwise pr works great

deepsidhu85 (approved):
Looks good to me

config/locales/en.yml
ericenns: Error message updated in 2dfb89f, I'm going to leave out display of the filename as that doesn't actually help map back to the `samplesWorkflowExecutionsAttributes` but I have updated it to include regex that is being applied. The index in the error message refers to the position in the array of `samplesWorkflowExecutionsAttributes`. So programmatically this can be easily mapped back.

JeffreyThiessen (approved):
Looks good :+1:

PR #986, opened 2025-03-17
Project Workflow Executions: Destroy Activity

ChrisHuynh333:
This story is part 2 of [STRY0016586](https://publichealthprod.service-now.com/now/nav/ui/classic/params/target/%2Frm_story.do%3Fsys_id%3D5271f91b33e2d21061dd45659d5c7b35) and adds the activities pertaining to project workflow deletion.

image
image

1. Start Sapporo (can be avoided if unavailable/not setup).
2. Select a project, setup automated pipeline(s), and add PE files to project samples to launch the automated pipeline so you have workflow executions to work with at the project level
3. If you didn't set up Sapporo, simply `Cancel` all the workflows so they cancel their run and become deletable.
4. Turn off Sapporo and launch an additional workflow so that it's stuck in `Prepared` state (and therefore non-deletable).
5. Delete 1 workflow using the `Delete` row action, and verify the activity log updates and the `More Details` content is accurate
6. Delete 1 workflow using the `Delete Workflows` button, and verify the activity log updates and the `More Details` content is accurate
7. Select more than 1 workflow and delete them, and verify the activity log updates  and the `More Details` content is accurate
8. Select workflows including the non-deletable workflow, and delete them with the `Delete Workflows` button, and verify the correct # of workflows is displayed (selection number minus 1)  and the `More Details` content is accurate
9. Select only the non-deletable workflow, and delete with the `Delete Workflows` button, and verify the activity log does not update/change

## PR acceptance checklist


deepsidhu85 (dismissed):
This looks great to me and works as expected!

app/javascript/controllers/activities/extended_details_controller.js
```diff
@@ -209,4 +213,27 @@ export default class extends Controller {
...
+      const fragment = document.createDocumentFragment();
+
+      listData.forEach((data) => {
+        const clone = template.content.cloneNode(true);
+        const li = clone.querySelector("li");
+
+        li.querySelector(
+          "span[data-activities--extended_details-target='itemName']",
+        ).textContent = data["workflow_name"];
+        li.querySelector("p > span:nth-child(2)").textContent =
```
joshsadam, line 230: **nitpick:** If I am following the logic correctly here, you are setting the value inside the PuidComponent here. You are giving it the workflow_id not a puid, whichis also in this row. Can you make sure the puid gets this component just to keep consistency.

app/components/activities/dialogs/activity_list_dialog_component.html.erb
```diff
@@ -44,7 +44,7 @@
     <template id="listRow" data-activities--extended_details-target="listRow">
       <li class="pt-3 pb-3 sm:pb-4 sm:pt-4">
```
joshsadam, line 45: **nitpick:** I think this would be better suited as a table with column headers.  When I look at it, I have a hard time figuring out what is what.

app/javascript/controllers/activities/extended_details_controller.js
deepsidhu85: So it's not actually displaying a puid here but instead the workflow_id is being displayed within the puid component. The name of the workflow, since it's automated is the sample name. I don't think we need to display a puid here, but like you mention in the comment below, maybe it would be better suited as a table with the columns workflow name and workflow id

app/components/activities/dialogs/activity_list_dialog_component.html.erb
ChrisHuynh333: Converted listing to use the table component 55d536ef627558b3c70b4184401f7ca23ce6f7cf

joshsadam: Looks great, thanks

joshsadam (approved):

deepsidhu85 (approved):
Looks good to me! Thanks for making those changes

PR #1047, opened 2025-04-24
Fix for importing malformed files

ksierks:
It appears that a user tried importing a file with a combination of different line endings, causing the file to be malformed.

The accepted EOL characters:
- Windows uses carriage return and line feed ("\r\n") 
- Macintosh uses just carriage return ("\r")
- Unix uses just line feed ("\n")

This PR displays an error message when the imported file is malformed.

STRY0016977

image

1. Navigate to a samples page that the logged in user has access to update.
2. From the `Sample Actions` dropdown, under the `Import` section, hit the `Import Metadata` button.
3. Select a file that has a combination of different line endings. (Verify with Notepad++ or `cat -e`)
4. Select the `Sample ID Column`.
5. Hit the `Import Metadata` button.
6. Verify the sample metadata updated successfully.

### DIFF
--- a/app/services/base_spreadsheet_import_service.rb
+++ b/app/services/base_spreadsheet_import_service.rb
@@ -41,7 +41,7 @@ class BaseSpreadsheetImportService < BaseService
     raise FileImportError, I18n.t('services.spreadsheet_import.invalid_file_extension')
   end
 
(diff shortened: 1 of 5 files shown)


ericenns (approved):
Looks good to me!

PR #1062, opened 2025-05-06
Import Metadata UI: Enhancements

ChrisHuynh333:
This PR adds some additional enhancements to the import metadata UI:
- Auto-selects sample column
- Metadata selections don't reset if sample column selection is changed
- Error state if a spreadsheet is uploaded with no metadata

Error state:
image

Test sample autoselection:
1. Navigate to a group or project you can import metadata to, and upload `/test/fixtures/files/metadata/valid.csv`.
2. Verify `sample_name` is auto-selected, and metadatafields 1, 2, and 3 are displayed in the selected metadata list.
3. Verify the following:
- Moving all the metadata to the `Available` list disables the submit button
- De-selecting sample ID column disables the submit button and hides the metadata listing
- Selecting a field other than `sample_name` for sample ID column leaves only the unselected metadatafield values in the metadata listing, and `sample_name` is not there.

Test Error state:
1. Upload the file `test/fixtures/files/batch_sample_import/project/valid.csv`
2. Verify the error message appears (because only sample, description, and project puid columns are available in the spreadsheet, all of which are ignored).
3. Verify the submit button is disabled.

## PR acceptance checklist


test/system/groups/samples_test.rb
```diff
@@ -820,14 +819,13 @@ def retrieve_puids
       click_button I18n.t('shared.samples.actions_dropdown.import_metadata')
       within('#dialog') do
         attach_file 'file_import[file]', Rails.root.join('test/fixtures/files/metadata/valid_with_puid.csv')
-        find('#file_import_sample_id_column', wait: 1).find("option[value='sample_puid']").select_option
-        within "ul[id='available']" do
+        within "ul[id='Available']" do
```
deepsidhu85, line 822: Can you update this id and the one for `Selected` below to use translations as that's how we pass them into the two lists js controller. This comment applies to other locations within this file and other test files for which changes were made to in the PR

deepsidhu85 (changes requested):
This is looking good, just a comment which applies to the test files in this PR

deepsidhu85:
While testing this out I observed some weird behaviour. See screencast below

[Screencast from 2025-05-08 15-02-34.webm](https://github.com/user-attachments/assets/1f89d8b5-fc0d-4c2c-972a-2de3a29f86e7)

app/javascript/controllers/metadata/file_import_controller.js
```diff
@@ -91,36 +65,52 @@ export default class extends Controller {
...
+        break;
+      }
+    }
     this.#enableTarget(this.sampleIdColumnTarget);
   }
 
   #addMetadataColumns() {
     const ignoreList = [
-      "sample id",
+      "sample_name",
```
joshsadam, line 85: **nitpick:** Can you move this list into a private class variable?  This improves readability, avoids re-creating the array on every method call.

app/javascript/controllers/metadata/file_import_controller.js
```diff
@@ -131,11 +121,40 @@ export default class extends Controller {
...
+
+  #enableDialogState() {
+    this.#enableTarget(this.submitButtonTarget);
+    if (this.hasMetadataColumnsTarget) {
+      this.#addMetadataColumns();
+    }
+  }
+
+  #disableErrorState() {
+    this.errorTarget.classList.add("hidden");
```
joshsadam, line 145: **nitpick:** Need to add aria-hidden

app/javascript/controllers/metadata/file_import_controller.js
```diff
@@ -131,11 +121,40 @@ export default class extends Controller {
...
+  }
+
+  #enableErrorState() {
+    this.errorTarget.classList.remove("hidden");
```
joshsadam, line 149: **nitpick:** Will need to remove aria-hidden

app/javascript/controllers/metadata/file_import_controller.js
```diff
@@ -131,11 +121,40 @@ export default class extends Controller {
 
   #disableTarget(target) {
     target.disabled = true;
```
joshsadam, line 123: **nitpick:** Add aria-disabled here

joshsadam (changes requested):
Just a few quick things I noticed

ChrisHuynh333:
> While testing this out I observed some weird behaviour. See screencast below
> Screencast.from.2025-05-08.15-02-34.webm

This is current expected behaviour:
- sample_name isn't added to metadata when unselected as it's part of the ignore list.
- when a metadata field is selected for sample, it's removed from the metadata listing, and if then unselected from sample selection, is readded to selected list, regardless of where it was previously. I didn't think it was worth adding logic to re-add it to the available list if it was previously there (could do so if desired though). If you were wondering about my PR description `Metadata selections don't reset if sample column selection is changed`, if you test on main, and you were to add all the metadata to the available list, and then changed sample selection, all the metadata would move back to the selected list, and that is what I'm trying to prevent

app/javascript/controllers/metadata/file_import_controller.js
ChrisHuynh333: Added in 6d28d0933b2db50a54f5374216f231a4fa600f9c

app/javascript/controllers/metadata/file_import_controller.js
ChrisHuynh333: Thanks for catching these 6d28d0933b2db50a54f5374216f231a4fa600f9c

app/javascript/controllers/metadata/file_import_controller.js
ChrisHuynh333: 6d28d0933b2db50a54f5374216f231a4fa600f9c

app/javascript/controllers/metadata/file_import_controller.js
ChrisHuynh333: 6d28d0933b2db50a54f5374216f231a4fa600f9c

joshsadam:
> > While testing this out I observed some weird behaviour. See screencast below
> > Screencast.from.2025-05-08.15-02-34.webm
> 
> This is current expected behaviour:
> 
> * sample_name isn't added to metadata when unselected as it's part of the ignore list. (I was contemplating adding to the dialog description the ignore list so it's less confusing to the user)
> * when a metadata field is selected for sample, it's removed from the metadata listing, and if then unselected from sample selection, is readded to selected list, regardless of where it was previously. I didn't think it was worth adding logic to re-add it to the available list if it was previously there (could do so if desired though). If you were wondering about my PR description `Metadata selections don't reset if sample column selection is changed`, if you test on main, and you were to add all the metadata to the available list, and then changed sample selection, all the metadata would move back to the selected list, and that is what I'm trying to prevent

I think the simplest way to remove confusion is to reset the state below the changed field to it's original.  So in this case, since all the fields are selected initially, if you change the sample id column, revert this back to all fields being selected.

deepsidhu85:
> ou change the sample id co

Yeah I think this would be the simplest and clearest way!

ChrisHuynh333:
> > ou change the sample id co
> 
> Yeah I think this would be the simplest and clearest way!

Reverted in 4277116b280e550b7d6b5b276fad870693b3b818

joshsadam (approved):

deepsidhu85 (approved):
Looks good to me!

PR #1163, opened 2025-07-30
Workflow execution error refactor

### LINKED ISSUE #1155: Workflow status job can crash when namespace/files are missing.
Workflow status job should exit with appropriate error when namespace/files are removed mid job. This can happen when a workflow is started (user or automated), and then the files/namespace are deleted while the workflow is running.

Above case should also send cancel request to WES.

JeffreyThiessen:
Fixes #1155 

Refactored how jobs check if a workflow execution object is in the correct state before starting the related service.
This pre check also sets the state to error and queue's a cleanup job if the workflow execution is not in a recoverable state (such as missing a namespace because a user deleted it while the jobs are running).

Comments have been added to note what needs to be changed in future refactors to simplify and correct code flow. These changes were left off of this PR as it already was getting quite large.

## PR acceptance checklist


.vscode/settings.json
ericenns, line 1: **nitpick (if-minor):** I noticed there are a lot of Canadian English spellings here, thoughts on just adding this extension and adding `en-CA` to the language setting s that stuff like `colour` or `Cancelation` don't have to be added? https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker-canadian-english

app/models/workflow_execution.rb
```diff
@@ -25,6 +25,7 @@ class WorkflowExecution < ApplicationRecord
   accepts_nested_attributes_for :samples_workflow_executions
 
   validates :metadata, presence: true, json: { message: ->(errors) { errors }, schema: METADATA_JSON_SCHEMA }
+  validate :namespace
```
ericenns, line 28: **nitpick (blocking):** Are you trying to validate presence of namespace here? if so it should be `validates :namespace, presence: true`. Also we could consider to add `, if: :initial?` here to only validate on create with `on: :create` so that it wouldn't fail validate later if the namespace was deleted.

app/models/workflow_execution.rb
```diff
@@ -25,6 +25,7 @@ class WorkflowExecution < ApplicationRecord
...
   validate :validate_namespace
```
ericenns, line 29: **nitpick (blocking):** Can you add a `, on: :create` here so that this validation is only called when a workflow execution is created, since `namespace` should never be updated after the fact.

ericenns (changes requested):
I left a few comments, in addition to those could you also add tests for the new workflow_execution base job that you created. mainly just to test the logic in the methods not queuing the jobs.

.vscode/settings.json
JeffreyThiessen: Added in https://github.com/phac-nml/irida-next/pull/1163/commits/4b73c59298c054381086234feb4d39b7af0b3bab. Although it seems the jury is still out on "Cancelation"

PR #1165, opened 2025-07-31
Advanced Search: Metadata fields sort

deepsidhu85:

This PR updates the advanced search component to retrieve the namespace metadata_fields in sorted (alphabetically) order. Addresses [STRY0018762] (https://publichealthprod.service-now.com/rm_story.do?sys_id=bd4b79ca474f6250f24c0c21516d4397)

image

1. Create a couple of projects in a group, add samples to the projects, and add metadata to samples
2. View the samples page for the projects, and activate the `Advanced Search`
3. In the field dropdown, verify the metadata fields are sorted alphabetically
4. Now view the samples at the group level and activate the `Advanced Search`
5.  In the field dropdown, verify the metadata fields are sorted alphabetically

## PR acceptance checklist

### DIFF
--- a/app/controllers/concerns/metadata.rb
+++ b/app/controllers/concerns/metadata.rb
@@ -32,7 +32,7 @@ module Metadata
   # Builds list of fields available for advanced search
   def advanced_search_fields(namespace)
     @sample_fields = %w[name puid created_at updated_at attachments_updated_at]
-    @metadata_fields = namespace.metadata_fields
+    @metadata_fields = namespace.metadata_fields.sort
   end
 
   # Sets default metadata template if none selected
--- a/app/controllers/concerns/metadata.rb
+++ b/app/controllers/concerns/metadata.rb
@@ -32,7 +32,7 @@ module Metadata
   # Builds list of fields available for advanced search
   def advanced_search_fields(namespace)
     @sample_fields = %w[name puid created_at updated_at attachments_updated_at]
-    @metadata_fields = namespace.metadata_fields.sort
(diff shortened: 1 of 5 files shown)


ericenns (commented):
So this works but not quite sure this is how we want to solve this. Only reason being is that the `metadata_fields` in `app/models/group.rb` already returns the list sorted. Instead I think it would make sense to update `metadata_fields` in `app/models/namespace.rb` to sort. That way other methods that call `metadata_fields` would also benifet from it being sorted e.g. template creation (we do an extra sort there currently), and ordering of metadata fields when viewing samples in a project.

We can discuss this if you like, since there are other cases where we sort the `metadata_fields` in views and if we could remove that it would simplify the code. (search for `.metadata_fields` in the code)

deepsidhu85:
> So this works but not quite sure this is how we want to solve this. Only reason being is that the `metadata_fields` in `app/models/group.rb` already returns the list sorted. Instead I think it would make sense to update `metadata_fields` in `app/models/namespace.rb` to sort. That way other methods that call `metadata_fields` would also benifet from it being sorted e.g. template creation (we do an extra sort there currently), and ordering of metadata fields when viewing samples in a project.
> 
> We can discuss this if you like, since there are other cases where we sort the `metadata_fields` in views and if we could remove that it would simplify the code. (search for `.metadata_fields` in the code)

Definitely a good idea to do the sort in one place. Updated in 3a3e6108f7bbce539aeef9e4ae7f07fdf6ebb60f

ericenns (approved):
Looks good to me!

PR #1667, opened 2026-03-23
GraphQL: BulkUpdateSampleMetadata

ChrisHuynh333:
This PR is part 2 of 2 for [STRY0019725](https://publichealthprod.service-now.com/nav_to.do?uri=rm_story.do?sys_id=086ebf733ba532505c163064c3e45abc%26sysparm_view=scrum) and creates the new `BulkUpdateSampleMetadata` mutation that now accepts multiple sample inputs and uses the new `metadata/bulk_update_service`. This will allow users to update multiple samples within a namespace at once.

Mutation example:
```
mutation updateMetadata {
  bulkUpdateSampleMetadata(input: { 
    metadata: {
      INXT_SAM_A2EBWJDFW6: {new_metadata_1:"value1", new_metadata_2:"value2"},
      sample_name_1: {new_metadata_3:"value3", new_metadata_4:"value4"}
    },
    projectPuid:"INXT_PRJ_A2EBWJDDRH"}) {
    	overallStatus,
        status,
        errors {
          path
          message
        }
      }
}
```

Access the API through `http://localhost:3000/graphiql` and test the following cases using multiple samples at once:
1. Test updating metadata using `project_puid`, `project_gid`, `group_puid`, and `group_gid`.
2. Test updating metadata using sample `name`, `gid`, and `puid` as identifiers. Note: use stringified JSON if the identifier has special characters that need to be escaped
3. Test with invalid sample identifiers and verify appropriate errors occur
4. Test where all, partial, and none of the metadata updates will work, and verify the appropriate errors occur
5. Test overwriting metadata and verify the metadata updates and the old value is overwritten
6. Test updating metadata with the same existing value, and verify an error stating metadata is unchanged occurs.

## PR acceptance checklist


deepsidhu85 (changes requested):
Still need to test this out but just have a few comments to start

app/graphql/mutations/bulk_update_sample_metadata.rb
```diff
@@ -0,0 +1,144 @@
+# frozen_string_literal: true
+
+module Mutations
+  # Mutation that updates metadata for samples
+  class BulkUpdateSampleMetadata < BaseMutation # rubocop:disable Metrics/ClassLength
+    description 'Update metadata for samples.'
+
+    argument :bulk_metadata, GraphQL::Types::JSON, required: true,
```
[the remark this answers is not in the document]
ChrisHuynh333: Changed in b060c9251c8c8956323ff2aebbc538bbb86c4aa2

ericenns (changes requested):
Just leaving my initial review here as it suggests some drastic changes, basically I think the `status` field should be more like the `status` from `updateSampleMetadata` e.g.

`updateSampleMetadata`:
```json
"status": {
  "added": [],
  "updated": [],
  "deleted": [],
  "not_updated": [],
  "unchanged": [
    "new_metadata_1",
    "new_metadata_2"
  ]
},
```

suggested `bulkUpdateSampleMedata`:
```json
"status": {
  "INXT_SAM_A2ENQXTKLT": {
    "added": [],
    "updated": [],
    "deleted": [],
    "not_updated": [],
    "unchanged": [
      "new_metadata_1",
      "new_metadata_2"
    ]
  },
  ...
},
```

This way a tool using the API has a clearer idea of what changes have been persisted, which is easier then parsing a status message string. This would also mean that it should not have any errors present, errors should only occur when sample could not be located, or permissions prevent update.

ChrisHuynh333:
> Just leaving my initial review here as it suggests some drastic changes, basically I think the `status` field should be more like the `status` from `updateSampleMetadata` e.g.
> 
> `updateSampleMetadata`:
> 
> ```json
> "status": {
>   "added": [],
>   "updated": [],
>   "deleted": [],
>   "not_updated": [],
>   "unchanged": [
>     "new_metadata_1",
>     "new_metadata_2"
>   ]
> },
> ```
> 
> suggested `bulkUpdateSampleMedata`:
> 
> ```json
> "status": {
>   "INXT_SAM_A2ENQXTKLT": {
>     "added": [],
>     "updated": [],
>     "deleted": [],
>     "not_updated": [],
>     "unchanged": [
>       "new_metadata_1",
>       "new_metadata_2"
>     ]
>   },
>   ...
> },
> ```
> 
> This way a tool using the API has a clearer idea of what changes have been persisted, which is easier then parsing a status message string. This would also mean that it should not have any errors present, errors should only occur when sample could not be located, or permissions prevent update.

The mutation has been updated to return the full list of `metadata_changes` as status. `Samples` field has been removed as the `status` will now list all the samples with changes (I can put this back in if desired). As discussed, it will also return an `overallStatus` stating `successful`, `successful with errors`, and `unsuccessful`

app/graphql/mutations/bulk_update_sample_metadata.rb
```diff
@@ -0,0 +1,145 @@
...
+             description: 'The Node ID of the project to be updated. For example, `gid://irida/project/a84cd757-dedb-4c64-8b01-097020163077`' # rubocop:disable Layout/LineLength
+    argument :project_puid, ID, # rubocop:disable GraphQL/ExtractInputType
+             required: false,
+             description: 'Persistent Unique Identifier of the project. For example, `INXT_PRJ_AAAAAAAAAA`.'
+    validates required: { one_of: %i[project_id project_puid group_id group_puid] }
+
+    field :errors, [Types::UserErrorType], description: 'A list of errors that prevented the mutation.'
+    field :overall_status, GraphQL::Types::String, null: true,
+                                                   description: 'Overall description of mutation after completion'
+    field :status, GraphQL::Types::JSON, null: true, description: 'The status of the mutation.'
```
deepsidhu85: Lets update this description to `The status of the mutation which lists the fields which were added, removed, updated, and unchanged`

app/graphql/mutations/bulk_update_sample_metadata.rb
```diff
@@ -0,0 +1,145 @@
...
+
+    def resolve(args) # rubocop:disable Metrics/MethodLength,Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      namespace = retrieve_namespace(args)
+
+      if namespace.nil?
```
deepsidhu85: Move the code in this block into it's own method and then here we can just do a `return METHOD_NAME`. The 2 blocks of code below it which also return the user errors, can be updated in a similar fashion

deepsidhu85 (changes requested):
Just a few more comments below

app/graphql/mutations/bulk_update_sample_metadata.rb
ChrisHuynh333: Updated in 5a92e1cd17714f23e567151ca77891f7ad0f711b
added `not_found` as that will be added in a follow up commit as discussed

app/graphql/mutations/bulk_update_sample_metadata.rb
ChrisHuynh333: Thanks, I cleaned up basically anywhere that called the return hash, and made that into its own method with kwargs where we can default to `nil` values since most return values will be nil 
5a92e1cd17714f23e567151ca77891f7ad0f711b

app/services/samples/metadata/bulk_update_service.rb
```diff
@@ -99,15 +113,23 @@ def determine_sample_identifier_type(sample_identifier)
...
+          'gid'
         else
           'name'
         end
       end
 
+      def parse_gid(sample_identifier)
+        IridaSchema.parse_gid(sample_identifier, { expected_type: Sample }).model_id
+      rescue StandardError => e
+        Rails.logger.error "An error occurred: #{e.message}"
```
JeffreyThiessen: Please update this error message to something a bit more descriptive of the error that is occurring, such as: "An error occurred while attempting to parse Sample gid with following message: #{e.message]"

JeffreyThiessen: Also, I think we should explicitly return `nil` in the error case (or its expected that nil is returned from the Rails.logger.error return, add a comment specifying that). When trying to trace the error case through the code I was unsure of how this would behave if this occurred from the original find_sample call.

JeffreyThiessen (changes requested):
Looks good overall. Just one small nitpick on making the code slightly more readable.

app/services/samples/metadata/bulk_update_service.rb
ChrisHuynh333: Updated in bc541b1bb9f246e79821ef32136cb3c284b2f227

JeffreyThiessen (approved):
Looks good 👍

deepsidhu85 (approved):
Looks good to me!

ericenns (approved):
Looks good to me!

PR #1688, opened 2026-03-30
Application Settings: Personal Access Tokens

deepsidhu85:
## What does this PR do and why?
_Describe in detail what your merge request does and why._

This PR updates the `ApplicationSettings` model to include  two new settings `require_personal_access_token_expiry` and `max_personal_access_token_lifetime_in_days` which allows the system admins to make personal access token expiry mandatory and set its maximum lifetime in days. Current defaults are `require_personal_access_token_expiry: false` and  `max_personal_access_token_lifetime_in_days: 365`. The maximum personal access token lifetime in days is only enforced if the expiration is mandatory.

The datepicker has also been updated to display the maximum allowed date for personal access tokens (bots and user) if it is set by the system admin and the related functionality (display of selectable dates) has been modified for this new setting

## Screenshots or screen recordings
_Screenshots are required for UI changes, and strongly recommended for all other pull requests._

image

image

image

image

image

image

image

image

image

## How to set up and validate locally
_Numbered steps to set up and validate the change are strongly suggested._

1. Checkout this branch and verify that all the datepickers (members, group links, bot accounts, user personal access tokens) work as previously intended with the minimum date displayed
2. In rails console, run `ApplicationSetting.current.update(require_personal_access_token_expiry: true)`
3. For datepickers related to personal access tokens (user and bot accounts), verify that the maximum date allowed is displayed, and the calendar does not allow you to input/select dates past this maximum date
4. For datepickers not related to personal access tokens (members, group links) verify there is no maximum date displayed and all dates past the minimum date are selectable
5. Run step 2 again and change `max_personal_access_token_lifetime_in_days` to a different value other than the default `365`. Verify the user -> personal access token create form and the bot modal create forms display the correct maximum allowed date

## PR acceptance checklist
This checklist encourages us to confirm any changes have been analyzed to reduce risks in quality, performance, reliability, security, and maintainability.

- [X] I have evaluated the [PR acceptance checklist](https://phac-nml.github.io/irida-next/docs/development/development_processes/code_review#acceptance-checklist) for this PR.

PR #1830, opened 2026-04-28
Datepicker: Combobox

ChrisHuynh333:
This PR adds a new datepicker utilizing the combobox pattern (inspired by [W3C](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-datepicker/)).

This feature has been added to the group and members tables, and new group link and new member modals. This has not been added to new bot/PAT expiry yet.

Once we've determined this combobox datepicker is stable, we will migrate it from `combobox_datepicker_component` to `datepicker_component` and remove the old datepicker

The major changes vs our original datepicker is the following:
- Decoupled mouse click and keyboard focus logic. The major change associated with this is if a user clicks the input (which opens the datepicker), tabbing will not 'go into' the datepicker, rather it will continue along the DOM, excluding the datepicker. This simplifies focus logic.
- New and updated keyboard features:
  - When users now tab to the datepicker input box, instead of opening upon focus, the user will need to hit `ArrowDown` to open the datepicker
  - Once opened, tabbing/alt-tabbing is trapped into the datepicker
  - To close, the user will use `escape`
  - `space` selects the highlighted date, but does not close the datepicker (`Enter` will select and close)
  - Added `shift+PageUp/Down` shortcuts, which jumps to the previous/next year
  - `pageUp/Down` with and without `shift` will focus on the same date in the different year/month.
  - `Home/End` have been updated to navigate to the start/end of the week

Note: `max_date` logic will still need to be applied once [PR1688](https://github.com/phac-nml/irida-next/pull/1688) is merged. The old `datepicker` in new bot/PAT modal will be replaced with this `combobox_datepicker` at that time as well.

[Screencast from 2026-05-01 17-23-40.webm](https://github.com/user-attachments/assets/14da954d-fa63-4183-88dd-4de83fcfc50d)

Test the datepicker in the group_links and members table, as well as new member and group_links modals.
- Verify date selection still behaves as expected
- Test the above keyboard changes and verify the behave as explained above
- Open/close the datepicker in various ways and verify the caret up/down arrow rotates as expected

## PR acceptance checklist


app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -0,0 +1,773 @@
...
+        year++;
+        month = 0;
+      } else {
+        month++;
+      }
+    }
+    return { year: year, month: month };
+  }
+
+  #getFormattedStringDate(year, monthIndex, date) {
```
joshsadam: Local dates converted via toISOString() can shift the calendar day
 `#getFormattedStringDate` uses `new Date(year, monthIndex, date).toISOString().split("T")[0]`. `toISOString()` is UTC-based, so for timezones ahead of UTC the local calendar date can become the previous calendar day in the string (e.g. Asia/Tokyo). That corrupts `data-date`, min-date comparisons, keyboard navigation targets, and submitted values. Prefer formatting with local getters (`getFullYear`, `getMonth`+1, `getDate`) or a small local YYYY-MM-DD formatter without UTC conversion.

```javascript
export function formatLocalCalendarDate(year, monthIndex, day) {
  const d = new Date(year, monthIndex, day);
  if (
    d.getFullYear() !== Number(year) ||
    d.getMonth() !== Number(monthIndex) ||
    d.getDate() !== Number(day)
  ) {
    throw new RangeError(`Invalid local date: ${year}-${monthIndex + 1}-${day}`);
  }
  const y = d.getFullYear();
  const m = String(d.getMonth() + 1).padStart(2, "0");
  const dd = String(d.getDate()).padStart(2, "0");
  return `${y}-${m}-${dd}`;
}
```

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
```diff
@@ -0,0 +1,427 @@
...
+
+    this.#addCalendarTemplate();
+
+    // Position the calendar
+    this.#initializeDropdown();
+
+    this.#findNextFocusableElement();
+  }
+
+  disconnect() {
```
joshsadam, line 71: `disconnect()` can throw if the calendar node was never resolved
  `#addCalendarTemplate` logs and continues when `getElementById` fails (`this.#calendar` stays undefined), but `disconnect` always calls `this.#calendar.remove()`. Turbo visits/navigation that disconnect the controller in that state will raise. Guard with optional chaining or an `if (this.#calendar)` before remove.

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
```diff
@@ -0,0 +1,427 @@
...
+      error,
+    );
+  }
+
+  // used by datepicker/calendar.js
+  focusDatepickerInput() {
+    this.datepickerInputTarget.focus();
+  }
+
+  focusNextFocusableElement() {
```
joshsadam: `focusNextFocusableElement()` may call focus on undefined
  `#findNextFocusableElement` sets `#nextFocusableElementAfterInput` from `focusable[index + 1]`. If the input is not in the computed focusable list (`index === -1`) or is last, `focus()` is invoked on `undefined` and throws. Worth guarding or falling back (e.g. to `HTMLElement.prototype.focus` on a known container or no-op with a console warning).

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -0,0 +1,773 @@
...
+    return this.clearButtonTarget;
+  }
+
+  focusCurrentDate() {
+    const currentDate =
+      this.calendarTarget.querySelectorAll('[tabindex="0"]')[0];
+    focusDate(this.calendarTarget, currentDate);
+  }
+
+  #getDateAriaLabel(year, month, day) {
```
joshsadam, line 766: Per-day aria-label only defined for "en" and "fr"
  `#getDateAriaLabel` returns `undefined` for other `localeValue` values (e.g. future locales, or values like `"en-CA"` if ever passed). Screen readers then get cells without a proper accessible name. Default to one of the branches, I18n-driven strings, or `Intl.DateTimeFormat` for the active locale.

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -0,0 +1,773 @@
...
+  #selectedDate;
+  #selectedYear;
+  #selectedMonthIndex;
+
+  #minDate;
+  #minDateMessage;
+  #autosubmit;
+
+  idempotentConnect() {
+    console.log(this.localeValue);
```
joshsadam: Remvoe logging

joshsadam (changes requested):
This looks really good.  I added some small code quality improvement requests that I would like to see, even though this is v1.

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: Removed ee51d8b925992445cd9583747a6a64cf8d1a3a38

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: `en` is now default ee51d8b925992445cd9583747a6a64cf8d1a3a38

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
ChrisHuynh333: Actually forgot to remove all this logic since it's unnecessary ee51d8b925992445cd9583747a6a64cf8d1a3a38

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
ChrisHuynh333: ee51d8b925992445cd9583747a6a64cf8d1a3a38

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: Thanks for catching this 88b33f0ac20f6d8c3e5aca374395914387d88310

I had to remove the error catch as `.getDate !== date` will throw an error with how the logic works for keyboard navigation between months. I currently have it where it +/- 7 to the date for `handleVertialNavigation`, and `new Date` can intuitively figure out that date 34 == 3 in a new month, for example. This would require a lot of extra work, specifically going backwards (need to check if previous month ends in 30, 31, or 28/29 for Feb). I could go down that route but I don't think it's necessary?

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -0,0 +1,772 @@
...
+      this.#selectedMonthIndex = this.#todaysMonthIndex + 1;
+    } else {
+      this.#selectedMonthIndex = this.#todaysMonthIndex;
+    }
+
+    this.idempotentConnect();
+  }
+
+  // handles Shift+Tab out of calendar into datepicker input
+  tabBackToInput(event) {
```
joshsadam: This is not wired to anything.  Just legacy code?

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -0,0 +1,772 @@
+import { Controller } from "@hotwired/stimulus";
+import {
+  DAYS_IN_MONTH,
+  CALENDAR_CLASSES,
+} from "controllers/datepicker/constants";
```
joshsadam: I see new constant files in this directory, but this points to the legacy ones.

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
```diff
@@ -0,0 +1,407 @@
...
+      distance: 10,
+      onShow: () => this.#onShow(),
+      onHide: () => this.#onHide(),
+    });
+  }
+
+  #onShow() {
+    document.addEventListener("keydown", this.boundHandleGlobalKeydown);
+    this.#calendar.addEventListener("focusin", this.boundHandleCalendarFocus);
+    this.#arrowSvg.classList.add("rotate-180");
```
joshsadam, line 99: Update `aria-expanded` when you show the input (it's marked as `false` in the UI). And then `false` again in the `onHide` method.

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
```diff
@@ -0,0 +1,407 @@
...
+  handleGlobalKeydown(event) {
+    // Escape: close calendar
+    if (event.key === "Escape") {
+      this.hideCalendar();
+      this.setInputValue(this.#selectedDate);
+      this.datepickerInputTarget.focus();
+      return;
+    }
+
+    // If we tab off the last datepicker element, we want to force focus onto the next focusable element after
```
joshsadam: Tab key handler comment doesn’t match behavior. The comment says tabbing off the last element should move to “the next focusable element after the datepicker input”, but the logic actually cycles focus back to the first calendar element (a focus trap). The trap may be intended (PR body says tabbing is trapped), but the comment is misleading and will cause future “fixes” to regress behavior. Fix: update the comment to reflect the intentional focus trap, or adjust the behavior to match the comment (whichever is desired).

joshsadam (changes requested):
1. We might need to look at the focus styles, as they overlap the current date:

Image

2. If I select the date input and the dropdown opens, then tab to the next input, the dropdown stays open.

Image

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
ChrisHuynh333: Updated in a4737e9ba7352ae9a2863c3376619ecb51911a2b

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
ChrisHuynh333: This is already handled by `floating_dropdown_controller`

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: Updated in a4737e9ba7352ae9a2863c3376619ecb51911a2b

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: Yup, removed in a4737e9ba7352ae9a2863c3376619ecb51911a2b

ChrisHuynh333:
> 1. We might need to look at the focus styles, as they overlap the current date:
> 
> Image
> 2. If I select the date input and the dropdown opens, then tab to the next input, the dropdown stays open.
> 
> Image

I've added `:focus` specific classes for `td` date nodes in `application.tailwind.css` for the datepicker to handle both these cases.
- First to shrink the outline so there's no overlap.
- Secondly, I've specified focus based on `:focus` rather than the default `:focus-visible` in part to handle your second comment. Like the `W3C` datepicker, if the user clicks the text box itself, focus is not shifted, but if the user clicks the arrow, focus is shifted to the tabbable date node (this is newly added in this commit). This requires `:focus` and not `:focus-visible` to enable the styling after clicking.

aae549cd61d0792a5f24004883e62d1e489dcee2

ChrisHuynh333:
> > 1. We might need to look at the focus styles, as they overlap the current date:
> > 
> > Image
> > 2. If I select the date input and the dropdown opens, then tab to the next input, the dropdown stays open.
> > Image
> 
> I've added `:focus` specific classes for `td` date nodes in `application.tailwind.css` for the datepicker to handle both these cases.
> 
> * First to shrink the outline so there's no overlap.
> * Secondly, I've specified focus based on `:focus` rather than the default `:focus-visible` in part to handle your second comment. Like the `W3C` datepicker, if the user clicks the text box itself, focus is not shifted, but if the user clicks the arrow, focus is shifted to the tabbable date node (this is newly added in this commit). This requires `:focus` and not `:focus-visible` to enable the styling after clicking.
> 
> [aae549c](https://github.com/phac-nml/irida-next/commit/aae549cd61d0792a5f24004883e62d1e489dcee2)



> > 1. We might need to look at the focus styles, as they overlap the current date:
> > 
> > Image
> > 2. If I select the date input and the dropdown opens, then tab to the next input, the dropdown stays open.
> > Image
> 
> I've added `:focus` specific classes for `td` date nodes in `application.tailwind.css` for the datepicker to handle both these cases.
> 
> * First to shrink the outline so there's no overlap.
> * Secondly, I've specified focus based on `:focus` rather than the default `:focus-visible` in part to handle your second comment. Like the `W3C` datepicker, if the user clicks the text box itself, focus is not shifted, but if the user clicks the arrow, focus is shifted to the tabbable date node (this is newly added in this commit). This requires `:focus` and not `:focus-visible` to enable the styling after clicking.
> 
> [aae549c](https://github.com/phac-nml/irida-next/commit/aae549cd61d0792a5f24004883e62d1e489dcee2)

Added closing datepicker when tabbing off input (eg: click input which opens calendar, then tab off) c62c22e352d7b2b6dfcc5928b31e9233724fac63

joshsadam (dismissed):

ericenns (changes requested):
This is looking good, but the keyboard behaviour in the grid doesn't quite match the behaviour listed in the wcag page, e.g:

Key | Function
-- | --
Home | Moves focus to the first day (e.g. Sunday) of the current week.
End | Moves focus to the last day (e.g. Saturday) of the current week.
PageUp | Changes the grid of dates to the previous month.Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month.
Shift+PageUp | Changes the grid of dates to the same month in the previous year.Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month.
PageDown | Changes the grid of dates to the next month.Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month.
Shift+PageDown | Changes the grid of dates to the same month in the next year.Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month.

ChrisHuynh333:
> This is looking good, but the keyboard behaviour in the grid doesn't quite match the behaviour listed in the wcag page, e.g:
> 
> Key	Function
> Home	Moves focus to the first day (e.g. Sunday) of the current week.
> End	Moves focus to the last day (e.g. Saturday) of the current week.
> PageUp	Changes the grid of dates to the previous month.Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month.
> Shift+PageUp	Changes the grid of dates to the same month in the previous year.Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month.
> PageDown	Changes the grid of dates to the next month.Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month.
> Shift+PageDown	Changes the grid of dates to the same month in the next year.Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month.

Thanks, 
Updated Home/End to navigate to start/end of week. This will also take into account disabled and outOfMonth dates 6668bc835ac2df09260afa905f551eb51e2f1102

PageUp/Down now navigate to the same date node. This will also take into account `minDate` and has the starting logic/function for verifying `maxDate`. Also cleaned up `pageUp/Down` logic into a single function since there was a lot of shared logic 635410a73b4ec4ed2042503b1d35ca8a30005e54

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
```diff
@@ -0,0 +1,430 @@
...
+      trigger: this.datepickerInputTarget.parentElement,
+      dropdown: this.#calendar,
+      distance: 10,
+      autoSize: false,
+      onShow: () => this.#onShow(),
+      onHide: () => this.#onHide(),
+    });
+  }
+
+  #onShow() {
```
joshsadam, line 95: The popup state is currently announced on the wrapper `div` instead of on the combobox. `FloatingDropdown` manages `aria-expanded` on `trigger`, and this controller passes `datepickerInputTarget.parentElement`, while the element with `role="combobox"` remains `aria-expanded="false"`. Could you set `aria-expanded` on `datepickerInputTarget` in `#onShow`/`#onHide` or otherwise make the input the managed ARIA element so screen readers receive the open/closed state from the combobox itself?

app/components/combobox_datepicker/v1/calendar/grid_component.html.erb
```diff
@@ -0,0 +1,43 @@
+<div class="justify-center p-1">
+  <table class="min-w-full table-fixed" role="grid" aria-label="<%= t('components.datepicker.aria_label.calendar') %>">
```
joshsadam: The grid is named only as `Calendar`, so screen reader users lose the displayed month/year context when the popup opens or PageUp/PageDown changes months. The APG datepicker labels the grid from the current month/year heading. Could this table use `aria-labelledby` pointing at a current `Month YYYY` element instead of a static `aria-label`?

app/components/combobox_datepicker/v1/calendar/header_component.html.erb
```diff
@@ -0,0 +1,64 @@
+<div class="mb-2 flex items-center justify-between">
```
joshsadam, line 1: When the header controls or PageUp/PageDown change the displayed month, there is no live announcement of the new `Month YYYY`. The APG datepicker makes that month/year text a polite live region and uses it as the grid label. Could you add a current month/year element, update it from Stimulus whenever the calendar reloads, and give it `aria-live="polite"`?

app/components/combobox_datepicker/v1/input_field_component.html.erb
```diff
@@ -0,0 +1,84 @@
...
+    <% if @required %>
+      aria-required="true"
+    <% end %>
+    <% if @errored %>
+      aria-describedby="<%= @error_id %>"
+      aria-invalid="true"
+    <% end %>
+  >
+
+  <div
```
joshsadam: The calendar opener should be an actual control. In the APG datepicker, the icon is a `button` with `tabindex="-1"` and an accessible name such as "Choose Date", which keeps it out of the Tab sequence but still exposes it to touch screen readers. This `div` has no role/name, so VoiceOver/TalkBack users do not get an operable choose-date control. Could you render it as a `button type="button"` with `aria-label` and `tabindex="-1"`?

app/components/combobox_datepicker/v1/input_field_component.html.erb
```diff
@@ -0,0 +1,84 @@
...
+    <% end %>
+  </label>
+<% end %>
+
+<div class="relative max-w-sm" aria-live="polite">
+  <div class="pointer-events-none absolute inset-y-0 inset-s-0 flex items-center ps-3.5">
+    <%= icon(:calendar_dots, color: nil, size: nil, class: "size-6 text-slate-500 dark:text-slate-400") %>
+  </div>
+
+  <input
```
joshsadam, line 20: The combobox does not have a persistent date-format description. The APG datepicker associates the format hint with the input through `aria-describedby`, while this component only sets `aria-describedby` when there is an error. Could you add a visually hidden format hint and keep its id on the input, appending the error id when validation fails and restoring the hint id when the error is cleared?

app/components/combobox_datepicker/v1/calendar/grid_component.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <thead class="mb-1">
+      <tr>
+        <% @days_of_the_week.each do |day| %>
+          <th scope="col" class="dow h-6 text-center text-sm font-medium text-slate-500 dark:text-slate-400">
+            <%= day %>
```
joshsadam: The weekday headers render abbreviated labels from the locale file. The APG datepicker uses the `abbr` attribute so screen readers can announce the full weekday name while the visual header stays compact. Could you pass both short and full day names, or add full-day translations, and render the headers as `th abbr="Sunday"` etc.?

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -0,0 +1,769 @@
...
+      this.focusCurrentDate();
+    } else {
+      this.comboboxDatepickerV1InputOutlet.hideCalendar();
+      this.comboboxDatepickerV1InputOutlet.focusDatepickerInput();
+    }
+  }
+
+  // clear selection by clicking clear button
+  clearSelection() {
+    this.comboboxDatepickerV1InputOutlet.setInputValue("");
```
joshsadam, line 571: `clearSelection()` updates the input but leaves focus on the Clear button inside the still-open dialog. After a dialog action that clears or finishes the date value, keyboard and screen reader users should be returned to the combobox. Could this hide the calendar and focus `datepickerInputTarget` after clearing, including the non-autosubmit path?

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
```diff
@@ -0,0 +1,430 @@
...
+    }
+    return false;
+  }
+
+  // without event.preventDefault() on Enter, form is submitted
+  handleEnterDirectInput(event) {
+    if (event.key === "Enter") {
+      event.preventDefault();
+      this.directInput(event);
+    } else if (event.key === "ArrowDown") {
```
joshsadam, line 313: This handler opens the APG `Down Arrow` / `Alt+Down Arrow` path, but it does not prevent the browser default for `ArrowDown`. That can scroll the page or interact with browser chrome before the dialog opens. Could you call `event.preventDefault()` before `toggleCalendar(event)` for the ArrowDown branch?

app/components/combobox_datepicker/v1/calendar/grid_component.html.erb
```diff
@@ -0,0 +1,43 @@
+<div class="justify-center p-1">
```
joshsadam, line 1: The APG datepicker includes a polite live region at the bottom of the dialog for keyboard help when focus enters the grid. Without an equivalent here, the custom Home/End/PageUp/PageDown behavior is discoverable only by trial and error. Could you add a localized help string in an `aria-live="polite"` element, announce it once per popup open when focus enters the grid, and reset that flag when the popup closes?

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -0,0 +1,769 @@
...
+    } else if (this.yearTarget.value > this.#selectedYear) {
+      this.nextYear();
+    }
+  }
+
+  previousYear() {
+    // if minDate exists, check if user tried to hard type in a year amount earlier than minDate's year
+    if (this.#minDate) {
+      const minDate = new Date(this.#minDate);
+      const minYear = minDate.getFullYear();
```
joshsadam: `new Date(this.#minDate).getFullYear()` can read the previous year for `YYYY-MM-DD` strings in time zones west of UTC. For example, `new Date("2026-01-01")` is Dec 31, 2025 locally in this app timezone, so a Jan 1 min date can allow the year input to move back to 2025. Could this parse the ISO parts directly or use `getUTCFullYear()` consistently with the existing `getUTCMonth()` call?

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -0,0 +1,769 @@
...
+      targetDateNode = getDateNode(this.calendarTarget, targetFullDate);
+    }
+    focusDate(this.calendarTarget, targetDateNode);
+  }
+
+  // navigate to start or end of the week by Home and End keys, excluding disabled and outOfMonth dates
+  #handleWeekNavigation(event, direction) {
+    const currentWeekNode = event.target.parentElement;
+    const activeDateNodes = currentWeekNode.querySelectorAll(
+      ':not([aria-disabled="true"]):not([data-date-within-month-position="outOfMonth"]',
```
joshsadam: The Home/End handler currently builds an invalid selector: the second `:not(...)` is missing its closing `)`. Pressing Home or End can throw a `SyntaxError` instead of moving focus to the start/end of the week, which is one of the APG grid keys. Could you fix the selector and add coverage for both keys so this path is exercised?

joshsadam (changes requested):

app/components/combobox_datepicker/v1/input_field_component.html.erb
ChrisHuynh333: Updated 4c2e141ed57b49685214de988cb77b3efa875bed

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: Updated 4c2e141ed57b49685214de988cb77b3efa875bed

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
ChrisHuynh333: Updated 4c2e141ed57b49685214de988cb77b3efa875bed

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: Updated 4c2e141ed57b49685214de988cb77b3efa875bed

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: Updated 4c2e141ed57b49685214de988cb77b3efa875bed

app/components/combobox_datepicker/v1/calendar/grid_component.html.erb
ChrisHuynh333: Updated 4c2e141ed57b49685214de988cb77b3efa875bed

app/components/combobox_datepicker/v1/calendar/grid_component.html.erb
ChrisHuynh333: Hidden aria-live region in heard with `MM YYYY` and linked to here with `aria-labelledby` 0467022a02820ed2e46ec5b4760774b990da753a

app/components/combobox_datepicker/v1/calendar/header_component.html.erb
ChrisHuynh333: Added in 0467022a02820ed2e46ec5b4760774b990da753a

app/components/combobox_datepicker/v1/calendar/grid_component.html.erb
ChrisHuynh333: I'm going to leave this for now, although I agree it would be beneficial to have these somewhere, it's not really viable to use `aria-keyshortcuts`, and W3C's aria-live is a very basic navigation message and also doesn't go into these advances shortcuts

app/components/combobox_datepicker/v1/input_field_component.html.erb
ChrisHuynh333: Going to leave this for now, although we ask for a specific date format, the backend will accept any valid date format and so doesn't prevent a user's input no matter the format

app/javascript/controllers/combobox_datepicker/v1/input_controller.js
ChrisHuynh333: Moved handling `aria-expanded` out of floating dropdown and manually handling in input controller b24689370f5fda3f38ed103a73a609c686064fe1

app/components/combobox_datepicker/v1/input_field_component.html.erb
```diff
@@ -0,0 +1,87 @@
...
+    <% if @required %>
+      aria-required="true"
+    <% end %>
+    <% if @errored %>
+      aria-describedby="<%= @error_id %>"
+      aria-invalid="true"
+    <% end %>
+  >
+
+  <button
```
ericenns: nitpick(blocking): Can you update this to include the same styling as the dropdown button on the combobox component. That way we get visual indication that it is a button on hover and have the correct cursor displayed.

ericenns (changes requested):
This is looking great, just one minor suggestion.

app/components/combobox_datepicker/v1/input_field_component.html.erb
ChrisHuynh333: Updated in fb7d09dd755bf4cbcc282dacf153d3ef2f932248

ericenns (dismissed):
Looks good to me!

ericenns (approved):

PR #1860, opened 2026-05-07
Client File Import

ksierks:
This PR implements client-side functionality for the file import, moving metadata processing and sample updates from the server to the browser using Web Workers and GraphQL mutations. Future work will include success message, progress bar, and error messages.

#### Key Changes
- Implemented a web worker (linelist_import_worker.js) to extract sample data from uploaded file and prepare it for batch GraphQL mutations
- Updated the LinelistImportDialogComponent to accept a group or project
- Reused file_import_controller.js in the new linelist_import_controller.js

N/A

#### Enable Feature Flag
1. Start the server: 
```bash
bin/setup
```
2. Open Rails console: 
```bash
bin/rails console
```
3. Enable the flag: 
```bash
Flipper.enable(:client_linelist_imports_v1)
```

#### Manual Tests
Verify Data Import:
1. Create a CSV or Excel file with sample metadata.
1. Login to irida-next.
1. Navigate to a project or group with existing samples that the user has access to update.
1. Select `Sample Actions` -> `Import metadata` from the dropdown menu on the samples page.
1. Upload the test file.
1. Verify sample table refreshed with the imported data.

Verify Web Worker Execution:
1. Open browser `DevTools` (F12)
2. Go to `Console` tab and check for:
    - Success messages: "Main thread received: {status: 'done', result: ...}"
    - Any error messages from the worker
3. Check `Network` tab to verify BulkUpdateSampleMetadata GraphQL mutation calls

## PR acceptance checklist


app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,120 @@
-import { Controller } from "@hotwired/stimulus";
+import * as XLSX from "xlsx";
+import _ from "lodash";
```
joshsadam: Lodash is only used here for `pick` and `chunk`. Both are easy to replace with a few lines of native JS — see the next two comments. Would it be possible to drop this dependency and the `vendor/javascript/lodash.js` file?

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,120 @@
...
+import Controller from "controllers/metadata/file_import_controller";
 
+const ROW_CHUNK_SIZE = 2;
```
joshsadam: `linelist_export_worker.js` already defines a small `chunk` helper (lines 34–41) that can be copied here. No library needed for this.

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,120 @@
...
+
+  #processRows() {
+    const allRows = XLSX.utils.sheet_to_json(this._worksheet, {
+      header: this.headers,
+      range: 1,
+    });
+
+    const rows = allRows.map((row) => [
+      row[this.sampleIdColumnTarget.value],
+      _.pick(row, this.columns),
```
joshsadam: Nit: `_.pick(row, this.columns)` can be replaced with native JS:
```javascript
Object.fromEntries(this.columns.map((col) => [col, row[col]]))
```

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,120 @@
...
+    super.handleSubmit();
+  }
+
+  #processRows() {
+    const allRows = XLSX.utils.sheet_to_json(this._worksheet, {
+      header: this.headers,
+      range: 1,
+    });
+
+    const rows = allRows.map((row) => [
```
joshsadam, line 71: `this.columns` is populated from all valid non-ignored headers in the file. The sortable-list UI lets the user select a subset of those columns, and that selection ends up in the `file_import[metadata_columns][]` hidden inputs built by the sortable-lists controller — but those inputs are never read on this path.

The server-side `Samples::Metadata::FileImportService` reads the selected columns from the form params and also checks `ignore_empty_values` to decide whether empty cells should clear existing metadata or be skipped.

Can we read the selected column values from the hidden inputs at submit time (e.g. `querySelectorAll('[name="file_import[metadata_columns][]"]')`) and pass the `ignore_empty_values` checkbox state to the worker? Otherwise the client import will update columns the user did not select and handle empty values differently from the server path.

app/components/data_imports/linelist_import_dialog/v2/component.html.erb
```diff
@@ -20,10 +23,12 @@
 
     <%= tag.div id: "import-metadata-dialog-content", data: linelist_data do %>
       <%= form_for(:file_import, url: "#", method: :get, data: {
-        action: "submit->viral--dialog#close submit->linelist-import#submit"
+        action: "submit->viral--dialog#close"
```
joshsadam: The form action `submit->viral--dialog#close` allows the browser to run the form's default GET submission. That navigation fires right after `#processRows()` posts messages to the worker, which would abort the in-flight GraphQL requests before they complete.

The export controller avoids this by calling `event.preventDefault()` and `event.stopPropagation()` in its `submit` handler before starting work. Can we do the same here — either by wiring the form's `submit` event to a method that prevents default, or by adding `type="button"` to the submit button so no form navigation happens?

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,120 @@
...
+      row[this.sampleIdColumnTarget.value],
+      _.pick(row, this.columns),
+    ]);
+
+    const rowChunks = _.chunk(rows, ROW_CHUNK_SIZE);
```
joshsadam: Each 2-row chunk is posted to the worker as a separate message, and the worker starts a new `fetch` for each. Because `onmessage` is `async`, a large file fans out many concurrent GraphQL mutations immediately.

The export worker handles chunking entirely inside the worker, awaiting each request in sequence in a `for...of` loop. Can we move row iteration into the import worker as well and `await` each mutation before starting the next? As part of that, `ROW_CHUNK_SIZE` could live in the worker (matching where `SAMPLE_CHUNK_SIZE` lives in the export worker), and a larger chunk size would reduce round-trips for typical files.

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,120 @@
...
+      );
+
+      worker.onerror = (error) => {
+        console.error("Worker failed:", error.message);
+        this.#terminateWorker();
+      };
+
+      // Listen for messages from the worker
+      worker.onmessage = (event) => {
+        this._chunksReceived++;
```
joshsadam: The worker only surfaces errors when the HTTP request or JSON parsing fails. `bulkUpdateSampleMetadata` can also return domain errors in `data.bulkUpdateSampleMetadata.errors` (e.g. sample not found, duplicate identifier). Those currently come back in a `{ type: "done" }` message and are silently treated as success.

There are two places to fix this:
1. **Worker**: inspect `result.errors` after each mutation and post `{ type: "error" }` when errors are present.
2. **Controller `onmessage`**: check `event.data.type` before incrementing `_chunksReceived`. Right now both `"done"` and `"error"` responses are counted identically, so even if the worker posts an error the controller treats it as success.

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,120 @@
...
+    projectPuid: String,
+  };
+
   connect() {
-    this.worker = null;
+    super.connect();
+    this._chunksReceived = 0;
+    this._hasErrors = false;
+    this._totalChunks = 0;
+    this._worker ||= this.#buildWorker();
```
joshsadam: `||=` means the worker is built once per Stimulus lifecycle and never rebuilt. After `#terminateWorker()` sets `this._worker = null`, any subsequent import will call `this._worker.postMessage(...)` on `null` and throw a `TypeError`.

The export controller avoids this by stopping and recreating the worker at the start of each operation. Can we do the same here — e.g. call `this._worker?.terminate()` and rebuild at the start of `handleSubmit()` instead of lazily initializing in `connect()`?

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,120 @@
...
+  static values = {
+    graphqlUrl: String,
+    groupPuid: String,
+    projectPuid: String,
+  };
+
   connect() {
-    this.worker = null;
+    super.connect();
+    this._chunksReceived = 0;
```
joshsadam: `_chunksReceived` and `_totalChunks` are initialized here but never reset between imports. On a second submit in the same Stimulus session, `_chunksReceived` continues from where the previous run left off, so `#terminateWorker()` fires at the wrong threshold (or not at all).

Can we reset these at the start of `handleSubmit()` instead?

joshsadam (changes requested):
This is a great start.  I have added some comments; if they are to be updated in a future PR, just say that.  Just wanted to highlight what I saw so we could catch everything.

app/javascript/controllers/linelist_import_controller.js
ksierks: Ok. We used to use lodash. I created our own pick & chunk functions and placed them in a collections helper here, 792bae16fa34944afa046fc0932215cf51bb5143.

app/javascript/controllers/linelist_import_controller.js
ksierks: Copied in 792bae16fa34944afa046fc0932215cf51bb5143.

app/javascript/controllers/linelist_import_controller.js
ksierks: Replaced in 792bae16fa34944afa046fc0932215cf51bb5143.

app/javascript/controllers/linelist_import_controller.js
ksierks: I liked how you used state to keep track of when an import is validating, in progress, errored, done, etc. I removed `_totalChunk`s & `_chunksReceived` for now.
414d43ddb76d2c610b3207ab163e4adf5f358ab6

app/javascript/controllers/linelist_import_controller.js
ksierks: 22fc732428e2d5102de732eb8bb40450918cb4ad moves building the worker within `handleSubmit()`.

app/javascript/controllers/linelist_import_controller.js
ksierks: I initially thought we'd want to parallelize the graphql calls. All computations have been moved to the worker in 414d43ddb76d2c610b3207ab163e4adf5f358ab6.
If it's ok, I'd like to keep the chunk size small during implementation.

app/components/data_imports/linelist_import_dialog/v2/component.html.erb
ksierks: Does this work? 2566b1d705c3f4cd23fc632c222ebbde8a94ca8d
I think I may have accidentally called `handleSubmit` on submit form and button click.

app/javascript/controllers/linelist_import_controller.js
ksierks: 834d4f715047b25e82ba669673e67f8a289861d8

app/javascript/controllers/linelist_import_controller.js
ksierks: If it's ok, I'd like to address this in a future PR that displays the errors to the user.

app/javascript/controllers/linelist_import_controller.js
joshsadam: Thanks

app/javascript/utilities/collection.js
```diff
@@ -0,0 +1,41 @@
+/**
```
joshsadam, line 1: Can you add JS tests for these, add a new test file `test/javascript/utilities/collection.test.js`

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,136 @@
...
+      const worksheetName = workbook.SheetNames[0];
+      this._worksheet = workbook.Sheets[worksheetName];
+      this.headers = XLSX.utils.sheet_to_json(this._worksheet, {
+        header: 1,
+      })[0];
+      super.addSampleIDInputOptions();
+    };
+  }
+
+  handleSubmit() {
```
joshsadam: Update this to be, as the submit button can still trigger a reload:
```javascript
handleSubmit(event) {
  event.preventDefault();
  event.stopPropagation();
}
```

joshsadam (changes requested):

app/javascript/utilities/collection.js
ksierks: 30d512f660ffe3117f566863460a35616c20ea10

app/javascript/controllers/linelist_import_controller.js
ksierks: 118ad0394178a3605a63832d2d3214ecf7a9d9e1

app/components/data_imports/linelist_import_dialog/v2/component.html.erb
joshsadam: I would take the submit action of the button in `app/views/shared/samples/metadata/file_imports/_form_fields.html.erb` if you can, and let the form itself handle it.  That way if it is a click, keypress, or whatever the form handles it correctly.

joshsadam: It might be easier to change the type on the form submit button to `button`.  Then you you could get ride of `method: :get, url: "#` on the form itself. and drop the `submit-end`

app/javascript/workers/linelist_import_worker.js
```diff
@@ -0,0 +1,182 @@
+const ROW_CHUNK_SIZE = 2;
+const SUPPORTED_MIME_TYPES = [
+  "text/csv",
+  ".tsv",
```
joshsadam: TSV uploads are always rejected by the worker. `SUPPORTED_MIME_TYPES` includes the literal string ".tsv", which is a file extension, not a MIME type. Browsers report TSV picks as `text/tab-separated-values`

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,138 @@
...
+
+    const ignoreEmptyValues = document.querySelector(
+      '[name="file_import[ignore_empty_values]"]',
+    ).checked;
+
+    const rows = allRows.map((row) => [
+      row[this.sampleIdColumnTarget.value],
+      omitBy(
+        pick(row, selectedMetadataColumns),
+        (value) => ignoreEmptyValues && !value,
```
joshsadam: `ignore_empty_values` is more aggressive client-side than server-side. Predicate `(value) => ignoreEmptyValues && !value `drops `null`, `undefined`, `""`, `0`, `false`. The server path in `app/services/samples/metadata/file_import_service.rb:44` uses `metadata.compact!`, which only drops `nil`. A spreadsheet cell holding 0 (or text "0" coerced to number) will be silently stripped on the client path while it would have been retained on the legacy path. Tighten the predicate (for example, `(value) => ignoreEmptyValues && (value === null || value === undefined || value === "")`).

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,138 @@
...
+  static values = {
+    graphqlUrl: String,
+    groupPuid: String,
+    projectPuid: String,
+  };
+
   connect() {
-    this.worker = null;
+    super.connect();
+    this._filetype = null;
```
joshsadam: `_filetype` vs `_fileType` typo. `connect()` initializes `this._filetype = null`, but `readFile` writes `this._fileType` and `#processRows` reads `this._fileType`. The lowercase field is dead. If the intent of `connect()` was to reset state between Stimulus reconnections, rename to `this._fileType = null`.

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,17 +1,138 @@
...
+  }
+
+  #processRows() {
+    const allRows = XLSX.utils.sheet_to_json(this._worksheet, {
+      header: this.headers,
+      range: 1,
+      defval: null,
+    });
+
+    const selectedMetadataColumns = Array.from(
```
joshsadam, line 62: Use scoped lookups for form inputs. `#processRows` reaches into the global document with `document.querySelectorAll('[name="file_import[metadata_columns][]"]')` and `document.querySelector('[name="file_import[ignore_empty_values]"]')`. Scope these to the controller's element (`this.element.querySelectorAll(...)`) so the controller does not pick up unrelated inputs and so `ignore_empty_values` does not throw on `null.checked` if the partial ever omits the checkbox.

app/javascript/utilities/collection.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Builds a new object containing only the provided keys from the source object.
+ *
+ * @param {Object} item - The source object.
+ * @param {string[]} keys - Keys to copy to the returned object.
+ * @returns {Object} An object containing only the selected keys.
+ */
+export function pick(item, keys) {
+  return Object.fromEntries(keys.map((key) => [key, item[key]]));
```
joshsadam: Right now this will drop keys if the value is `undefined` which is different then tha lodash version.  Either document this in the JSDoc, or a more complete version would be:

```javascript
/**
 * Builds a new object containing only the provided keys from the source object.
 *
 * Keys that are not own properties of `item` are omitted from the result, so
 * the returned object never carries `undefined` placeholders for missing keys.
 * Matches the semantics of `lodash.pick`.
 *
 * @param {Object|null|undefined} item - The source object. `null` or `undefined` returns `{}`.
 * @param {string[]} keys - Keys to copy to the returned object.
 * @returns {Object} A new object containing only the selected own keys present on `item`.
 *
 * @example
 * pick({ id: 1, name: "Alice" }, ["id", "name", "role"]);
 * // => { id: 1, name: "Alice" }
 *
 * @example
 * pick(null, ["id"]);
 * // => {}
 */
export function pick(item, keys) {
  if (item == null) return {};

  return Object.fromEntries(
    keys
      .filter((key) => Object.prototype.hasOwnProperty.call(item, key))
      .map((key) => [key, item[key]]),
  );
}
```

joshsadam (changes requested):

app/javascript/utilities/collection.js
ksierks: Tests should reflect changes in 3ed2912dba126c709d2c90be1be2f1046be8c228.

app/javascript/controllers/linelist_import_controller.js
ksierks: 78d6c4c55312105b7fb46b9133b57a57bdfa1ba7

app/javascript/controllers/linelist_import_controller.js
ksierks: 58da430a2e0ccb8780c08bb9564895bb46591839

app/javascript/controllers/linelist_import_controller.js
ksierks: Sorry, are you asking me to update the server-side condition to `ignoreEmptyValues && (value === null || value === undefined || value === "")` to match the more aggressive client-side condition? Or update the client-side condition to be less aggressive like this, 595525b462cb94e3b3db8b75c3a17ce2ba12a5a2?

app/javascript/workers/linelist_import_worker.js
ksierks: Good catch! I copied this list from /irida-next/app/views/shared/samples/metadata/file_imports/_form_fields.html.erb.
97a5582de9ef3e2565a4ed02f283801876e0dd21

test/javascript/utilities/collection.test.js
```diff
@@ -0,0 +1,101 @@
...
+  describe("pick", () => {
+    it("returns a new object containing only the selected keys", () => {
+      const result = pick({ id: 1, name: "Alice", role: "admin" }, [
+        "name",
+        "role",
+      ]);
+      expect(result).toEqual({ name: "Alice", role: "admin" });
+    });
+
+    it("includes missing keys with undefined values", () => {
```
joshsadam: Test is still named "includes missing keys with undefined values" but the updated `pick` omits missing keys. Tests pass (likely because `toEqual` treats { id: 1 } as matching { id: 1, name: undefined }), but the test name is misleading.

joshsadam (commented):
Getting an error when testing this out:

Image

Disregard this, as it was not a problem with this PR.

test/javascript/utilities/collection.test.js
ksierks: b2fdd5ae06e92243eda8f706643ae44abd25d833

app/components/data_imports/linelist_import_dialog/v2/component.html.erb
ksierks: I'm trying not to modify /irida-next/app/views/shared/samples/metadata/file_imports/_form_fields.html.erb as it's shared and I don't want to break the other component.
The `handleSubmit()` now has `event.preventDefault()` and `event.stopPropagation()` to prevent the form submission. I removed closing the dialog on `submit-end` and reused the JavaScript function from the export line list controller in 237c1397e741689197f4181e71f51e0829994406.

joshsadam: The enter button still does not submit the form.   To get it to work, add `submit->linelist-import#handleSubmit` to the form in `app/components/data_imports/linelist_import_dialog/v2/component.html.erb`

ksierks: The enter key worked for me, but here's the change 765f956d4a262c9817a401eea86ce96cf3659c31.

joshsadam (approved):

ChrisHuynh333 (approved):

PR #1883, opened 2026-05-13
Combobox Datepicker: Replace PAT datepicker and implement max_date logic

ChrisHuynh333:
This PR changes the following:
- Switches the old datepicker for PATs to use the new `combobox_datepicker`. With this, `max_date` logic has now been added to `combobox_datepicker`.
- Removes the footer notes about min/max dates
- Changes today's date styling to be a dot, and is visible when today's date is disabled
- Some shared logic between `min_date` and `max_date` has also been refactored to prevent duplicate logic.

Max date:
image

Today's date:
image

Test max date:
1. In `rails c`, run: `ApplicationSetting.current.update(require_personal_access_token_expiry: true)`
2. Navigate to a project or group with at least `maintainer` access level, and create a new bot.
3. Test the datepicker with the `max_date` boundary (clicking and keyboard). This includes making sure you can't navigate past the max date with month/year changes, and keyboard changes (end, pageDown, shift+pageDown, ArrowDown/Right).

Test today's date:
1. In the datepicker preview with no min date, verify today's date styling (re-)appears when opening the datepicker and changing months forward/back, and when you select today's date, the today's date dot styling is overridden by the selected date styling

## PR acceptance checklist


ericenns (commented):
This is looking pretty good, but I found a couple issues.

Firstly, min and max date messages are not associated with an aria-describedby so the screen reader doesn't announce them. I'm wondering if we just don't include these messages as the date picker itself prevents selection of dates. It would keep the UI simple if we didn't have these messages and other accessible examples that I have found do not have messaging for min and max date.

Secondly, the current selected date has a contrast issue (this might have been existing), see screenshot (maybe change text to white which would fix contrast issue).

Image

ChrisHuynh333:
> This is looking pretty good, but I found a couple issues.
> 
> Firstly, min and max date messages are not associated with an aria-describedby so the screen reader doesn't announce them. I'm wondering if we just don't include these messages as the date picker itself prevents selection of dates. It would keep the UI simple if we didn't have these messages and other accessible examples that I have found do not have messaging for min and max date.
> 
> Secondly, the current selected date has a contrast issue (this might have been existing), see screenshot (maybe change text to white which would fix contrast issue).
> 
> Image

Fixed styling and removed min/max date footnotes in b01866a801568b8c6d7d4484a7ca26af0351be00

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -46,6 +48,8 @@ export default class extends Controller {
   #selectedYear;
   #selectedMonthIndex;
 
+  #maxDate;
```
joshsadam, line 51: We probably need to add `this.#maxDate` validation on typed input around line 284. `directInput()` only rejects dates below `minDate`; it never checks `#maxDate` / `invalidMaxDateValue`. Legacy datepicker v1 does (max check first, then min). PAT now uses combobox_datepicker with max_date, so a user can type a date past the lifetime cap, get a green field, and only fail on submit. Mirror datepicker v1: `if (this.#maxDate && dateInput > this.#maxDate) { ... invalidMaxDateValue }`before the min branch.

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
```diff
@@ -344,18 +366,25 @@ export default class extends Controller {
...
+    if (minDate || maxDate) {
       // get all the date nodes within current calendar, and all dates prior the minDate index will be disabled
       const allDates = Array.from(
         this.calendarTarget.querySelectorAll("[data-date]"),
       );
-      for (let i = 0; i < allDates.indexOf(minDate); i++) {
-        this.#replaceDateStyling(
-          allDates[i],
-          CALENDAR_CLASSES["DISABLED_DATE"],
+      // disable dates from start of calendar to minDate
```
joshsadam, line 377: `aria-current` on today when date is disabled. 

Today's dot styling is applied whenever today != selected, without skipping `aria-disabled` cells. That matches the PR goal ("visible even when disabled") but sets `aria-current="true"` on a non-selectable day, which can mislead assistive tech. Consider removing `aria-current` (or the dot) when today is disabled, while keeping a purely visual indicator if needed.

joshsadam (changes requested):
This looks really good, just 2 small items I found

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: Going to leave this. My follow-up PR to this current PR was actually to remove the remaining front-end validation (used by members and group links tables), which doesn't use maxDate.

However in testing this, I found a bug where if a user submitted an invalid date, that date would render onto the calendar (eg: if I submitted 2022-01-01, which is below minDate, the calendar would then show Jan 2022). This was fixed in eab885b9c37af6e89efaea6676f54868f9edaaf5 15e0a2d50a7f3ecd81a2e675a51facf1d87d536f

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ChrisHuynh333: Discussed with Eric and he's okay keeping dot styling and aria-current on today's date even when disabled

joshsadam (approved):

ericenns (approved):
Looks good to me!

PR #1894, opened 2026-05-14
Group transfer form object with form error summary

ksierks:
This PR adds a `form_error_summary` to the transfer section of the group general settings page. A transfer form object was also created to be used by the`form_error_summary`. 

Summary of changes:
* `app/forms/transfer_form.rb` is a form object used to encapsulate a group transfer request, it includes validation for transfer params.
* `app/controllers/groups_controller.rb` has been updated to use this new form object for `edit` and `transfer` endpoints.
* `app/views/groups/` views have been updated to use the new form object and to use the `form_error_summary` helper
* `app/services/groups/transfer_service.rb` has been updated to require a group transfer_form. In addition validations have been moved out of the service and into the form object.

image

To test the following error messages:
  - **blank** 
    Replace the following line within the `transfer` action of the groups_controller.rb:
  `@transfer_form = ::Groups::TransferForm.new(group_transfer_params.merge(group: @group))`
  with
  `@transfer_form = ::Groups::TransferForm.new({ new_namespace_id: nil}.merge(group: @group))`
  - **namespace_not_found**
    Create new groups `Group1` & `Group2`. Try transferring `Group1` to `Group2`, but before submitting delete `Group2`. 
  - **namespace_group_exists**
    Create `Group1` -> `SubgroupA` and `Group2`->`SubgroupA`. Try transferring `SubgroupA` to the other group. 
  - **other_than** 
    Comment out the following line from the `edit` action of the groups_controller.rb:
  `@authorized_namespaces -= [@group]`
    Try transferring the same group.

## PR acceptance checklist


app/forms/transfer_form.rb
ericenns, line 1: nitpick (blocking): Can you namespace this under a `Groups` module, since I anticipate us having more transfer forms and this makes it clearer that this one is specific to `Groups`.

app/views/groups/_edit_advanced_transfer_form.html.erb
```diff
@@ -7,9 +7,13 @@ if Flipper.enabled?(:v2_select2, Current.user)
   end) do |form| %>
   <%= render partial: "shared/form/required_field_legend" %>
 
-  <div class="form-field">
+  <% invalid_new_namespace_id = @transfer_form.errors.include?(:new_namespace_id) %>
+
+  <div class="form-field <%= 'invalid' if invalid_new_namespace_id %>">
     <% form_id = "transfer_namespace_id_input" %>
 
+    <%= form_error_summary(form, target_overrides: { new_namespace_id: form_id }) %>
```
ericenns: nitpick (blocking): Now that we have a form object you can remove passing `target_overrides` and setting `form_id`.

Instead remove `for:` from `form.label :new_namespace_id` call, and in the `render Select2Component` call pass `form.field_id(:new_namespace_id)`

config/locales/en.yml
```diff
@@ -14,6 +14,8 @@ en:
         value: Value
       sample/search_group:
         condition: Condition
+      transfer_form:
+        new_namespace_id: New namespace
```
ericenns: I think we should rename this to `New parent group` to match the form label better.

app/forms/transfer_form.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# Form object for transferring groups
+class TransferForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Validations::Callbacks
+
+  attribute :new_namespace_id, :string
```
ericenns: nitpick(blocking): Can we rename this to `new_parent_id` since that matches the attribute that we are updating better e.g. `parent_id`.

app/forms/transfer_form.rb
```diff
@@ -0,0 +1,33 @@
...
+  attribute :group_id, :string
```
ericenns: nitpick (blocking): Instead of passing three attribtutes for the group, I think it would be better to just have a singular `group` attribute. Or add an additional `old_parent_id` attribute which can be used to confirm that the `new_parent_id` is not equal to the `old_parent_id` which can have a specific error message like: `is the same as existing parent group`

app/forms/transfer_form.rb
```diff
@@ -0,0 +1,33 @@
...
+    new_namespace_id.present? && group_name.present? && group_path.present?
+  }
+
+  def initialize(attributes = {})
+    super
+  end
+
+  private
+
+  def group_exists_in_namespace
```
ericenns: ntipick(blocking): can you rename this to `group_does_not_conflict_with_existing_group` since that matches the behaviour of the method. Please also update the translation key to match new name.

app/controllers/groups_controller.rb
```diff
@@ -116,15 +118,16 @@ def destroy
   end
 
   def transfer # rubocop:disable Metrics/AbcSize
-    new_namespace ||= Namespace.find_by(id: params.require(:new_namespace_id))
+    @transfer_form = ::TransferForm.new(group_transfer_params.merge(group_id: @group.id, group_name: @group.name,
+                                                                    group_path: @group.path))
+    new_namespace ||= Namespace.find_by(id: @transfer_form.new_namespace_id)
```
ericenns: nitpick(blocking): This can be removed along with passing `new_namespace` to `execute` for `Groups::TransferService`, instead add a method to the transfer form called `new_parent` which finds the Group with id matching `new_parent_id`. Should also add validationg to the transfer form to ensure that a Group with `new_parent_id` exists.

ericenns (changes requested):
Looking pretty good, but left some comments on how this can be improved.

app/forms/transfer_form.rb
ksierks: 31b92f695230fa06a23eefca7898cc46fdfb3c4d

app/views/groups/_edit_advanced_transfer_form.html.erb
ksierks: 54ed010bcd11c9549898ee9214b94b9e7c68d505

config/locales/en.yml
ksierks: b8d7646745003040396b9e6105f2392dae1a0e65

app/forms/transfer_form.rb
ksierks: 58e9db552a3250d4105441d7d166b1865ddaae3c

app/forms/transfer_form.rb
ksierks: 962c1ee709875cb2528e97e2087c9a739d3b0281

app/controllers/groups_controller.rb
ksierks: d417c13a74a3f59625bb3a5214e5ea85ed5dcf1d

app/forms/transfer_form.rb
ksierks: d169e1c77dc0c6ccb28162dd9d9f2c67a299bdac & ad5bb2331d0317e0f7481ba6038a264c7aff492e
I added a new validation that compares the `new_parent_id` to the `old_parent_id` like you asked, but I think the `group_does_not_conflict_with_existing_group` validation covers this case already.
I'm also not sure what attribute type you'd like me to use on group.

ericenns: The existing validation did cover it, but it would not have an appropriate error message if the old parent id was equal to the new parent id.

If `new_parent_id` is present and equal to `old_parent_id` we should skip `group_does_not_conflict_with_existing_group` validation so that we don't add an error message when the one for `old_parent_id` matching `new_parent_id` would already cover this.

ksierks: I've renamed and updated `group_does_not_conflict_with_existing_group` to only return one error message. 62a89cc0d4b2e5bfd40920efaedd19012af97bab

app/services/groups/transfer_service.rb
```diff
@@ -3,11 +3,15 @@
...
+      super(group, user)
+      @transfer_form = transfer_form
     end
 
-    def execute(new_namespace) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
-      validate(new_namespace)
+    def execute # rubocop:disable Metrics/AbcSize, Naming/PredicateMethod
+      return false unless @transfer_form.valid?
+
+      new_namespace ||= @transfer_form.new_parent
```
ericenns: nitpick(blocking): This can be changed from `||=` to `=` as we don't set `new_namespace` prior to this call so no need for conditional assignment.

config/locales/en.yml
ericenns, line 1: nitpick(blocking): Please add translation for `new_parent_id` within `activemodel.attributes.group/transfer_form` namespace so that error messages are translated.

config/locales/fr.yml
ericenns, line 1: nitpick(blocking): Please add translation for `new_parent_id` within `activemodel.attributes.group/transfer_form` namespace so that error messages are translated.

Image

ericenns (changes requested):
This is looking good, just a couple comments regarding variable assignment and missing translations.

app/services/groups/transfer_service.rb
ksierks: 152237f0661ae2854901b3e274f750dc0f778a30

config/locales/en.yml
ksierks: 152237f0661ae2854901b3e274f750dc0f778a30

config/locales/fr.yml
ksierks: 152237f0661ae2854901b3e274f750dc0f778a30

ericenns (approved):

PR #1911, opened 2026-05-20
Bug: Dropdown caret arrow rotation

ChrisHuynh333:
This PR fixes a bug for the dropdown caret rotation where the rotation did not happen in non-dev environments. This is due to when setting the caret in JS via `querySelector`, it was using `svg.caret-down-icon`  class, which isn't applied in non-dev environments. This has now been fixed and if caret exists for the dropdown, the `querySelector` will just look for `svg`

1. Verify dropdowns with the caret icon still rotates in both v1 and v2

## PR acceptance checklist

### DIFF
--- a/app/components/dropdown/v1/component.html.erb
+++ b/app/components/dropdown/v1/component.html.erb
@@ -3,7 +3,6 @@
   data-controller="dropdown--v1<%= tooltip? ? ' pathogen--tooltip' : '' %>"
   data-dropdown--v1-skidding-value="<%= skidding %>"
   data-dropdown--v1-distance-value="<%= distance %>"
-  data-dropdown--v1-caret-value="<%= caret %>"
 >
   <%= render Viral::BaseComponent.new(tag: :button, **system_arguments) do %>
     <% if icon_name.present? %>
--- a/app/components/dropdown/v1/component.html.erb
+++ b/app/components/dropdown/v1/component.html.erb
@@ -25,7 +25,9 @@
     <% end %>
 
     <% if caret %>
-      <%= icon(:caret_down, size: :sm, class: "ml-2 flex-none") %>
+      <span data-dropdown--v1-target="caret" class="ml-2 inline-flex flex-none">
+        <%= icon(:caret_down, size: :sm) %>
+      </span>
     <% end %>
   <% end %>
 
--- a/app/javascript/controllers/dropdown/v1_controller.js
+++ b/app/javascript/controllers/dropdown/v1_controller.js
@@ -1,7 +1,7 @@
 import { Controller } from "@hotwired/stimulus";
 
 export default class extends Controller {
-  static targets = ["trigger", "menu"];
+  static targets = ["trigger", "menu", "caret"];
   static values = {
     position: String,
     trigger: String,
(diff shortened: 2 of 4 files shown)


app/javascript/controllers/dropdown/v1_controller.js
```diff
@@ -22,7 +22,7 @@ export default class extends Controller {
   connect() {
     this.element.setAttribute("data-controller-connected", "true");
     if (this.caretValue) {
-      this.#caret = this.element.querySelector("svg.caret-down-icon");
+      this.#caret = this.element.querySelector("svg");
```
joshsadam: Long-term, this is brittle. Let's update this to use a stimulus target.

In the dropdown templates:
```html
<% if caret %>
  <span data-dropdown--v1-target="caret" class="ml-2 flex-none inline-flex">
    <%= icon(:caret_down, size: :sm) %>
  </span>
<% end %>
```

In the controller:
```javascript
static targets = ["trigger", "menu", "caret"];
// connect / onShow / onHide
if (this.hasCaretTarget) {
  this.caretTarget.classList.toggle("rotate-180", expanded);
}
```

ChrisHuynh333: Updated to use `classList.add/remove`, seems toggle might overcomplicate things needing to evaluate `aria-expanded` each time? I can do that though if you think it's better 7d52931625a089f0c73fa291fdb979091b1a11f1

app/javascript/controllers/dropdown/v1_controller.js
```diff
@@ -10,8 +10,6 @@ export default class extends Controller {
     caret: Boolean,
```
joshsadam: Suggestion: Now that we toggle rotation with `hasCaretTarget` / `caretTarget`, the `caret: Boolean` Stimulus values in v1 and v2 look unused. Could we remove them from `static values` so we are not carrying dead config?

app/components/dropdown/v1/component.html.erb
```diff
@@ -25,7 +25,9 @@
     <% end %>
```
joshsadam, line 25: Suggestion: `data-dropdown--v1-caret-value` (and the v2 `dropdown--v2-caret-value` equivalent) was only needed when JS gated caret lookup on `caretValue`. With the caret Stimulus target, `hasCaretTarget` is enough. Could we drop these data attributes from both templates?

joshsadam (changes requested):
Nice fix. I added a couple of suggestions to drop leftover `caret-value` wiring now that we rely on `hasCaretTarget`. Looks good once those are tidied up.

app/javascript/controllers/dropdown/v1_controller.js
ChrisHuynh333: Thanks, cleaned up in 7cb5ec2d13ac866d28e4c215cdce00b31da4b253

app/components/dropdown/v1/component.html.erb
ChrisHuynh333: 7cb5ec2d13ac866d28e4c215cdce00b31da4b253

joshsadam (approved):

PR #1925, opened 2026-05-26
Form error summaries for edit workflow

ksierks:
This PR adds a `form_error_summary` to the edit workflow execution dialogs.

image

1. Navigate to edit a workflow exection.
2. Enter a blank name.
3. Verify a form error summary is displayed.

## PR acceptance checklist

### DIFF
--- a/app/models/workflow_execution.rb
+++ b/app/models/workflow_execution.rb
@@ -28,7 +28,7 @@ class WorkflowExecution < ApplicationRecord # rubocop:disable Metrics/ClassLengt
   validates :metadata, presence: true, json: { message: ->(errors) { errors }, schema: METADATA_JSON_SCHEMA }
   validate :validate_namespace, on: :create
   validate :validate_workflow_available, if: :initial?
-  validates :name, presence: true, if: -> { !submitter&.automation_bot? }
+  validates :name, presence: true
 
   enum :state,
        { initial: 0, prepared: 1, submitted: 2, running: 3, completing: 4, completed: 5, error: 6, canceling: 7,
--- a/app/views/workflow_executions/_edit_dialog.html.erb
+++ b/app/views/workflow_executions/_edit_dialog.html.erb
@@ -11,12 +11,14 @@
   </div>
 
   <%= form_with(model: workflow_execution, url: workflow_execution_path, method: :patch, class: "grid gap-4") do |form| %>
+    <%= render partial: "shared/form/required_field_legend" %>
+
     <%= form_error_summary(form) %>
 
     <% invalid_name = @workflow_execution.errors.include?(:name) %>
 
     <div class="form-field <%= 'invalid' if invalid_name %>">
-      <%= form.label :name %>
+      <%= form.label :name, data: { required: true } %>
 
       <%= form.text_field :name,
                       autofocus: true,
--- a/app/views/workflow_executions/_edit_dialog.html.erb
+++ b/app/views/workflow_executions/_edit_dialog.html.erb
@@ -21,14 +21,15 @@
       <%= form.label :name, data: { required: true } %>
 
       <%= form.text_field :name,
-                      autofocus: true,
                       value: @workflow_execution.name,
                       placeholder:
                         t("workflow_executions.edit_dialog.name_placeholder") %>
 
-      <%= render "shared/form/field_errors",
+      <% if invalid_name %>
+        <%= render "shared/form/field_errors",
               id: form.field_id(:name, "error"),
               errors: @workflow_execution.errors.full_messages_for(:name) %>
+      <% end %>
     </div>
 
     <div class="flex h-5 items-center">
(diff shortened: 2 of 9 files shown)


app/models/workflow_execution.rb
```diff
@@ -28,7 +28,7 @@ class WorkflowExecution < ApplicationRecord # rubocop:disable Metrics/ClassLengt
   validates :metadata, presence: true, json: { message: ->(errors) { errors }, schema: METADATA_JSON_SCHEMA }
   validate :validate_namespace, on: :create
   validate :validate_workflow_available, if: :initial?
-  validates :name, presence: true, if: -> { !submitter&.automation_bot? }
+  validates :name, presence: true
```
joshsadam, line 31: **Question:** Was dropping `if: -> { !submitter&.automation_bot? }` intentional? The edit UI now treats name as required (legend, placeholder), so tightening validation may be deliberate.

On `main`, workflow executions with an automation-bot submitter could still be saved with a blank `name` (e.g. cleared via edit). Any such rows—or future code that saves without a name—would start failing on unrelated `save` calls (status/cancel jobs, etc.). Launch still sets a name for automated runs, so we may be fine, but this is a wider contract change than the dialog work alone.

Could we confirm product intent? If edit-only was the goal, we could keep the conditional or scope `presence` to `:update` and add a model test for automation-bot + blank `name`.

app/views/workflow_executions/_edit_dialog.html.erb
```diff
@@ -9,24 +10,31 @@
...
+
+    <%= form_error_summary(form) %>
+
+    <% invalid_name = @workflow_execution.errors.include?(:name) %>
+
+    <div class="form-field <%= 'invalid' if invalid_name %>">
+      <%= form.label :name, data: { required: true } %>
+
       <%= form.text_field :name,
                       autofocus: true,
```
joshsadam: **Suggestion:** `FormErrorSummaryHelper` notes we should remove autofocus on fields covered by the summary so `form-error-summary` + `focusWhenVisible` can receive focus after a failed submit; `autofocus: true` on the name field wins that race.

For first open without errors, we could mirror `devise/registrations/edit.html.erb` (`autofocus: resource.errors.none?`) or drop autofocus and rely on dialog/tab order. Same change in `groups/` and `projects/` `_edit_dialog` partials.

app/views/workflow_executions/_edit_dialog.html.erb
```diff
@@ -9,24 +10,31 @@
...
                       value: @workflow_execution.name,
                       placeholder:
                         t("workflow_executions.edit_dialog.name_placeholder") %>
 
+      <%= render "shared/form/field_errors",
```
joshsadam: **Nit:** Could we wrap `field_errors` in `<% if invalid_name %>` like `projects/samples/_form_fields`? That avoids an empty `<ul>` when there are no name errors. Same in the groups/projects partials.

joshsadam (changes requested):

app/models/workflow_execution.rb
ksierks: I asked @ericenns about dropping this condition. The intent is to treat name as required all the time. I may be wrong, but I thought if you launch or edit a workflow execution, a name is required. And automation bots always provide a name for workflow executions. Is it possible to confirm by querying a live database?

app/views/workflow_executions/_edit_dialog.html.erb
ksierks: b107012ac453210138f5792fb83cba6b799a233f

app/views/workflow_executions/_edit_dialog.html.erb
ksierks: I forgot. b107012ac453210138f5792fb83cba6b799a233f

app/views/workflow_executions/_edit_dialog.html.erb
joshsadam: This needs to be done in all 3 `_edit_dialog.html.erb`

app/views/workflow_executions/_edit_dialog.html.erb
joshsadam: This needs to be done in all 3 `_edit_dialog.html.erb`

app/views/workflow_executions/_edit_dialog.html.erb
ksierks: 1da3972178b07d8ddcad0151e13ab0f5e4147dff

app/views/workflow_executions/_edit_dialog.html.erb
ksierks: 1da3972178b07d8ddcad0151e13ab0f5e4147dff

ericenns (approved):
Looks good to me!

joshsadam (approved):

PR #1937, opened 2026-05-27
Refactor TransferService/TransferJob to be Idempotent

JeffreyThiessen:

This PR changes the flow of Samples::TransferJob and related service so that it can be recovered and continued in the event of a job failure. 

To do this, it has been broken into ordered steps to
1. transfer samples
2. update metadata
3. update sample counts and generate activities
4. collect errors and broadcast turbo stream

To make this process recoverable, an extra database call must be made to fetch the list of samples that has been transferred. 
Extra metadata is added to the Logidze data, 'transfer_job_id` and `previous_project_id`. These allow us to determine the source of a sample that has already been transferred, which is needed for error reporting.

The new code flow prevents the edge case where a sample is attempted to be transferred from a different project than it originates from. 

A new index for the `samples` table has been added. 
`((log_data -> 'h' -> -1 -> 'm' ->> 'transfer_job_id'))"`
This allows us to more efficiently fetch the latest log data from samples.

The testing suite for transferring samples is quite extensive, and additional tests have been added to test the containable process. 

Transfer some samples manually through the UI and verify that the progress bar advances correctly.

## PR acceptance checklist


app/jobs/samples/transfer_job.rb
```diff
@@ -2,34 +2,100 @@
...
+      @broadcast_target = broadcast_target
+      @service = Samples::TransferService.new(@namespace, @current_user)
 
-      if namespace.errors.empty?
+      @authorization_error = false
+
+      pre_transfer_check
+      @service.update_progress_bar(5, 100, broadcast_target)
+
+      step :transfer_step, start: 0
```
deepsidhu85: If we move these into their own method such as `run_steps`  this should remove the rubocop disable for the method

config/locales/en.yml
```diff
@@ -1972,6 +1972,7 @@ en:
       transfer:
         empty_new_project_id: Destination project was not selected.
         empty_sample_ids: The sample ids are empty.
+        invalid_new_project: The selected destination project is invalid.
```
deepsidhu85: I think we should change this to `The destination project does not exist`

app/services/samples/transfer_service.rb
```diff
@@ -270,63 +199,42 @@ def build_transferred_project_sample_ids(project_sample_ids_to_transfer,
...
-            # target project and prevented the transfer.
-            @namespace.errors.add(:samples,
-                                  I18n.t('services.samples.transfer.sample_exists', sample_name: name,
-                                                                                    sample_puid: puid))
-          end
+    def add_transfer_errors(all_sample_ids, transferred_sample_ids, new_project_id) # rubocop:disable Metrics/MethodLength
+      untransferred_sample_ids = all_sample_ids - transferred_sample_ids
+      return if untransferred_sample_ids.empty?
+
+      filtered_samples = filter_sample_ids(untransferred_sample_ids, 'transfer', true)
```
deepsidhu85: As the `filter_sample_ids` method already has `append_namespace_errors` set to `true`, we can remove `true` from this call

deepsidhu85 (changes requested):
This is looking good, just a few comments to start

app/services/samples/transfer_service.rb
JeffreyThiessen: Good catch, updated in https://github.com/phac-nml/irida-next/pull/1937/commits/9b0d9fd59c814b34770596c890126763237cd70d

app/jobs/samples/transfer_job.rb
JeffreyThiessen: Fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/9b0d9fd59c814b34770596c890126763237cd70d

config/locales/en.yml
JeffreyThiessen: fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/69ef8a715f9d0c985bf7328c40656e72e8b2f069

app/jobs/samples/transfer_job.rb
```diff
@@ -2,34 +2,104 @@
...
+    def perform(namespace, current_user, new_project_id, sample_ids, broadcast_target = nil)
+      @namespace = namespace
+      @current_user = current_user
+      @new_project_id = new_project_id
+      @sample_ids = sample_ids
+      @broadcast_target = broadcast_target
+      @service = Samples::TransferService.new(@namespace, @current_user)
 
-      if namespace.errors.empty?
+      @authorization_error = false
```
deepsidhu85: I think we should change this variable. Reason being is that `authorization_error`  implies that an error related to authorization was encountered, but `rescue BaseSampleService::BaseError, TransferService::TransferError => e`  has multiple other errors that sets the `@authorization_error` variable such as the new project isn't found

deepsidhu85 (changes requested):
This looks good just one more small comment

app/jobs/samples/transfer_job.rb
JeffreyThiessen: Fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/6f09dc6713b70cad817b404db1dddfea4fe0299e

test/services/samples/transfer_service_test.rb
```diff
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
```
ericenns, line 1: nitpick(blocking): Can you move the job specific tests back into `test/jobs/samples/transfer_job_test.rb` and have this file just contain tests for the public methods in the service that are called by the job. To ensure they do what they are supposed to.

app/jobs/samples/transfer_job.rb
```diff
@@ -2,34 +2,104 @@
...
+      @new_project_id = new_project_id
+      @sample_ids = sample_ids
+      @broadcast_target = broadcast_target
+      @service = Samples::TransferService.new(@namespace, @current_user)
 
-      if namespace.errors.empty?
+      @pre_transfer_error = false
+
+      pre_transfer_check
+      @service.update_progress_bar(5, 100, @broadcast_target)
```
ericenns: nitpick(blocking): This progress update should be performed with a step, so that when the job is interrupted and re-runs it doesn't set the progress back down to 5 percent and then jump back up.

app/jobs/samples/transfer_job.rb
```diff
@@ -2,34 +2,104 @@
...
+      run_job_steps
+
+      return_data
+    end
+
+    private
+
+    def run_job_steps
+      step :transfer_step, start: 0
+      @service.update_progress_bar(95, 100, @broadcast_target)
```
ericenns: nitpick(blocking): Progress bar updates should be performed within a step so that they are not re-ran when job is interrupted and continues execution.

app/services/samples/transfer_service_v2.rb
```diff
@@ -19,37 +19,23 @@ class TransferService < BaseSampleService # rubocop:disable Metrics/ClassLength
...
-    # Performs authorization checks, validation, and orchestrates the transfer process.
-    # For maintainers, also validates that source and target projects share a common ancestor.
-    #
-    # @param new_project_id [Integer] the ID of the target project
-    # @param sample_ids [Array<Integer>] IDs of samples to transfer
-    # @param broadcast_target [String, nil] optional Turbo broadcast target for progress updates
-    #
-    # @return [Array<Integer>] IDs of successfully transferred samples
-    # @raise [BaseSampleService::BaseError] on validation or authorization failures
-    def execute(new_project_id, sample_ids, broadcast_target = nil)
```
ericenns, line 33: nitpick(blocking): Can you add back in an execute method that performs all the steps (So calls the methods you have broken up). Which we can use for graphql for now, so that we are not calling the job there.

test/services/projects/samples/destroy_service_test.rb
```diff
@@ -107,7 +107,7 @@ def setup
         assert_equal({ 'metadatafield1' => 1, 'metadatafield2' => 1 }, @subgroup12b.metadata_summary)
         assert_equal({ 'metadatafield1' => 3, 'metadatafield2' => 3 }, @group12.metadata_summary)
 
-        ::Samples::TransferService.new(project30.namespace, @user).execute(@project31.id, [sample33.id])
+        ::Samples::TransferJob.perform_now(project30.namespace, @user, @project31.id, [sample33.id])
```
ericenns: nitpick(blocking): Can we just remove the call to the TransferService and refactor the test so that this is not needed.

ericenns (changes requested):
Looks like a good start, but I found a couple issues.

ericenns (changes requested):
Since this is quite a substantial change in behaviour, can you instead make versioned TransferJob and TransferService (e.g. current would be v1 an updated would be v2) and only use v2 if a feature flag is set e.q. `samples_transfer_v2` that way if we start encountering issues after deploy we can roll back easily.

app/jobs/samples/transfer_job.rb
JeffreyThiessen: fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/fb78f8e26ca3abc6c2812ba426a0584977db5d05

app/jobs/samples/transfer_job.rb
JeffreyThiessen: fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/fb78f8e26ca3abc6c2812ba426a0584977db5d05

test/services/projects/samples/destroy_service_test.rb
JeffreyThiessen: fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/dd4a1b5a438d3f7a2481963efe72574af5d00190

app/services/samples/transfer_service_v2.rb
JeffreyThiessen: I have put the original code flow back which the graphql mutation now uses again. All the new code is within the V2 service/job

test/services/samples/transfer_service_test.rb
JeffreyThiessen: The original test cases for the original service/job have been returned, and new files have been added for the v2 tests. These have been restructured and there are now additional tests for the service methods.

test/controllers/samples/transfers_controller_test.rb
```diff
@@ -47,6 +47,21 @@ class TransfersControllerTest < ActionDispatch::IntegrationTest
...
+      assert_enqueued_jobs 1, only: ::Samples::TransferJobV2 do
+        post samples_transfer_path(namespace_id: @namespace.id, format: :turbo_stream),
+             params: {
+               transfer: {
+                 new_project_id: @project2.id,
+                 sample_ids: [@sample1.id, @sample2.id]
+               },
+               broadcast_target: 'a_broadcast_target'
+             }
+      end
```
ericenns, line 62: nitpick(blocking): At the end of the test can you ensure that the feature flag is disabled (you can see examples of this in  other tests within the codebase).

test/controllers/samples/transfers_controller_test.rb
```diff
@@ -72,5 +87,20 @@ class TransfersControllerTest < ActionDispatch::IntegrationTest
...
+      assert_enqueued_jobs 1, only: ::Samples::TransferJobV2 do
+        post samples_transfer_path(namespace_id: @project1.namespace.id, format: :turbo_stream),
+             params: {
+               transfer: {
+                 new_project_id: @project2.id,
+                 sample_ids: [@sample1.id, @sample2.id]
+               },
+               broadcast_target: 'a_broadcast_target'
+             }
+      end
```
ericenns, line 105: nitpick(blocking): Same comment as above.

app/graphql/mutations/transfer_samples.rb
```diff
@@ -78,6 +78,7 @@ def transfer_samples(project, new_project_id, sample_gids) # rubocop:disable Met
         next
       end
 
+      # TODO: this should switch to using a job, and return a job id that can be queried for the job status and results.
```
ericenns: nitpick(blocking): I think this TODO can be removed, if you want you can create an issue but you already have another PR to do this.

ericenns (changes requested):

test/controllers/samples/transfers_controller_test.rb
JeffreyThiessen: Fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/e595d3732667f614a9b9eab7659d3139fd3c3382

test/controllers/samples/transfers_controller_test.rb
JeffreyThiessen: All tests that enable the feature flag now disable it in an `ensure` block or a `teardown` block. https://github.com/phac-nml/irida-next/pull/1937/commits/e595d3732667f614a9b9eab7659d3139fd3c3382

app/graphql/mutations/transfer_samples.rb
JeffreyThiessen: Fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/e595d3732667f614a9b9eab7659d3139fd3c3382

app/controllers/samples/transfers_controller.rb
```diff
@@ -21,9 +21,16 @@ def create
       @broadcast_target = params[:broadcast_target]
       new_project_id = transfer_params[:new_project_id]
       sample_ids = transfer_params[:sample_ids]
-      Samples::TransferJob.set(wait_until: 1.second.from_now).perform_later(@namespace, current_user,
-                                                                            new_project_id,
-                                                                            sample_ids, @broadcast_target)
+
+      if Flipper.enabled?(:v2_sample_transfer)
```
deepsidhu85: This can be refactored to set the job to a variable

```
job = Samples::TransferJob
job = Samples::TransferJobV2 if Flipper.enabled?(:v2_sample_transfer)

job.set(....).perform_later(...)

```

app/services/samples/transfer_service_v2.rb
```diff
@@ -0,0 +1,337 @@
...
+      project_sample_ids_to_transfer
+    end
+
+    # Perform the actual sample transfer with database-level locking.
+    #
+    # Uses PostgreSQL advisory locks to prevent concurrent modifications and
+    # an outer join to detect name conflicts. Samples with conflicts are skipped.
+    #
+    # @param new_project [Project] the target project
+    # @param project_sample_ids_to_transfer [Hash{Integer => Array<Integer>}] samples to transfer
```
deepsidhu85, line 128: Update this method comment to contain the two new parameters

app/services/samples/transfer_service_v2.rb
```diff
@@ -0,0 +1,337 @@
...
+    end
+
+    # Organize samples by their source project.
+    #
+    # Groups sample IDs by project to allow processing samples from multiple
+    # source projects in a single transfer operation.
+    #
+    # @param transferrable_samples [ActiveRecord::Relation] samples to organize
+    # @return [Hash{Integer => Array<Integer>}] mapping from project ID to sample IDs
+    def organize_samples_by_project(transferrable_samples)
```
deepsidhu85, line 118: This method should be able to be simplified using a more idiomatic ruby approach


```suggestion
def organize_samples_by_project(transferrable_samples)
    transferrable_samples.group_by(&:project_id).transform_values { |samples| samples.map(&:id) }
end
```

app/services/samples/transfer_service_v2.rb
```diff
@@ -0,0 +1,337 @@
...
+    def find_transferred_samples_with_log_data_group_by_project(sample_ids, new_project_id, transfer_job_id)
+      transferred_samples = Sample.with_log_data.select(
+        "samples.*, #{LATEST_META} ->> 'previous_project_id' AS previous_project_id"
+      ).where(
+        id: sample_ids, project_id: new_project_id
+      ).where(
+        "#{LATEST_META} ->> 'transfer_job_id' = ?", transfer_job_id
+      )
+
+      return [] if transferred_samples.empty?
```
deepsidhu85: This method is returning an array or a hash depending on the condition, should we not just be returning one type of data structure? Since the `group_by` returns a hash, we should be able to remove the `return [] if transferred_samples.empty?` entirely and just chain `.group_by(&:previous_project_id)` to the final `where` above. This will return a empty hash if no results are found

deepsidhu85 (changes requested):
Just have some comments below to help clean up a few things

test/services/samples/transfer_service_v2_test.rb
```diff
@@ -0,0 +1,270 @@
+# frozen_string_literal: true
```
deepsidhu85, line 1: I think it may be worthwhile to add a few more tests particularly for:

1) Multiple transfers to the same destination project at the same time
2) If filtering the samples returns an empty data structure (don't believe we have a test for this edge case)

test/system/dashboard/projects_test.rb
```diff
@@ -292,6 +292,70 @@ def setup
       end
     end
 
+    test 'should update samples count after a sample transfer v2' do
```
deepsidhu85, line 295: This test is probably just better off as a service test rather than a system test

deepsidhu85 (changes requested):
Just a few more comments related to the tests

deepsidhu85:
Remember to also update the `features.yml` to contain the new feature flag `v2_sample_transfer`

app/controllers/samples/transfers_controller.rb
JeffreyThiessen: Fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/f32b807fa6e64ee0c11f292061d93a20839abfa4

app/services/samples/transfer_service_v2.rb
JeffreyThiessen: Fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/f32b807fa6e64ee0c11f292061d93a20839abfa4

app/services/samples/transfer_service_v2.rb
JeffreyThiessen: Fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/f32b807fa6e64ee0c11f292061d93a20839abfa4

app/services/samples/transfer_service_v2.rb
JeffreyThiessen: Fixed in https://github.com/phac-nml/irida-next/pull/1937/commits/f32b807fa6e64ee0c11f292061d93a20839abfa4

test/services/samples/transfer_service_v2_test.rb
JeffreyThiessen: Changes in https://github.com/phac-nml/irida-next/pull/1937/commits/2a353997a9697c0e3cb471fc7735a09484285cf8

For 1: A new test has been added
For 2: This was actually not a real edge case as the code could never reach that kind of state. This is because in the original v1 service the code flow was very different and the filtering was used differently. I have reworked the function to remove the unnecessary code path.

test/system/dashboard/projects_test.rb
JeffreyThiessen: The function is exactly in line with all of the other tests in this file that check the state of projects/samples after a service has been executed. If this test should be moved to a service test file I think that it should be done as a part of a separate test file refactor for all tests like this in the file.

deepsidhu85 (approved):
Looks good to me!

ericenns (approved):
Looks good to me!

PR #1982, opened 2026-06-12
Added UI to change visibility of parent group

deepsidhu85:

This PR adds in the UI to change a parent group's visibility between private and public.
- New card added for `Change group visibility` on the parent group General Settings page
- Descriptions and points added for changing from private to public, and vice versa
- Button added to launch confirmation window which has the user type the group path to confirm that they intended to make these changes

image

image

image

image

Subgroups and Projects don't have Change visibility action as it can only be done at the parent group level

image

image

1. Verify tests pass
2.  Log in as user0@email.com
3. Update the Bacillus group to `public` from the Bacillus -> General Settings page by clicking the `Change visibility` button and confirming
4. Log in as a user which doesn't have access to the group. Verify that you can see all the subgroups and projects with `GUEST` access level
5. Verify that you cannot update the visibility at the subgroup and project level
6. Logout and log in as user0@email.com
7. Update the Bacillus group back to `private` from the Bacillus -> General Settings page by clicking the `Change visibility` button and confirming
8. Log in as a user from step 4. Verify that you cannot see any of the subgroups and projects under Bacillus

## PR acceptance checklist

### DIFF
--- a/app/services/base_group_service.rb
+++ b/app/services/base_group_service.rb
@@ -13,12 +13,12 @@ class BaseGroupService < BaseService
     descendants_to_update = @group.self_and_descendants_of_type(
       [Group.sti_name,
        Namespaces::ProjectNamespace.sti_name]
-    ).where(public: false).sort_by { |ns| ns.full_path.length }
+    ).where(public: false).order('LENGTH(routes.path) ASC')
 
     descendants_to_update.each do |descendant|
       next if descendant.public == true
 
-      descendant.update(public: true)
+      descendant.update!(public: true)
 
       key = descendant.is_a?(Group) ? 'group.update' : 'namespaces_project_namespace.update'
 
@@ -31,7 +31,7 @@ class BaseGroupService < BaseService
     descendants_to_update = @group.self_and_descendants_of_type(
       [Group.sti_name,
        Namespaces::ProjectNamespace.sti_name]
-    ).where(public: true).sort_by { |ns| ns.full_path.length }
+    ).where(public: true).order('LENGTH(routes.path) ASC')
 
     descendants_to_update.each do |descendant|
       next if descendant.public == false
--- a/app/services/base_group_service.rb
+++ b/app/services/base_group_service.rb
@@ -13,12 +13,12 @@ class BaseGroupService < BaseService
     descendants_to_update = @group.self_and_descendants_of_type(
       [Group.sti_name,
        Namespaces::ProjectNamespace.sti_name]
-    ).where(public: false).order(type: :asc)
+    ).where(public: false).sort_by { |ns| ns.full_path.length }
 
     descendants_to_update.each do |descendant|
       next if descendant.public == true
 
-      descendant.update!(public: true)
+      descendant.update(public: true)
 
       key = descendant.is_a?(Group) ? 'group.update' : 'namespaces_project_namespace.update'
 
@@ -31,12 +31,12 @@ class BaseGroupService < BaseService
     descendants_to_update = @group.self_and_descendants_of_type(
       [Group.sti_name,
        Namespaces::ProjectNamespace.sti_name]
-    ).where(public: true).order(type: :asc)
+    ).where(public: true).sort_by { |ns| ns.full_path.length }
 
     descendants_to_update.each do |descendant|
       next if descendant.public == false
 
-      descendant.update(public: false)
+      descendant.update!(public: false)
 
       key = descendant.is_a?(Group) ? 'group.update' : 'namespaces_project_namespace.update'
 
--- a/app/services/groups/create_service.rb
+++ b/app/services/groups/create_service.rb
@@ -17,27 +17,28 @@ module Groups
         group.public = true
       end
 
-      group.save
+      ActiveRecord::Base.transaction do
+        group.save
 
-      if group.persisted?
-        @group.create_activity key: 'group.create',
-                               owner: current_user
-        if group.parent.nil?
-          Members::CreateService.new(current_user, group, {
-                                       user: current_user,
-                                       access_level: Member::AccessLevel::OWNER
-                                     }).execute
+        if group.persisted?
+          @group.create_activity key: 'group.create',
+                                 owner: current_user
+          if group.parent.nil?
+            Members::CreateService.new(current_user, group, {
+                                         user: current_user,
+                                         access_level: Member::AccessLevel::OWNER
+                                       }).execute
 
-        else
-          @group.parent.create_activity key: 'group.subgroups.create',
-                                        owner: current_user,
-                                        parameters: {
-                                          created_group_id: @group.id,
-                                          created_group_puid: @group.puid,
-                                          action: 'group_subgroup_create'
-                                        }
+          else
+            @group.parent.create_activity key: 'group.subgroups.create',
+                                          owner: current_user,
+                                          parameters: {
+                                            created_group_id: @group.id,
+                                            created_group_puid: @group.puid,
+                                            action: 'group_subgroup_create'
+                                          }
+          end
         end
-
       end
 
       group
--- a/app/services/groups/create_service.rb
+++ b/app/services/groups/create_service.rb
@@ -13,7 +13,9 @@ module Groups
     def execute # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       authorize! group.parent, to: :create_subgroup? if params[:parent_id].present?
 
-      group.public = true if params[:parent_id].present? && group.parent.public?
+      if Flipper.enabled?(:global_groups, current_user) && params[:parent_id].present? && group.parent.public?
+        group.public = true
+      end
 
       group.save
 
--- a/app/services/groups/transfer_service.rb
+++ b/app/services/groups/transfer_service.rb
@@ -29,10 +29,12 @@ module Groups
 
       @group.update(parameters)
 
-      if parameters[:public] == true
-        update_descendants_to_public
-      elsif parameters[:public] == false
-        update_descendants_to_private
+      if Flipper.enabled?(:global_groups, current_user)
+        if parameters[:public] == true
+          update_descendants_to_public
+        elsif parameters[:public] == false
+          update_descendants_to_private
+        end
       end
 
       create_activities(old_namespace, new_namespace)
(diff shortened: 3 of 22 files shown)


app/services/groups/update_service.rb
```diff
@@ -9,21 +9,24 @@ def initialize(group, user = nil, params = {})
       super
     end
 
-    def execute # rubocop:disable Metrics/AbcSize
-      authorize! @group, to: :update?
+    def execute # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+      authorize! @group, to: :update? unless params.key?(:public)
+      authorize! @group, to: :change_visibility? if (params.to_unsafe_h.size == 1) && params.key?(:public)
```
joshsadam: Issue: I think the permission check here might be wrong.

Right now we skip `update?` if the params include `public`. But we only run `change_visibility?` when `public` is the only param.

What if someone sends `public` and `name` together? We skip both checks and still save. A Maintainer could change visibility even though only Owners should.

Maybe we could do:
- if params has `public` → check `change_visibility?`
- otherwise → check `update?`

Also, we only show this in the UI for parent groups. Should the service also stop early if `@group.parent` is set? That way nobody could change a subgroup through the API.

app/views/groups/_edit_change_visibility.html.erb
```diff
@@ -0,0 +1,59 @@
...
+        <% end %>
+      <% else %>
+        <% t("groups.edit.advanced.change_visibility.public.points_html").each do |item| %>
+          <li><%= item %></li>
+        <% end %>
+      <% end %>
+    </ul>
+
+    <div class="mt-4 grid gap-4">
+      <%= button_to t("groups.edit.advanced.change_visibility.submit"),
```
joshsadam, line 47: Issue: Two things on this button.

1. `method: :patch` should go on `button_to` itself (like the delete button on this page). Right now it's inside `data:` and I don't think that works the same way.

2. We use `@group` for the URL and the confirm text. If the user just tried to save a blank path, `@group.path` is empty in memory. The other forms here use `group_path` from the current URL so they still work after a failed save.

Maybe something like:
<%= button_to ..., group_path, method: :patch, ... %>

config/locales/en.yml
```diff
@@ -867,6 +867,26 @@ en:
...
+              warning_html: You are going to change the visibility of <strong>%{group_name}</strong> to <strong>public</strong>. Once you confirm, all subgroups and projects under this group will also be publicly visible to all users of the application.
+          description:
+            private: This group is currently private
+            public: This group is currently public
+          private:
+            points_html:
+            - Changing a group to private will change all descendants (groups and projects) to private. Only users with explicit membership will be able to access.
+          public:
+            points_html:
+            - Changing a group to public will change of all descendants (groups and projects) to public. All users with and without explicit membership will be able to access.
```
joshsadam: Nit: Small typo — "change of all descendants" should be "change all descendants" (drop the "of").

joshsadam (changes requested):
I still think a switch would be a much better UI.  Just a couple things below to look at.

app/views/groups/_edit_change_visibility.html.erb
deepsidhu85: Fixed in Fixed in 3e4dad55aeb6067aad9b09478f2d9bbeee6903c6.

app/services/groups/update_service.rb
deepsidhu85: Fixed in 3e4dad55aeb6067aad9b09478f2d9bbeee6903c6.

We don't need to return early from the service as the model validation handles that for us

config/locales/en.yml
deepsidhu85: Thanks, I had fixed it in the french translation and missed it in the en file. Fixed in 3e4dad55aeb6067aad9b09478f2d9bbeee6903c6.

app/views/groups/_edit_change_visibility.html.erb
```diff
@@ -0,0 +1,59 @@
...
+        <div class="mb-4">
+          <% t("groups.edit.advanced.change_visibility.confirm.points_html", group: group.path).each do |item| %>
+            <p class="mb-2 text-base text-slate-900 dark:text-white"><%= item %></p>
+          <% end %>
+        </div>
+      </div>
+    </div>
+
+    <p class="mb-2 text-base text-slate-900 dark:text-white">
+      <% if @group.public? %>
```
joshsadam: Why can you not just use `group` like above and keep it consistent ...

joshsadam (changes requested):

app/views/groups/_edit_change_visibility.html.erb
deepsidhu85: Updated in ee16eeadd52a5d420b6cb5216c945faec031bd93

app/services/groups/update_service.rb
```diff
@@ -9,21 +9,26 @@ def initialize(group, user = nil, params = {})
       super
     end
 
-    def execute # rubocop:disable Metrics/AbcSize
-      authorize! @group, to: :update?
+    def execute # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity,Metrics/MethodLength
```
joshsadam, line 12: This file is getting heavy in responsibilities and branching; I think we might need to break this up a bit.  Maybe consider keeping the update service focused on name/path/description, and move visibility state transition into a dedicated service that owns propagation and activity behavior.

joshsadam (changes requested):

app/services/groups/update_service.rb
joshsadam: Fine, it's good

joshsadam (dismissed):

ericenns (changes requested):
Tried reviewing this but noticed that the code refers to `global_groups` feature flag but no entry in `config/features.yml` can you add it there with default to true in development please.

deepsidhu85:
> Tried reviewing this but noticed that the code refers to `global_groups` feature flag but no entry in `config/features.yml` can you add it there with default to true in development please.

Looks like I had a bad rebase once a PR was merged into main which changed the format of the features.yml file. Updated in bc14886211d83b2144fc5f241af49c170f70fcf8

app/services/base_group_service.rb
```diff
@@ -31,12 +31,12 @@ def update_descendants_to_private
     descendants_to_update = @group.self_and_descendants_of_type(
       [Group.sti_name,
        Namespaces::ProjectNamespace.sti_name]
-    ).where(public: true).order(type: :asc)
+    ).where(public: true).sort_by { |ns| ns.full_path.length }
```
ericenns: nitpick(blocking): Please update this to use database sorting `.order("LENGTH(routes.path) DESC")`

app/services/base_group_service.rb
```diff
@@ -13,12 +13,12 @@ def update_descendants_to_public
     descendants_to_update = @group.self_and_descendants_of_type(
       [Group.sti_name,
        Namespaces::ProjectNamespace.sti_name]
-    ).where(public: false).order(type: :asc)
+    ).where(public: false).sort_by { |ns| ns.full_path.length }
```
ericenns: nitpick (blocking): Can you update this to use database sorting, instead of in memory sorting. `.order("LENGTH(routes.path) DESC")`

app/services/groups/update_service.rb
ericenns, line 12: nitpick(blocking): Can you wrap the database operations within this method to be in a transaction, that way when using `update!` if error encountered it will rollback the transaction.

app/services/groups/create_service.rb
```diff
@@ -13,7 +13,9 @@ def initialize(user = nil, params = {})
     def execute # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
ericenns, line 13: nitpick(blocking): Can you wrap the database operations within this method to be in a transaction, that way when using `update!` if error encountered it will rollback the transaction.

app/services/groups/transfer_service.rb
```diff
@@ -29,10 +29,12 @@ def execute # rubocop:disable Metrics/AbcSize,Naming/PredicateMethod,Metrics/Met
```
ericenns, line 29: nitpick(blocking): Can you wrap the database operations within this method to be in a transaction, that way when using `update!` if error encountered it will rollback the transaction.

app/services/base_group_service.rb
```diff
@@ -13,12 +13,12 @@ def update_descendants_to_public
...
 
     descendants_to_update.each do |descendant|
       next if descendant.public == true
 
-      descendant.update!(public: true)
+      descendant.update(public: true)
```
ericenns: nitpick(blocking): Please change this back to `update!` so that when using transaction it is rolled back.

ericenns (changes requested):
Looking good, just a few comments.

app/services/base_group_service.rb
deepsidhu85: Updated in 6387c94efd12c28696bf2f874070bb8a86448b9b

app/services/base_group_service.rb
deepsidhu85: Updated in 6387c94efd12c28696bf2f874070bb8a86448b9b

app/services/groups/update_service.rb
deepsidhu85: Updated in 6387c94efd12c28696bf2f874070bb8a86448b9b

app/services/groups/create_service.rb
deepsidhu85: Updated in 6387c94efd12c28696bf2f874070bb8a86448b9b

app/services/groups/transfer_service.rb
deepsidhu85: Updated in 6387c94efd12c28696bf2f874070bb8a86448b9b

app/services/base_group_service.rb
deepsidhu85: Updated in 6387c94efd12c28696bf2f874070bb8a86448b9b

ericenns (approved):
Looks good to me!

PR #2012, opened 2026-06-22
Client file import feedback

ksierks:
Adds a progress window, flash, and dialog to display the progress and completion details of the line list import.

Successful import

https://github.com/user-attachments/assets/d3e02ebc-6a03-487f-b677-cb1220c8b11b

Unsuccessful import

https://github.com/user-attachments/assets/cdc74912-0a1a-4fef-a9d4-0efb619f1943

Enable the new feature:
1. Launch rails console and run the following:
```irb
Flipper.enable(:client_linelist_imports_v1)
```

Test the new feature:
1. Navigate to a samples page that the logged in user has access to update.
1. Hit the `Import Metadata` button.
1. Select a file that has headers with trailing spaces to import.
1. Select the `Sample Identifier Column`.
1. Select the `Selected Metadata`.
1. Hit the `Import Metadata` button.
1. Verify the progress window is displayed and updating correctly.
1. Verify a flash message is displayed if the import is successful.
1. Verify a dialog with alerts is displayed if the import is unsuccessful.

## PR acceptance checklist

### DIFF
--- a/app/javascript/controllers/linelist_import_controller.js
+++ b/app/javascript/controllers/linelist_import_controller.js
@@ -21,13 +21,21 @@ export default class extends Controller {
     graphqlUrl: String,
     groupPuid: String,
     projectPuid: String,
+    minimumVisibleDurationMs: {
+      type: Number,
+      default: 3500,
+    },
+    importStartedMessage: {
+      type: String,
+      default: "Starting metadata import...",
+    },
     importedRecordsMessage: {
       type: String,
       default: "Imported %{current} of %{total} records",
     },
     importCompleteMessage: {
       type: String,
-      default: "The metadata import is complete.",
+      default: "The metadata import is complete",
     },
     errorMessage: {
       type: String,
@@ -78,7 +86,7 @@ export default class extends Controller {
     closeDialog(this.element, this.application);
     this._operationId = `${Date.now()}-${Math.random().toString(36).slice(2, 7)}`;
     this._progressWindowOpenedAt = null;
-    showProgressWindow(this, "Importing samples...");
+    showProgressWindow(this, t(this.importStartedMessageValue));
     this.#processRows();
   }
 
--- a/app/javascript/controllers/linelist_import_controller.js
+++ b/app/javascript/controllers/linelist_import_controller.js
@@ -1,18 +1,22 @@
 import * as XLSX from "xlsx";
 import Controller from "controllers/metadata/file_import_controller";
 import { omitBy, pick } from "utilities/collection";
-import { closeDialog } from "utilities/dialog";
+import { closeDialog, ensureDialog } from "utilities/dialog";
+import { ensureFlash } from "utilities/flash";
 import { t } from "utilities/message_formatter";
 import {
-  clearProgressWindowDismissTimeout,
-  dismissProgressWindow,
   scheduleProgressWindowDismiss,
   showProgressWindow,
   updateProgressWindow,
 } from "utilities/progress_window";
 
 export default class extends Controller {
-  static targets = ["alertTemplate", "dialogTemplate", "progressTemplate"];
+  static targets = [
+    "alertTemplate",
+    "dialogTemplate",
+    "flashTemplate",
+    "progressTemplate",
+  ];
   static values = {
     graphqlUrl: String,
     groupPuid: String,
@@ -21,9 +25,9 @@ export default class extends Controller {
       type: String,
       default: "Imported %{current} of %{total} records",
     },
-    successMessage: {
+    importCompleteMessage: {
       type: String,
-      default: "The metadata was imported successfully!",
+      default: "The metadata import is complete.",
     },
     errorMessage: {
       type: String,
@@ -133,17 +137,17 @@ export default class extends Controller {
 
         if (payload.type === "progress") {
           console.log("Progress: ", payload);
-          this.onProgress(payload);
+          this.#onProgress(payload);
         }
 
         if (payload.type === "done") {
           console.log("Complete: ", payload);
-          this.onDone();
+          this.#onDone();
         }
 
         if (payload.type === "error") {
           console.log("Error: ", payload);
-          this.onError(payload);
+          this.#onError(payload);
         }
       };
     } else {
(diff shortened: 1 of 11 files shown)


app/javascript/utilities/message_formatter.js
```diff
@@ -0,0 +1,6 @@
+export function t(template, vars = {}) {
```
joshsadam, line 1: Nit: Now that `t()` is a shared utility (extracted from the export controller), could we add a small Vitest file under `test/javascript/utilities/`? A few cases for single/multiple `%{key}` substitution and missing keys would lock in the expected behaviour — see `test/javascript/utilities/collection.test.js` for the existing pattern.

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -1,18 +1,53 @@
...
+    errorMessage: {
+      type: String,
+      default: "Unexpected error while importing metadata: %{message}",
+    },
   };
 
   connect() {
     super.connect();
     this._fileType = null;
+    this._operationId ||= null;
```
joshsadam, line 50: Issue: `linelist_export_controller` initializes progress-window lifecycle state in `connect()` and resets it at the start of each operation, but the import controller doesn't yet. Without that parity, a dismissed progress window can block updates on the next import (`progressWindowDismissed` stays `true`), and a pending dismiss timeout from a prior import can remove the new one.

Could we mirror the export controller pattern?

```javascript
// connect()
this._dismissProgressWindowTimeout ||= null;
this.progressWindowDismissed ??= false;

// handleSubmit() — before showProgressWindow
clearProgressWindowDismissTimeout(this);
this.progressWindowDismissed = false;
```

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -126,6 +162,60 @@ export default class extends Controller {
     return worker;
   }
 
+  #errorMessage(error) {
```
joshsadam: Issue: `#errorMessage` substitutes server error text into `outerHTML` and inserts it with `insertAdjacentHTML`. Bulk-update errors include the sample identifier from the uploaded file (see `services.samples.metadata.bulk_update.sample_not_found`), so a crafted identifier could render as live HTML in the error dialog.

Could we set the alert message via `textContent` instead of string-replacing into HTML? For example, clone the template, find the message element, and assign `textContent` before appending the node (or use a dedicated placeholder element with `data-placeholder`).

app/javascript/utilities/flash.js
```diff
@@ -0,0 +1,19 @@
...
+    flash = createFlash(controller, flashId);
+  }
+
+  return flash;
+}
+
+function createFlash(controller, flashId) {
+  const flash = controller.flashTemplateTarget.content.cloneNode(true);
+  flash.firstElementChild.id = flashId;
+  const flashes = document.getElementById("flashes");
```
joshsadam, line 16: Nit: If `#flashes` isn't in the DOM, `appendChild` will throw and break the success path. Could we guard here (similar to how `ensureDialog` bails when `_operationId` is missing) and return `null` when the container isn't found?

app/javascript/controllers/linelist_import_controller.js
```diff
@@ -49,6 +84,9 @@ export default class extends Controller {
     this._worker?.terminate();
     this._worker ||= this.#buildWorker();
```
joshsadam: Issue: On a second submit while an import is still running, `this._worker?.terminate()` leaves a dead worker reference in place, and `||=` skips building a new one — so `#processRows()` can post to a terminated worker. Could we route this through `#terminateWorker()` (which nulls the ref) and always assign a fresh worker?

```javascript
this.#terminateWorker();
this._worker = this.#buildWorker();
```

joshsadam (changes requested):
Thanks for adding import progress, success flash, and error dialog feedback — the shared `progress_window` / `message_formatter` extraction and the template-based UI approach fit well with the export flow.

app/javascript/utilities/message_formatter.js
ksierks: a5d56e8aa391b3fb9b93c2e0c6585deb6f951d1c

app/javascript/controllers/linelist_import_controller.js
ksierks: fa87e33c7811a67fb03c6b1aa4b00f385a270fb9

app/javascript/utilities/flash.js
ksierks: 5fdf111c7893e498f4a8c3f1c64eaf7b1c3cf63f

app/javascript/controllers/linelist_import_controller.js
ksierks: 2d79ee2c9bc2f4f356be206ec2d317bcb2ccadbd

app/javascript/controllers/linelist_import_controller.js
ksierks: I made the change in 4e2243a66420c8173ffa869af4f2ffde629d0c0a, but dismissProgressWindow sets `controller._operationId = null` and `ensureDialog` & `ensureFlash` depends on this to display the import results.

joshsadam: This still needs a small follow-up.

 The reset needs to happen before `showProgressWindow()`. Right now `showProgressWindow()` runs while `progressWindowDismissed` may still be `true`, so the first update can return early and
  not show the new progress window.

Could you move these before `showProgressWindow()` in the `handleSubmit` method?

```javascript
  clearProgressWindowDismissTimeout(this);
  this.progressWindowDismissed = false;
```

Also, in the shared `utilities/progress_window.js`, `dismissProgressWindow()` does not appear to set `_operationId = null`, so I do not think that part blocks `ensureDialog()` or `ensureFlash()` as written.

ksierks: ad3b7e297b83b7cc1c60af4805cc1f08781a5ed0 for dismissing window before showing progress window.

`controller._operationId = null` was removed in the initial commit 4e2243a66420c8173ffa869af4f2ffde629d0c0a.

joshsadam (approved):

PR #2023, opened 2026-06-26
Archive projects: DB and policy updates

deepsidhu85:

This PR updates the policies for projects to deny certain actions if the project is archived. Added new db column `archived_at` to `namespaces` table. [STRY0021019](https://publichealthprod.service-now.com/rm_story.do?sys_id=3266ba843b464790d0d23a0eb3e45a9a)

1. Ensure tests pass

## PR acceptance checklist

### DIFF
--- a/db/migrate/20260615152555_add_archived_at_to_namespaces.rb
+++ b/db/migrate/20260615152555_add_archived_at_to_namespaces.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# Add archived_at column to namespace table
+class AddArchivedAtToNamespaces < ActiveRecord::Migration[8.1]
+  def change
+    add_column :namespaces, :archived_at, :datetime
+  end
+end
--- a/db/migrate/20260615152555_add_archived_at_to_namespaces.rb
+++ b/db/migrate/20260615152555_add_archived_at_to_namespaces.rb
@@ -4,5 +4,6 @@
 class AddArchivedAtToNamespaces < ActiveRecord::Migration[8.1]
   def change
     add_column :namespaces, :archived_at, :datetime
+    add_index :namespaces, :archived_at
   end
 end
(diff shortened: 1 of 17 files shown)


ericenns (changes requested):
This is looking like a good start, but I think it would be good to add an `Archivable` model concern that adds scopes `with_archived` and `without_archived`. Then the policy scopes that need to exclude archived projects can use `.without_archived` or we could add that as the default scope. I'm also wondering how we handle `samples_count` e.g. should archived projects be excluded from the count. Otherwise a Group with only an archived project that has samples could have weird rendering issues on the Sample page.

deepsidhu85:
> This is looking like a good start, but I think it would be good to add an `Archivable` model concern that adds scopes `with_archived` and `without_archived`. Then the policy scopes that need to exclude archived projects can use `.without_archived` or we could add that as the default scope. I'm also wondering how we handle `samples_count` e.g. should archived projects be excluded from the count. Otherwise a Group with only an archived project that has samples could have weird rendering issues on the Sample page.

Updated to use `archivable` concern in 5f7ca75ed4490c46e9e30f63868f52c1a1cf2df5. The aggregated samples_count will not include samples from archived projects, which will be updated in a follow-up pr

db/migrate/20260615152555_add_archived_at_to_namespaces.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Add archived_at column to namespace table
+class AddArchivedAtToNamespaces < ActiveRecord::Migration[8.1]
+  def change
+    add_column :namespaces, :archived_at, :datetime
```
ericenns, line 6: nitpick(blocking): Can you update this migration to include creating an index on this column, that will speed up query performance.

ericenns (changes requested):
Looking pretty good, just a comment on database migration

db/migrate/20260615152555_add_archived_at_to_namespaces.rb
deepsidhu85: Added in 2778c8aa7e83708b7052698608afdd615cc10ab8

ericenns (approved):
Looks good to me!

PR #2034, opened 2026-07-02
A11y: Fix tabbing issue with treegrid component when using screenreader navigation

ericenns:

This PR fixes a tabbing issue within the treegrid component when using a screenreader (e.g. NVDA) that switches to browse mode (which intercepts keyboard navigation). 

e.g. When using the treegrid with NVDA and after tabbing to a `button` or `link` in the row, then using arrow key navigation (this bypasses javascript) followed by a tab key could get the user into a situation where they can't tab out of the component.

To address this, instead of manually handling tab keypresses, I instead changed the javascript to listen to `focusin` events and update tabindex appropriately (this fixes the tab index after using arrow key navigation with a screen reader).

After:

https://github.com/user-attachments/assets/586e1472-b19a-4617-a43d-8551341f2171

1. Start the server with `bin/setup`
2. Launch NVDA
3. Navigate to Projects page (e.g. treegrid)
4. Tab into the treegrid and then tab to a `button` or `link` within the first row.
5. Hit the down arrow key 10 times followed by a tab
6. Observe that you are within a row and can tab normally which will go to end of row then out of component. Follow that with shift tabbing and confirm you renter the component and then go out of the component. This confirms tab navigation still works after using arrow key navigation with a screenreader

## PR acceptance checklist

### DIFF
--- a/app/javascript/controllers/treegrid_controller.js
+++ b/app/javascript/controllers/treegrid_controller.js
@@ -91,6 +91,9 @@ export default class extends Controller {
 
   focusin(event) {
     const newFocusedRow = this.#getContainingRow(event.target);
+    if (!newFocusedRow) {
+      return;
+    }
 
     tabbable(this.element).forEach((elem) => {
       if (elem !== newFocusedRow && !newFocusedRow.contains(elem)) {
--- a/app/javascript/controllers/treegrid_controller.js
+++ b/app/javascript/controllers/treegrid_controller.js
@@ -164,8 +164,9 @@ export default class extends Controller {
     const newRow = rows[direction > 0 ? rows.length - 1 : 0];
 
     if (currentRow !== newRow) {
+      const cellIndex = tabbable(currentRow).indexOf(document.activeElement);
       currentRow.tabIndex = -1;
-      this.#focus(newRow, -1);
+      this.#focus(newRow, cellIndex);
     }
   }
 
--- a/app/javascript/controllers/treegrid_controller.js
+++ b/app/javascript/controllers/treegrid_controller.js
@@ -186,6 +186,8 @@ export default class extends Controller {
   }
 
   #focus(elem, cellIndex) {
+    elem.tabIndex = 0;
+    this.#setTabIndexForElementsInRow(elem, 0);
     if (cellIndex < 0) {
       elem.focus();
     } else {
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/treegrid_controller.js
```diff
@@ -83,39 +82,26 @@ export default class extends Controller {
...
     }
 
     event.preventDefault();
   }
 
-  #handleTab(event) {
-    const direction = event.shiftKey ? -1 : +1;
-    const currentRow = this.#getRowWithFocus();
+  focusin(event) {
+    const newFocusedRow = this.#getContainingRow(event.target);
```
joshsadam, line 93: Whenever you query the DOM, add a guard to ensure that the element you were looking for is found.  Just something like:
```javascript
if(!newFocusedRow) {
  return;
}
```
This keeps the current behavior unchanged for normal row focus, but avoids throwing if focus ever lands on a focusable element inside the treegrid controller that is not inside `.treegrid-row`.

joshsadam (changes requested):
Won't be able to test it out until in the office, but this looks really good.  One small comment added.  We don't have any integration tests for this do we?

joshsadam:
From playing with the example https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/, once you have entered a row with the arrow key, if you press the up or down arrow key you should be in the equivalent element in the next row.

https://github.com/user-attachments/assets/7f535d30-4b13-47c0-9112-892cd74b5038

app/javascript/controllers/treegrid_controller.js
ericenns: added in https://github.com/phac-nml/irida-next/pull/2034/commits/5ba3ec182a0028a3f66969a17d0a4502d4b443c1

ericenns:
> From playing with the example https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/, once you have entered a row with the arrow key, if you press the up or down arrow key you should be in the equivalent element in the next row.
> 
>  Screencast.from.2026-07-06.13-38-02.mp4

Fixed in https://github.com/phac-nml/irida-next/pull/2034/commits/91f19a87f40a2a17a4e30830554954194fc53193

joshsadam (approved):
Thanks for the changes, this looks great

PR #2038, opened 2026-07-06
WES Error Messages

ksierks:
When cleaning up completed or errored workflow executions, fetch run and stdout logs from the GA4GH WES API and create attachments for them. 

Depends on STRY0020468 & STRY0020999.
Resolves STRY0016840.

image

Test sapporo:
1. Submit a workflow execution that errors. (eg. snvphylnfc with no reference file)
1. Click the workflow execution.
1. Navigate to the `Summary` tab.
1. Verify the `stdout.txt` & `stderr.txt` files.
1. Repeat all steps with a successful workflow execution.

Test WES:
1. Verify test suite passes.

## PR acceptance checklist

### DIFF
--- a/app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
+++ b/app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
@@ -20,23 +20,33 @@ module WorkflowExecutions
       # Don't run service if already cleaned
       return if @workflow_execution.nil? || @workflow_execution.cleaned?
 
-      step :create_log_attachments
+      step :workflow_execution_logs
+      step :attach_stdout_log
+      step :attach_stderr_log
       step :clean_up_blob_run_directory
       step :update_to_cleaned
     end
 
     private
 
-    def create_log_attachments
-      return if @workflow_execution.stdout.attached? || @workflow_execution.stderr.attached?
+    def workflow_execution_logs
+      @workflow_execution_logs ||= WorkflowExecutions::CleanupService.new(@workflow_execution).execute
+    end
 
-      result = WorkflowExecutions::CleanupService.new(@workflow_execution).execute
-      stdout = result[:stdout]
-      stderr = result[:stderr]
+    def attach_stdout_log
+      return if @workflow_execution.stdout.attached?
 
-      if stdout.present?
-        @workflow_execution.stdout.attach(io: StringIO.new(stdout), filename: 'stdout.txt', content_type: 'text/plain')
-      end
+      stdout = @workflow_execution_logs[:stdout]
+
+      return if stdout.blank?
+
+      @workflow_execution.stdout.attach(io: StringIO.new(stdout), filename: 'stdout.txt', content_type: 'text/plain')
+    end
+
+    def attach_stderr_log
+      return if @workflow_execution.stderr.attached?
+
+      stderr = @workflow_execution_logs[:stderr]
 
       return if stderr.blank?
 
--- a/app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
+++ b/app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
@@ -20,7 +20,6 @@ module WorkflowExecutions
       # Don't run service if already cleaned
       return if @workflow_execution.nil? || @workflow_execution.cleaned?
 
-      step :workflow_execution_logs
       step :attach_stdout_log
       step :attach_stderr_log
       step :clean_up_blob_run_directory
@@ -34,9 +33,7 @@ module WorkflowExecutions
     end
 
     def attach_stdout_log
-      return if @workflow_execution.stdout.attached?
-
-      stdout = @workflow_execution_logs[:stdout]
+      stdout = workflow_execution_logs[:stdout]
 
       return if stdout.blank?
 
@@ -44,9 +41,7 @@ module WorkflowExecutions
     end
 
     def attach_stderr_log
-      return if @workflow_execution.stderr.attached?
-
-      stderr = @workflow_execution_logs[:stderr]
+      stderr = workflow_execution_logs[:stderr]
 
       return if stderr.blank?
 
--- a/app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
+++ b/app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
@@ -27,14 +27,25 @@ module WorkflowExecutions
 
     private
 
-    def create_log_attachments
+    def create_log_attachments # rubocop:disable Metrics/AbcSize
+      run_log_filename = 'run_log.json'
+      run_stdout_filename = 'run_stdout.txt'
+      existing_output_filenames = @workflow_execution.outputs.map { |output| output.filename.to_s }
+
+      if existing_output_filenames.include?(run_log_filename) || existing_output_filenames.include?(run_stdout_filename)
+        return
+      end
+
       # Create attachments for pipeline run & stdout logs if they exist
       result = WorkflowExecutions::CleanupService.new(@workflow_execution).execute
       run_log = result[:run_log]
       run_stdout = result[:run_stdout]
       files = []
-      files << { io: StringIO.new(sanitize(run_log.to_json)), filename: 'run_log.json' } if run_log.present?
-      files << { io: StringIO.new(sanitize(run_stdout.to_json)), filename: 'run_stdout.txt' } if run_stdout.present?
+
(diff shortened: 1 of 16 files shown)


ericenns (changes requested):
The story wasn't clear, but we don't want to the stdout and stderr to be stored as normal workflow attachments as the pipeline could produce files with the name you selected and that would cause issues. Additionally it looks like only `stdout` is being retrieved and we want it to also retrieve `stderr`. These should be stored as special attachments e.g. `has_one_attached :stdout` and `has_one_attached :stderr` and in a followup PR the UI should be updated to show these under the summary tab if they are present and another followup PR to make these accessible via the graphql api.

ksierks:
> The story wasn't clear, but we don't want to the stdout and stderr to be stored as normal workflow attachments as the pipeline could produce files with the name you selected and that would cause issues. Additionally it looks like only `stdout` is being retrieved and we want it to also retrieve `stderr`. These should be stored as special attachments e.g. `has_one_attached :stdout` and `has_one_attached :stderr` and in a followup PR the UI should be updated to show these under the summary tab if they are present and another followup PR to make these accessible via the graphql api.

I can do that, but I'm curious why you had stdout and stderr merged in STRY0020468?

ericenns:
> > The story wasn't clear, but we don't want to the stdout and stderr to be stored as normal workflow attachments as the pipeline could produce files with the name you selected and that would cause issues. Additionally it looks like only `stdout` is being retrieved and we want it to also retrieve `stderr`. These should be stored as special attachments e.g. `has_one_attached :stdout` and `has_one_attached :stderr` and in a followup PR the UI should be updated to show these under the summary tab if they are present and another followup PR to make these accessible via the graphql api.
> 
> I can do that, but I'm curious why you had stdout and stderr merged in STRY0020468?

I didn't request for them to be merged, I had originally asked our IO team to populate both endpoints for our institutional WES implementation with separated stdout and stderr. They were having trouble getting separate files and were providing the same file to both the stdout and stderr endpoints, so I asked them to just populate the stdout endpoint so that we wouldn't have to do anything special on our side to determine the same content and just persist the file once. I would love for the outputs to be separate but at the time they didn't know of a way to do it.

test/jobs/workflow_executions/workflow_execution_cleanup_job_test.rb
```diff
@@ -7,16 +7,92 @@ module WorkflowExecutions
...
+      assert workflow_execution.stdout.attached?
+      assert workflow_execution.stderr.attached?
+      assert_equal 'stdout.txt', workflow_execution.stdout.filename.to_s
+      assert_equal 'stderr.txt', workflow_execution.stderr.filename.to_s
+
+      assert_performed_jobs(1, only: WorkflowExecutionCleanupJob)
+      assert_enqueued_jobs(0, except: Turbo::Streams::BroadcastStreamJob)
+    end
+
+    test 'cleanup log attachment creation is idempotent when rerun' do
```
JeffreyThiessen: This test is not testing what it says it is testing.

Line 52 and 66 are separate job instances, so this code is running two separate jobs on the same workflow execution. This test is good to include for ensuring the job exits early (line 21 of workflow_execution_cleanup_job.rb), but line 61 of this test must be removed so we are not wiping out the effects of the first job.
This test should be renamed to something like "cleanup job exits early on rerun"

Please add 2 more test cases to `/test/services/workflow_executions/cleanup_job_test.rb` to that test the idempotency of the two new steps. They should follow a similar structure to `clean workflow execution run directory files with interrupt` by using `interrupt_job_after_step` to have the job rerun using `ActiveJob::Continuable`

More info on `ActiveJob::Continuable` : https://api.rubyonrails.org/classes/ActiveJob/Continuation.html
and testing continuable jobs: https://api.rubyonrails.org/classes/ActiveJob/Continuation/TestHelper.html

app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
```diff
@@ -20,10 +20,39 @@ def perform(workflow_execution)
...
     end
 
+    private
+
+    def workflow_execution_logs
+      @workflow_execution_logs ||= WorkflowExecutions::CleanupService.new(@workflow_execution).execute
+    end
+
+    def attach_stdout_log
+      return if @workflow_execution.stdout.attached?
```
JeffreyThiessen: This early return is not needed. See my comment on how to correctly test these steps.

app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
```diff
@@ -20,10 +20,39 @@ def perform(workflow_execution)
...
+
+      stdout = @workflow_execution_logs[:stdout]
+
+      return if stdout.blank?
+
+      @workflow_execution.stdout.attach(io: StringIO.new(stdout), filename: 'stdout.txt', content_type: 'text/plain')
+    end
+
+    def attach_stderr_log
+      return if @workflow_execution.stderr.attached?
```
JeffreyThiessen: This early return is not needed. See my comment on how to correctly test these steps.

JeffreyThiessen (changes requested):
Code looks good overall 👍, just need some additions for testing idempotent jobs

app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
ksierks: b68337b318a7cec286fa599fce953facfe0d92af

app/jobs/workflow_executions/workflow_execution_cleanup_job.rb
ksierks: b68337b318a7cec286fa599fce953facfe0d92af

test/jobs/workflow_executions/workflow_execution_cleanup_job_test.rb
ksierks: b68337b318a7cec286fa599fce953facfe0d92af uses `interrupt_job_after_step`. I removed the step `workflow_execution_logs` as it's not needed and the original test. I create two new tests, that test interruptions after each step `attach_stdout_log` & `attach_stdout_log`.

test/jobs/workflow_executions/workflow_execution_cleanup_job_test.rb
```diff
@@ -7,16 +7,92 @@ module WorkflowExecutions
...
+            WorkflowExecutionCleanupJob.perform_later(workflow_execution)
+          end
+        end
+      end
+
+      workflow_execution.reload
+      stdout_attachment_id = workflow_execution.stdout.attachment.id
+      stderr_attachment_id = workflow_execution.stderr.attachment.id
+
+      workflow_execution.update!(cleaned: false)
```
JeffreyThiessen: This line should be removed, otherwise below this line we are testing a state that cannot occur in the system. If this line is removed it does correctly test the potential edge case where the job might get double queued or something like that, and then runs on an already cleaned workflow execution.

JeffreyThiessen (changes requested):
Looks good, just one last tiny change

test/jobs/workflow_executions/workflow_execution_cleanup_job_test.rb
ksierks: The line `workflow_execution.update!(cleaned: false)` does not exist anymore. I think you might be looking at an old diff?

JeffreyThiessen: Oops my bad

JeffreyThiessen (dismissed):
Looks good 👍

app/services/workflow_executions/cleanup_service.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module WorkflowExecutions
+  # Service used to Cleanup a WorkflowExecution
+  class CleanupService < BaseService
```
ericenns: nitpick(blocking): Could you rename this to `RunLogFetchService` instead since that encompasses the purpose a bit better. Along with updating the class comment.

ericenns (changes requested):
This looks good, just a commenting on the service name to match the functionality better.

app/services/workflow_executions/cleanup_service.rb
ksierks: 4186c5cb66ea61d57dc96c583d35c0e51defb852

ericenns (approved):
Looks good to me!

PR #2067, opened 2026-07-13
GraphQL: WES Error Messages

ksierks:
This PR exposes the newly added stdout & stderr workflow execution attachments from #2038.

N/A

1. Run a workflow execution.
2. Navigate to `http://localhost:3000/graphiql`.
3. Run the following query:
```
query getLastWorkflowExecution {
  workflowExecutions(last:1) {
    totalCount
    nodes {
      id
      runId
      stderrUrl
      stdoutUrl
      samplesWorkflowExecutions {
        sample {
          id
        }
      }
    }
  }
}
```
4. Verify stdout & stderr.

## PR acceptance checklist


ChrisHuynh333 (dismissed):

app/graphql/types/workflow_execution_type.rb
```diff
@@ -24,6 +24,8 @@ class WorkflowExecutionType < Types::BaseObject
           complexity: 5,
           resolver: Resolvers::SamplesWorkflowExecutionsResolver
     field :state, String, null: true, description: 'WorkflowExecution state'
+    field :stderr, String, null: true, description: 'WorkflowExecution stderr download url'
```
ericenns: nitpick(blocking): Is there a reason why this is not returning an `AttachmentType` as done in other places where we provide access to the attachments via graphql?

I feel like it would be better to return an AttachmentType instead, that way the user through GraphQL can get details of the file e.g. size etc which would better allow them to confirm they downloaded the entire file etc.

app/graphql/types/workflow_execution_type.rb
```diff
@@ -24,6 +24,8 @@ class WorkflowExecutionType < Types::BaseObject
...
+    field :stdout, String, null: true, description: 'WorkflowExecution stdout download url'
```
ericenns: nitpick(blocking): Same comment as above.

ericenns (changes requested):

app/graphql/types/workflow_execution_type.rb
ksierks: AttachmentType is expecting a custom Attachment record. stdout & stderr uses has_one_attached which returns ActiveStorage::Attached::One.

ericenns: That makes sense, in that case would maybe renaming the field to `stderr_url` and `stdout_url` respectively make a bit more sense. We do that in AttachmentType e.g. there we have `attachment_url`. That way if we wanted to expose `stderr_byte_size` and `stderr_filename` in the future (not this pr, and not really a priority) we could.

ksierks: 2cb2354c43e3062eb287026da6e3e6b9fc69e3a1

JeffreyThiessen (approved):
Code looks great! Works as intended. 👍

ericenns (approved):
Looks good to me!

PR #2088, opened 2026-07-15
Add Reason to Sample Deletion

ksierks:
This PR adds a reason field to the form within the `Delete Sample(s)` modals and displays the reason under activity tracking.
Resolves STRY0017536.

image

image

image

Test single sample deletion from sample page:
1. Navigate to a sample page.
1. Click the `Remove` button.
1. Verify the `Delete Sample` modal is displayed.
1. Add a reason to the form.
1. Verify the reason is displayed under `Project Activity`.
1. Verify the sample deletion reason with `Sample.with_deleted.last` via rails console.

Test multiple sample deletion from project samples page:
1. Navigate to a project samples page.
1. Select samples within the table that are to be deleted.
1. Click the `Sample Actions` -> `Delete samples` button.
1. Verify the `Delete Samples` modal is displayed.
1. Add a reason to the form.
1. Verify the reason is displayed under `Project Activity`.
1. Verify the sample deletion reasons with `Sample.with_deleted.last(x)` via rails console.

Test multiple sample deletion from group samples page:
1. Navigate to a group samples page.
1. Select samples within the table that are to be deleted.
1. Click the `Sample Actions` -> `Delete samples` button.
1. Verify the `Delete Samples` modal is displayed.
1. Add a reason to the form.
1. Verify the reason is displayed under `Group Activity`.
1. Verify the sample deletion reasons with `Sample.with_deleted.last(x)` via rails console.

## PR acceptance checklist

### DIFF
--- a/app/models/audit_form.rb
+++ b/app/models/audit_form.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+# entity class for audit form
+class AuditForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Validations::Callbacks
+
+  MAX_REASON_LENGTH = 500
+
+  attribute :reason, :string
+
+  before_validation :normalize_reason
+
+  validates :reason, presence: true, length: { maximum: MAX_REASON_LENGTH }
+
+  private
+
+  def normalize_reason
+    self.reason = reason.to_s.squish
+  end
+end
--- a/app/models/audit_form.rb
+++ b/app/models/audit_form.rb
@@ -9,14 +9,19 @@ class AuditForm
   MAX_REASON_LENGTH = 500
 
   attribute :reason, :string
+  attr_accessor :user
 
   before_validation :normalize_reason
 
-  validates :reason, presence: true, length: { maximum: MAX_REASON_LENGTH }
+  validates :reason, presence: true, length: { maximum: MAX_REASON_LENGTH }, if: :reason_required?
 
   private
 
   def normalize_reason
     self.reason = reason.to_s.strip
+  end
+
+  def reason_required?
+    user.nil? || Flipper.enabled?(:sample_deletion_reason, user)
   end
 end
--- a/app/models/audit_form.rb
+++ b/app/models/audit_form.rb
@@ -17,6 +17,6 @@ class AuditForm
   private
 
   def normalize_reason
-    self.reason = reason.to_s.squish
+    self.reason = reason.to_s.strip
   end
 end
--- a/app/services/base_sample_destroy_service.rb
+++ b/app/services/base_sample_destroy_service.rb
@@ -28,10 +28,10 @@ class BaseSampleDestroyService < BaseService
       samples_deleted_count: deleted_samples_data.size,
       deleted_samples_data: deleted_samples_data
     }
-
+    activity_params = { samples_deleted_count: deleted_samples_data.size, action: 'sample_destroy_multiple' }
     ext_details = ExtendedDetail.create!(details: details)
-
     key = if params[:reason].present?
+            activity_params[:reason] = params[:reason]
             'namespaces_project_namespace.samples.destroy_multiple_with_reason'
           else
             'namespaces_project_namespace.samples.destroy_multiple'
@@ -39,12 +39,7 @@ class BaseSampleDestroyService < BaseService
 
     activity = project_namespace.create_activity key: key,
                                                  owner: current_user,
-                                                 parameters:
-                                                 {
-                                                   samples_deleted_count: deleted_samples_data.size,
-                                                   action: 'sample_destroy_multiple',
-                                                   reason: params[:reason]
-                                                 }
+                                                 parameters: activity_params
     activity.create_activity_extended_detail(extended_detail_id: ext_details.id,
                                              activity_type: 'sample_destroy_multiple')
   end
--- a/app/services/base_sample_destroy_service.rb
+++ b/app/services/base_sample_destroy_service.rb
@@ -24,8 +24,13 @@ class BaseSampleDestroyService < BaseService
   end
 
   def create_project_activity(project_namespace, deleted_samples_data)
-    ext_details = ExtendedDetail.create!(details: { samples_deleted_count: deleted_samples_data.size,
-                                                    deleted_samples_data: deleted_samples_data })
+    details = {
+      samples_deleted_count: deleted_samples_data.size,
+      deleted_samples_data: deleted_samples_data,
+      deletion_reason: params[:reason]
+    }
+
+    ext_details = ExtendedDetail.create!(details: details)
 
     activity = project_namespace.create_activity key: 'namespaces_project_namespace.samples.destroy_multiple',
                                                  owner: current_user,
--- a/app/services/base_sample_destroy_service.rb
+++ b/app/services/base_sample_destroy_service.rb
@@ -23,21 +23,27 @@ class BaseSampleDestroyService < BaseService
     sample.project.namespace.update_metadata_summary_by_sample_deletion(sample)
   end
 
-  def create_project_activity(project_namespace, deleted_samples_data)
+  def create_project_activity(project_namespace, deleted_samples_data) # rubocop:disable Metrics/MethodLength
     details = {
       samples_deleted_count: deleted_samples_data.size,
-      deleted_samples_data: deleted_samples_data,
-      deletion_reason: params[:reason]
+      deleted_samples_data: deleted_samples_data
     }
 
     ext_details = ExtendedDetail.create!(details: details)
 
-    activity = project_namespace.create_activity key: 'namespaces_project_namespace.samples.destroy_multiple',
+    key = if params[:reason].present?
+            'namespaces_project_namespace.samples.destroy_multiple_with_reason'
+          else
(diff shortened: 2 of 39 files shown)


ChrisHuynh333 (dismissed):

deepsidhu85 (commented):
Just an overall comment but I believe we should be saving the sample deletion reason on the actual record (requires a new attribute on the Sample model) and not just in an activity. We use logidze to keep an audit log of changes made to the sample model, but the way this PR is set up, the reason for deletion would not be populated in the sample object log data

ksierks:
> Just an overall comment but I believe we should be saving the sample deletion reason on the actual record (requires a new attribute on the Sample model) and not just in an activity. We use logidze to keep an audit log of changes made to the sample model, but the way this PR is set up, the reason for deletion would not be populated in the sample object log data

13f68be070c07fbb26b514ea82fd34d3ae0e4e93 adds the new migration. I'd like to keep the audit form, if that's ok?

app/services/groups/samples/destroy_service.rb
```diff
@@ -17,6 +17,7 @@ def destroy_samples
         samples.each do |sample|
           next unless sample.deleted?
 
+          sample.update_column(:deletion_reason, params[:reason]) # rubocop:disable Rails/SkipsModelValidations
```
deepsidhu85: Currently say for example we deleted 1000 samples, this would add 1000 db calls. I think it would be better for us to keep track of the sample ids in the loop and then make 1 call after the loop to the db to update those rows with the reason

app/services/projects/samples/destroy_service.rb
```diff
@@ -15,6 +15,7 @@ def destroy_samples
         samples.each do |sample|
           next unless sample.deleted?
 
+          sample.update_column(:deletion_reason, params[:reason]) # rubocop:disable Rails/SkipsModelValidations
```
deepsidhu85: Same comment as above, I think it would be better for us to keep track of the sample ids in the loop and then make 1 call after the loop to the db to update those rows with the reason

config/locales/en.yml
```diff
@@ -1896,12 +1896,14 @@ en:
           singular: 'The following sample has been selected for deletion:'
           zero: No samples have been selected for deletion
         loading: Loading...
+        reason_placeholder: Add a reason for activity tracking
```
deepsidhu85: Since we have the same placeholder added twice, put this under a shared key which can be used for both. Also, I think we should update this to `Provide a justification for deletion`

deepsidhu85 (changes requested):
This is coming along nicely, just a few small comments below

app/services/groups/samples/destroy_service.rb
ksierks: 23f936786cb5aed91d4df0adb3d6a7824fe9653f

app/services/projects/samples/destroy_service.rb
ksierks: 23f936786cb5aed91d4df0adb3d6a7824fe9653f

config/locales/en.yml
ksierks: 64985e669a4872fb8c1a51591004d00c3479bdb7

app/services/groups/samples/destroy_service.rb
```diff
@@ -6,21 +6,27 @@ module Samples
...
         samples.each do |sample|
           next unless sample.deleted?
 
+          deleted_sample_ids << sample.id
           update_metadata_summary(sample)
           add_deleted_sample_to_data(sample, sample.project.puid, sample.project.name)
         end
 
+        if deleted_sample_ids.any?
+          Sample.with_deleted.where(id: deleted_sample_ids).update_all(deletion_reason: params[:reason]) # rubocop:disable Rails/SkipsModelValidations
```
deepsidhu85: This will affect the other places in this PR where `Sample.with_deleted` was used. We should be able to limit this to just the deleted samples by using the `.only_deleted` scope

deepsidhu85 (changes requested):
Just a comment about scope used below

app/services/groups/samples/destroy_service.rb
ksierks: 9c282027096c79c84693723817a399f42e25b3d1

deepsidhu85 (changes requested):
This looks great just thinking we should remove the reason line from the dialog and just display it in the actual activity on the Activity pages. Maybe add it on the line before the `More details` button and can just read `Reason: REASON_FOR_DELETION`. The dialogs are mainly used just to list which samples were modified/deleted/transferred/cloned

ksierks:
> This looks great just thinking we should remove the reason line from the dialog and just display it in the actual activity on the Activity pages. Maybe add it on the line before the `More details` button and can just read `Reason: REASON_FOR_DELETION`. The dialogs are mainly used just to list which samples were modified/deleted/transferred/cloned

f8420f159eb23f59a410ece3cb31235cc73ed965

app/services/groups/samples/destroy_service.rb
```diff
@@ -42,19 +48,25 @@ def create_activities_and_update_samples_count
...
+                'group.samples.destroy'
+              end
+        group_activity = @namespace.create_activity key: key,
                                                     owner: current_user,
                                                     parameters:
                               {
                                 samples_deleted_count: total_deleted_samples_count,
-                                action: 'group_samples_destroy'
+                                action: 'group_samples_destroy',
+                                reason: params[:reason]
```
deepsidhu85: This would add the reason key to the parameters even if it isn't present.

We should be able to update this method to 

```
      def create_group_activity(total_deleted_samples_count) # rubocop:disable Metrics/MethodLength
        details = {
          deleted_samples_data: @deleted_samples_data[:group_data],
          samples_deleted_count: total_deleted_samples_count
        }
        activity_params = { samples_deleted_count: total_deleted_samples_count,
                            action: 'group_samples_destroy' }
        group_ext_details = ExtendedDetail.create!(details: details)
        if params[:reason].present?
          key = 'group.samples.destroy_with_reason'
          activity_params[:reason] = params[:reason]
        else
          key = 'group.samples.destroy'
        end

        group_activity = @namespace.create_activity key: key,
                                                    owner: current_user,
                                                    parameters: activity_params

        group_activity.create_activity_extended_detail(extended_detail_id: group_ext_details.id,
                                                       activity_type: 'group_samples_destroy')
      end
```

app/services/base_sample_destroy_service.rb
```diff
@@ -23,16 +23,27 @@ def update_metadata_summary(sample)
...
+          end
+
+    activity = project_namespace.create_activity key: key,
                                                  owner: current_user,
                                                  parameters:
                                                  {
                                                    samples_deleted_count: deleted_samples_data.size,
-                                                   action: 'sample_destroy_multiple'
+                                                   action: 'sample_destroy_multiple',
+                                                   reason: params[:reason]
```
deepsidhu85: Same comment as above. This can be refactored so the reason is only set for an activity if the reason is present in the params

deepsidhu85 (changes requested):
Last couple of comments to clean up setting the reason in the activity parameters otherwise looks great

app/services/groups/samples/destroy_service.rb
ksierks: cfdca7101d0703f0e61ac5b34154008219069112

app/services/base_sample_destroy_service.rb
ksierks: cfdca7101d0703f0e61ac5b34154008219069112

deepsidhu85 (dismissed):
Looks good to me!

app/models/audit_form.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
```
deepsidhu85: Lets rename this form to `SampleDeletionForm`

app/models/audit_form.rb
```diff
@@ -0,0 +1,27 @@
...
+  include ActiveModel::Validations::Callbacks
+
+  MAX_REASON_LENGTH = 500
+
+  attribute :reason, :string
+  attr_accessor :user
+
+  before_validation :normalize_reason
+
+  validates :reason, presence: true, length: { maximum: MAX_REASON_LENGTH }, if: :reason_required?
```
deepsidhu85: We can remove the if portion and remove the reason_required method. We can just handle this wherever it is called from

app/services/groups/samples/destroy_service.rb
```diff
@@ -6,21 +6,27 @@ module Samples
...
       private
 
-      def destroy_samples
+      def destroy_samples # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         samples = authorized_scope(Sample, type: :relation, as: :namespace_samples,
                                            scope_options: { namespace: @namespace,
                                                             minimum_access_level: Member::AccessLevel::OWNER })
                   .where(id: @sample_ids)
         @deleted_samples_data = { project_data: {}, group_data: [] }
+        deleted_sample_ids = []
```
deepsidhu85: Anything to do with deleted_sample_ids should have a check to see if the flipper flag `sample_deletion_reason` is enabled

deepsidhu85 (approved):
Looks good to me! Thanks for adding in the feature flag

PR #2089, opened 2026-07-15
Limit data export size

### LINKED ISSUE #2074: Limit the size of data exports
## Problem

Large data exports can use too much temporary disk space and interrupt background jobs. IRIDA Next currently lets users start these exports without checking the total file size first.

## What we will build

Block sample and analysis exports when the files they include are 30 GB or larger. Show a clear message in the export dialog and tell the user to download the files through the GraphQL API instead.

## How we will build it

- Add `max_data_export_size_gigabytes` to the application settings.
- Set the default to `30`.
- Only allow positive whole numbers for this setting.
- Treat 30 GB as `30.gigabytes` in Rails, which is 32,212,254,720 bytes.
- Add one service that totals the source file sizes from Active Storage.
- Use the file sizes already stored in the database. Do not download files to calculate the total.
- For sample exports:
  - Count files from the selected samples.
  - Count only the file formats selected by the user.
- For analysis exports:
  - Count workflow output files.
  - Count output files connected to each sample in the workflow.
- Do not apply this limit to linelist exports because they do not include source files.
- Check permissions and selected record IDs before calculating the size.
- Check the size before saving the export or adding a background job.
- Block an export when its source files are equal to or larger than the setting.
- Keep the export dialog open when the export is blocked.
- Show this message in English and French:
  - The size of the data you are trying to export is too large.
  - Data exports must be smaller than 30 GB.
  - Please download files through the GraphQL API.
- Show the current setting in the message instead of always showing 30 GB.
- Do not record blocked attempts in logs or a new database table.

## Testing

- Add application setting tests for:
  - The 30 GB default.
  - A custom value.
  - Zero, negative, and non-whole values being rejected.
- Add service tests for:
  - Sample exports with selected file formats.
  - Analysis exports with workflow and per-sample outputs.
  - Exports with no files.
  - Linelist exports returning a size of zero.
  - A file counted more than once when it would be copied more than once.
- Add export creation tests for:
  - A size below the limit being accepted and queued.
  - A size equal to the limit being blocked.
  - A size above the limit being blocked.
  - No export record or background job being created when blocked.
- Add a controller test that checks the `422` response and the translated message in the export dialog.
- Use the existing data export service and controller tests as examples.

## Delivery plan

1. **Limit data exports before they are queued**
   - Add the setting, file-size calculation, creation check, translated message, and tests.
   - Keep the current export job and GraphQL downloads unchanged.
   - No other work needs to be merged first.
   - Done when:
     - [ ] The setting defaults to 30 GB and rejects invalid values.
     - [ ] Sample and analysis export sizes count the correct files.
     - [ ] Exports at or above the limit are not saved or queued.
     - [ ] Users see the translated message in the export dialog.
     - [ ] Linelist exports continue to work.
     - [ ] Focused tests pass with `PARALLEL_WORKERS=4`.

joshsadam:

- Adds a `max_data_export_size_gigabytes` application setting (default 30 GB).
- Blocks sample and analysis exports whose source files are larger than the limit, before the export job runs.
- Shows the error inside the export dialog and keeps the dialog open. Closes #2074.
- [ServiceNow Issue](https://publichealthprod.service-now.com/rm_story.do?sys_id=1ebc223b47d27a50f24c0c21516d436f)

image

1. Run migrations with `bin/rails db:migrate`.
2. Run the tests: `PARALLEL_WORKERS=4 bin/rails test test/services/data_exports/export_source_size_calculator_test.rb test/services/data_exports/create_service_test.rb test/models/application_setting_test.rb test/controllers/data_exports_controller_test.rb`.
3. Start the app with `bin/dev`.
4. In the Rails console, lower the limit: `Irida::CurrentSettings.current_application_settings.update!(max_data_export_size_gigabytes: 1)`.
5. Try to export samples or analysis outputs larger than 1 GB and confirm the error appears in the dialog and the export is not created.
6. Confirm an export under the limit still works.

## PR acceptance checklist


test/services/data_exports/export_source_size_calculator_test.rb
```diff
@@ -0,0 +1,94 @@
...
+      expected = sample1.attachments.sum { |attachment| attachment.file.byte_size }
+
+      total_size = calculate('sample', 'ids' => [sample1.id, sample3.id], 'attachment_formats' => ['fastq'])
+
+      assert_equal expected, total_size
+    end
+
+    test 'sample export totals all attachments when every format is selected' do
+      sample1 = samples(:sample1)
+
```
ChrisHuynh333: Just a nitpick. For these first two tests, rather than using sample1, which only has fastq files attached, can you use a sample with multiple attachment formats attached (like sample22)? Just an extra layer of testing to test filtering within a samples own attachments in case the filtering gets changed/broken

ChrisHuynh333 (dismissed):
Looks good, just one nitpick below

test/services/data_exports/export_source_size_calculator_test.rb
joshsadam: Fix in 90a137e5e

JeffreyThiessen (dismissed):
Looks great, seems to work with the edge cases I could think of.

ericenns (approved):
Looks good to me!

PR #2105, opened 2026-07-21
Combobox Component renders Options via Slots

ksierks:
This PR is accepts options as slots for the ComboboxComponent which will be needed to replace all the Select2Components in a later PR.

N/A

1. Navigate to http://localhost:3000/rails/lookbook/inspect/combobox/with_slots.
1. Verify the combobox works as expected.

## PR acceptance checklist


joshsadam:
I was just looking through your PR and noticed that the option data goes through several conversions. The slots are rendered as HTML, parsed again with Nokogiri, and then JavaScript works out the selected option from its label text.

Ruby option data
    → HTML `<option>` or slot HTML
    → Nokogiri parses HTML
    → Ruby rewrites it as listbox HTML
    → browser parses HTML
    → JavaScript clones and filters DOM
    → JavaScript infers selection from display text

  That produces several competing sources of truth:

  - `<option selected>`
  - `data-selected`
  - `aria-selected`
  - hidden input value
  - visible input label
  - controller #option
  - rendered text used for filtering

  The resulting bugs are symptoms of that structure.

I like the goal of supporting richer option content, but I worry this makes the selection state difficult to follow. It can also select the wrong value when two options have the same label.

Would you be open to keeping the selected value on the main ComboboxComponent, rendering slot options directly, and using Nokogiri only as a compatibility step for the existing options: argument? JavaScript could then initialize the selected option by matching data-value instead of its label. I think that would give us one source of truth and make the accessibility state easier to manage as well.

You could set up the component to be called (example):

```html
<%= render ComboboxComponent.new(
    form:,
    field:,
    selected_value: model.user_id,
  ) do |combobox| %>
    <% users.each do |user| %>
      <% combobox.with_option(
        value: user.id,
        label: user.full_name,
        disabled: user.disabled?,
      ) do %>
        ...
      <% end %>
    <% end %>
  <% end %>
```

Important differences:

  - `selected_value` belongs to the parent module.
  - Individual options do not accept `selected:`.
  - Slots render directly through OptionComponent.
  - Nokogiri is used only by the legacy adapter.
  - Slot HTML is never rendered, reparsed, mutated, and serialized.
  - Content defaults to `label` when no block is supplied.
  - The hidden value is the canonical identity.

As always, maybe I am off what @ericenns wants for this update.

ksierks:
> I was just looking through your PR and noticed that the option data goes through several conversions. The slots are rendered as HTML, parsed again with Nokogiri, and then JavaScript works out the selected option from its label text.
> 
> Ruby option data → HTML `<option>` or slot HTML → Nokogiri parses HTML → Ruby rewrites it as listbox HTML → browser parses HTML → JavaScript clones and filters DOM → JavaScript infers selection from display text
> 
> That produces several competing sources of truth:
> 
> * `<option selected>`
> * `data-selected`
> * `aria-selected`
> * hidden input value
> * visible input label
> * controller #option
> * rendered text used for filtering
> 
> The resulting bugs are symptoms of that structure.
> 
> I like the goal of supporting richer option content, but I worry this makes the selection state difficult to follow. It can also select the wrong value when two options have the same label.
> 
> Would you be open to keeping the selected value on the main ComboboxComponent, rendering slot options directly, and using Nokogiri only as a compatibility step for the existing options: argument? JavaScript could then initialize the selected option by matching data-value instead of its label. I think that would give us one source of truth and make the accessibility state easier to manage as well.
> 
> You could set up the component to be called (example):
> 
> ```
> <%= render ComboboxComponent.new(
>     form:,
>     field:,
>     selected_value: model.user_id,
>   ) do |combobox| %>
>     <% users.each do |user| %>
>       <% combobox.with_option(
>         value: user.id,
>         label: user.full_name,
>         disabled: user.disabled?,
>       ) do %>
>         ...
>       <% end %>
>     <% end %>
>   <% end %>
> ```
> 
> Important differences:
> 
> * `selected_value` belongs to the parent module.
> * Individual options do not accept `selected:`.
> * Slots render directly through OptionComponent.
> * Nokogiri is used only by the legacy adapter.
> * Slot HTML is never rendered, reparsed, mutated, and serialized.
> * Content defaults to `label` when no block is supplied.
> * The hidden value is the canonical identity.
> 
> As always, maybe I am off what @ericenns wants for this update.

I think I was trying to copy the Select2Component format. Is this closer to what you were thinking?
0a904a1e8598464ad07238e985a4b12a378c7f1b

joshsadam:
> I think I was trying to copy the Select2Component format. Is this closer to what you were thinking? [0a904a1](https://github.com/phac-nml/irida-next/commit/0a904a1e8598464ad07238e985a4b12a378c7f1b)

Definitely moving in the right direction (again double check with @ericenns).  What still needs to be addressed:

1. JS still initializes from label text, not `data-value`. On connect it still does filter-by-visible-text → first match → `#setValue`. Duplicate labels can still pick the wrong option and overwrite the hidden value.
2. Callers must pass both `selected_name` and `selected_value`. That keeps two sources of truth at the call site instead of resolving the label from the option that matches `selected_value`.
3. No label fallback when the option block is empty — still `content.presence` only.
4. Slot options have no `id` (the Nokogiri path that assigned ids was removed), so aria-activedescendant is weaker for the slots path.

ksierks:
>> I think I was trying to copy the Select2Component format. Is this closer to what you were thinking? [0a904a1](https://github.com/phac-nml/irida-next/commit/0a904a1e8598464ad07238e985a4b12a378c7f1b)

>Definitely moving in the right direction (again double check with @ericenns).  What still needs to be addressed:

>1. JS still initializes from label text, not `data-value`. On connect it still does filter-by-visible-text → first match → `#setValue`. Duplicate labels can still pick the wrong option and overwrite the hidden value.
 02c7a4cb07d12d1c7871a90fdd52a7a6f112e71a
>2. Callers must pass both `selected_name` and `selected_value`. That keeps two sources of truth at the call site instead of resolving the label from the option that matches `selected_value`.
f6f581f9bb36ba2c9d25c8f1cf787aefe8593177
>3. No label fallback when the option block is empty — still `content.presence` only.
8c81372fd7bddcdc058c2916f216b48a2b98ff76
>4. Slot options have no `id` (the Nokogiri path that assigned ids was removed), so aria-activedescendant is weaker for the slots path.
235689b7bfc057992eedccdc03db7156fe244272

joshsadam:
Thanks for breaking these into separate commits. Points 1–3 now look addressed.

Point 4 adds the required IDs, but create_slot_listbox renders the slot components with `options.join`, reparses that HTML with Nokogiri, mutates it, and serializes it again. That restores the render → parse → mutate path from my original comment.

Could the IDs instead be assigned to each OptionComponent before it renders? That would let the template render the slots directly and keep Nokogiri limited to the legacy options: adapter.

I would also add a focused test proving an option without a block falls back to its label. The current slot preview always supplies content, so it does not exercise that fallback.

ksierks:
> Thanks for breaking these into separate commits. Points 1–3 now look addressed.
> 
> Point 4 adds the required IDs, but create_slot_listbox renders the slot components with `options.join`, reparses that HTML with Nokogiri, mutates it, and serializes it again. That restores the render → parse → mutate path from my original comment.
> 
> Could the IDs instead be assigned to each OptionComponent before it renders? That would let the template render the slots directly and keep Nokogiri limited to the legacy options: adapter.
> 
> I would also add a focused test proving an option without a block falls back to its label. The current slot preview always supplies content, so it does not exercise that fallback.

Sorry, I didn't know how else to do it. See ddf90dd09c67f23ba3f1dde13e494d3a3829baf7 for the option ids & 4143880ab7edbaf8e82f902766b0097b44cadebc for the option label fallback test.

joshsadam:
> Sorry, I didn't know how else to do it. See [ddf90dd](https://github.com/phac-nml/irida-next/commit/ddf90dd09c67f23ba3f1dde13e494d3a3829baf7) for the option ids & [4143880](https://github.com/phac-nml/irida-next/commit/4143880ab7edbaf8e82f902766b0097b44cadebc) for the option label fallback test.

This is really good, thanks for updating it.  I am heading out on vacation for the next 2 weeks so I will leave the final review to @ericenns

app/components/combobox/v1/component.rb
```diff
@@ -3,17 +3,30 @@
...
+      def initialize(form:, field:, options: nil, selected_value: nil, **combobox_arguments)
         @combobox_id = form.field_id(field)
         @listbox_id = "#{form.field_id(field)}_listbox"
         @form = form
         @field = field
-        @listbox_options = create_listbox(options)
-        @selected_option = selected_option(options)
+        @selected_value = selected_value
+
+        unless options.nil?
```
joshsadam, line 17: Issue: Rendering with no HTML `options` and zero `with_option` slots leaves `@selected_option` unset, so `combobox_arguments` raises `NoMethodError` on `@selected_option[:name]`. Before this PR `options:` was required, so that state was unreachable. For Select2 migration call sites that can legitimately have an empty collection, could we initialize `@selected_option = { name: '', value: '' }` in `initialize` so empty slots stay safe?

app/javascript/controllers/combobox/v1_controller.js
```diff
@@ -95,10 +95,10 @@ export default class extends Controller {
     this.#addListboxEventListeners(this.listboxTarget);
 
     // Initialize
-    if (this.#filter) {
-      const option = this.#filterOptions();
-      this.#setOption(option);
-      this.#setValue(option);
+    const initialOption = this.#findOptionByValue(this.hiddenTarget.value);
```
joshsadam, line 98: Issue: `connect` no longer calls `#setOption`, so a pre-selected value never gets `aria-selected` / `aria-activedescendant`. `#onComboboxFocus` also discards `#filterOptions()`'s return value, so `#option` stays `null` and `#populateCurrentFirstLastOptions` falls back to `#firstOption`. With duplicate labels, ArrowDown then a background click can commit the first text match and overwrite the hidden `data-value` (the new initialize test covers `connect`, but not reopen). Could we resolve the current row with `#findOptionByValue(this.hiddenTarget.value)` when rebuilding the list and prefer that over `#firstOption`?

app/javascript/controllers/combobox/utils.js
```diff
@@ -18,21 +18,61 @@ export function highlightOption(option, filter) {
   if (!filter) {
```
joshsadam, line 21: Issue: Filtering still uses `getLowercaseContent` (`textContent`), but `#setValue` now writes `data-label` into the input. When slot content's visible text differs from `label` — the main reason for slotted options — refocusing filters by a string that may not appear in the rendered text and can show "No results found" for a valid selection. Could we match against `data-label` as well as `textContent` so filter and selection stay consistent?

app/components/combobox/v1/component.rb
```diff
@@ -43,14 +56,25 @@ def combobox_disabled?
         @combobox_arguments[:disabled] == true
       end
 
-      def selected_option(options)
+      def selected_option(options, selected_value)
         fragment = Nokogiri::HTML.fragment(options)
         fragment.search('option').each do |option|
-          return { name: option.text, value: option['value'] } if option.key?('selected')
+          if option.key?('selected') || option['value'] == selected_value
```
joshsadam: Issue: On the legacy HTML `options` path, Nokogiri always gives string `option['value']`, so `selected_value: 3` never matches and the field renders with no selection. The slot path already coerces with `.to_s` in `selected_slot_option`. Could we use `selected_value.present? && option['value'].to_s == selected_value.to_s` here too (with `present?` so a nil `selected_value` does not match a blank prompt option)?

app/components/combobox/v1/component.rb
```diff
@@ -80,15 +104,28 @@ def create_listbox_grouped_options(fragment) # rubocop:disable Metrics/MethodLen
...
           listbox_group_option['data-value'] = option['value']
+          listbox_group_option['data-label'] = option.text
           listbox_group_option['aria-disabled'] = 'true' if option.key?('disabled')
           listbox_group_option.content = option.text
           option.replace(listbox_group_option)
         end
         fragment
       end
+
+      def assign_option_ids(option_nodes)
```
joshsadam: Nit: `assign_option_ids` looks unused (HTML-path IDs are set in `create_listbox_options`, and slots assign IDs in the template). Could we remove it if it was leftover from an earlier approach?

app/components/combobox/v1/component.html.erb
```diff
@@ -60,7 +58,14 @@
       "aria-label": combobox_arguments[:aria][:label],
       "data-combobox--v1-target": "listbox",
     ) do %>
-      <%= @listbox_options %>
+      <% if options? %>
+        <% options.each_with_index do |option, option_index| %>
+          <% option.id ||= listbox_option_id(option_index) %>
```
joshsadam: Nit: Assigning `option.id` in the template works, but we already use `before_render` for slot selection. Moving ID assignment there would keep the template declarative and colocate selection + ID resolution — e.g. `options.each_with_index { |option, i| option.id ||= listbox_option_id(i) }` before `selected_slot_option`.

joshsadam (changes requested):
This is coming along really well.  Just a few things to take a look at.  @ericenns can you finish the review while I am away?

app/javascript/controllers/combobox/utils.js
ksierks: I tried updating the preview to only use labels and filter the results. `getLowercaseContent` returns the textConent of the option node, which is the OptionComponent's content or label. I believe content & label are both highlighting & filtering as expected.

app/components/combobox/v1/component.html.erb
ksierks: 15adcae0a21c056f4db80f7271eac8ef8ac51f39

app/components/combobox/v1/component.rb
ksierks: 2386269eb44a6b76483a531b98087467db8eeb22

app/components/combobox/v1/component.rb
ksierks: 0d62e8d86e65a2849ff5577d7fe0424088d8904d

app/components/combobox/v1/component.rb
ksierks: b9a7b453cd3284ef31d910d05637d21db5766671

app/javascript/controllers/combobox/v1_controller.js
ksierks: Like this, 13ae381712854f363147e0c093eccc155bbc2df5?

app/javascript/controllers/combobox/v1_controller.js
joshsadam, line 1: Blocking: filtering ignores `data-label` for slotted options

Selection and the input display use `data-label`, but `#filterOptions` searches only rendered text. When a slot label differs from its content, initialization and focus filter on the label, remove the selected option, and show no results. Please search `data-label`, with `textContent` as the fallback, so filtering uses the same label the combobox displays.

test/javascript/controllers/combobox/v1_controller.test.js
```diff
@@ -375,4 +393,27 @@ describe("combobox v1 controller", () => {
     expect(combobox()).toHaveAttribute("aria-disabled", "true");
     expect(combobox()).not.toHaveAttribute("disabled");
   });
+
+  it("initializes from hidden data-value when labels are duplicated", async () => {
```
joshsadam, line 397: Major: add regression coverage for differing labels and slot content

Please cover an option whose `data-label` differs from its visible slot text. Verify that initialization or focus preserves the selected value and does not show the no-results state. This will ensure filtering continues to use the label displayed by the combobox.

joshsadam (commented):
Just a couple more things of note.  Thanks for making the other changes

test/javascript/controllers/combobox/v1_controller.test.js
ksierks: 3923c4adce0e8ade8757f6e7cda6e229fc522bd0

app/javascript/controllers/combobox/v1_controller.js
ksierks: 3923c4adce0e8ade8757f6e7cda6e229fc522bd0

test/components/combobox/v1/option_component_test.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module Combobox
+  module V1
+    class OptionComponentTest < ViewComponentTestCase
+      test 'renders label when no content is provided' do
+        render_inline(OptionComponent.new(value: 'user-1', label: 'User 1'))
+
+        assert_selector '[role="option"][data-value="user-1"][data-label="User 1"]', text: 'User 1'
+      end
+    end
```
joshsadam, line 35: Suggestion: Could we add OptionComponent examples for `disabled: true`, block content that differs from `label`, and an explicit `id`? Those branches are part of the new slot API (`content.presence || @label`, `aria-disabled`, and `id ||=` in the parent), and the current test only covers the label fallback.

test/components/combobox/v1/component_render_test.rb
```diff
@@ -39,6 +39,13 @@ class ComponentRenderTest < ViewComponentTestCase
         assert_selector 'button[data-combobox--v1-target="indicatorButton"][tabindex="-1"]'
       end
 
+      test 'renders safely with no html options and no slot options' do
+        render_component(options: nil)
+
+        assert_selector 'input[role="combobox"][value=""]'
+        assert_selector "##{build_form_builder.field_id(:field)}_hidden[value='']", visible: :hidden
+      end
+
```
joshsadam, line 56: Suggestion: Could we add a render test that passes HTML `options` together with `selected_value` (and no `selected` attribute)? `selected_option` now matches by value as well as the `selected` attribute, and that path is untested. Callers that omit `selected_value` should behave as before; this would lock in the dual HTML + `selected_value` API.

app/javascript/controllers/combobox/utils.js
```diff
@@ -11,28 +11,70 @@ export function isComboboxDisabled(combobox) {
 }
 
 export function getLowercaseContent(node) {
-  return node.textContent.toLowerCase();
+  const label = node.getAttribute?.("data-label");
```
joshsadam, line 14: Nit: When `data-label` is present, `getLowercaseContent` uses it and ignores slot text. That matches the slot design (search the label, not richer markup). A short comment here would help the next reader avoid treating `textContent` as searchable.

joshsadam (changes requested):
Just a couple more things I picked up on.

app/javascript/controllers/combobox/utils.js
ksierks: 7bf3d7dd8a6c2d89737d85906002aa85af33f416

test/components/combobox/v1/option_component_test.rb
ksierks: d272de5578bfdb108b8b66f5e470994c3f6ce09e

test/components/combobox/v1/component_render_test.rb
ksierks: The test `with_slots preview renders slot options` passes HTML options together with a selected_option.

joshsadam: Ah that one is covering slots + `selected_value`, not the html `options` path. What I was hoping for is html options with no `selected` attribute, plus `selected_value`, so we lock in that `selected_option` can match by value.

Something like:

```ruby
test 'selects html option by selected_value when selected attribute is missing' do
  html = options_for_select([['User 1', '1'], ['User 2', '2']])
  render_component(options: html, selected_value: '2')

  assert_selector 'input[role="combobox"][value="User 2"]'
  assert_selector "##{build_form_builder.field_id(:field)}_hidden[value='2']", visible: :hidden
end
```

ksierks: Oh, I see. 192b6b8730b276597a6677faf788b155f780a429

joshsadam (dismissed):

app/components/combobox/v1/component.html.erb
```diff
@@ -28,7 +28,6 @@
       data-combobox--v1-target="indicatorClearButton"
       data-action="mousedown->combobox--v1#onIndicatorMouseDown click->combobox--v1#onClearClick"
       aria-label="<%= t('combobox_component.clear_selection') %>"
-      title="<%= t('combobox_component.clear_selection') %>"
```
ericenns, line 31: nitpick (blocking): Can we keep title for now as it helps to provide a better explanation instead of just the icon for non screenreader users. (May need to disable an ERB rule to allow this as `aria-label` just helps screenreader users so for now both should be kept.

app/components/combobox/v1/component.html.erb
```diff
@@ -45,7 +44,6 @@
       data-combobox--v1-target="indicatorButton"
       data-action="mousedown->combobox--v1#onIndicatorMouseDown click->combobox--v1#onIndicatorClick"
       aria-label="<%= t('combobox_component.show_options') %>"
-      title="<%= t('combobox_component.show_options') %>"
```
ericenns, line 48: nitpick (blocking): Can we keep title for now as it helps to provide a better explanation instead of just the icon for non screenreader users. (May need to disable an ERB rule to allow this as `aria-label` just helps screenreader users so for now both should be kept.

ericenns (changes requested):
Looking good, just a comment on keeping `title` for now. In the future we may change to a tooltip which would be readable to both screen reader and non screen reader users.

app/components/combobox/v1/component.html.erb
ksierks: I hope this is ok, fa17fa50d3695ed06865c1e7db49310d44c40f74.

app/components/combobox/v1/component.html.erb
ksierks: I hope this is ok, fa17fa50d3695ed06865c1e7db49310d44c40f74.

ericenns (approved):
Looks good to me!

PR #2104, opened 2026-07-21
Advanced search: Add exists to metadata type operators

ChrisHuynh333:
## What does this PR do and why?
This PRs adds metadata type specific operators for exists `date/numeric_exists/not_exists`

## Screenshots or screen recordings
_Screenshots are required for UI changes, and strongly recommended for all other pull requests._

## How to set up and validate locally
1. Create a project with 4 samples with metadata like so:

| Sample Name  | numeric_field| date_field |
| ------------- | ------------- | ------------- |
| sample1 | 100 | 2026-01-01 |
| sample2 | 100.1 | 01-01-2026 |
| sample3 | abc | abc |
| sample4 (don't add metadata) | nil | nil |

2. Start an advanced search and verify the `exists/not_exists` operators do not render the value field
3. Create an advanced search with each of the new exist/not_exist operators, and expect the following results:

| Sample Name  | exists on numeric_field | not_exists on numeric_field | numeric_exists  on numeric_field| numeric_not_exists on numeric_field | date_exists on date_field | date_not_exists on date_field |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| sample1 | Y | N | Y | N | Y | N |
| sample2 | Y | N | Y | N | N | Y |
| sample3 | Y | N | N | Y | N | Y |
| sample4 | N | Y | N | Y | N | Y |

4. Test the above through GraphQL queries and verify you receive the same results

## PR acceptance checklist
This checklist encourages us to confirm any changes have been analyzed to reduce risks in quality, performance, reliability, security, and maintainability.

- [x] I have evaluated the [PR acceptance checklist](https://phac-nml.github.io/irida-next/docs/development/development_processes/code_review#acceptance-checklist) for this PR.

PR #2115, opened 2026-07-22
Advanced Search: Between Operators

ChrisHuynh333:
This PR is for issue [STRY0020705](https://publichealthprod.service-now.com/rm_story.do?sys_id=7a97b3563bbc8f5025838a0f23e45a54) and adds the between operators (`between`, `date_between`, `text_between`, and `numeric_between`) to the Advanced Search UI.

Note: GraphQL will come in a later PR

[Screencast from 2026-07-23 13-58-05.webm](https://github.com/user-attachments/assets/d0f0ac41-4eb2-4581-885e-d372ad0f368e)

Error handling:
Blank values:
image

`date_between` not receiving date:
image

`numeric_between` not receiving numbers:
image

Standard date fields (like `created_at`) still require date inputs:
image

1. Navigate to a samples page and initiate advanced searches using the standard `between` operator, and metadata `date_between`, `numeric_between` and `text_between` operators, and verify the filtering works as expected.
2. Test the above error handling cases and verify errors occur as expected

## PR acceptance checklist

### DIFF
--- a/app/javascript/controllers/advanced_search/v1_controller.js
+++ b/app/javascript/controllers/advanced_search/v1_controller.js
@@ -207,7 +207,7 @@ export default class AdvancedSearchController extends Controller {
           .replace(/CONDITION_INDEX_PLACEHOLDER/g, conditionIndex);
 
         const updatedCondition = this.#conditionElements(group)[conditionIndex];
-        const updatedValue = updatedCondition?.querySelector(".value");
+        const updatedValue = this.#getValueInput(updatedCondition);
         updatedValue?.classList.remove(...this.#hiddenClasses);
         this.#updateValueFieldForEnum(
           updatedValue,
@@ -592,6 +592,7 @@ export default class AdvancedSearchController extends Controller {
   }
 
   #getValueInput(condition) {
+    if (!condition) return null;
     const values = condition.querySelectorAll(".value");
     if (values.length === 0) {
       return null;
--- a/app/javascript/controllers/advanced_search/v1_controller.js
+++ b/app/javascript/controllers/advanced_search/v1_controller.js
@@ -13,6 +13,7 @@ export default class AdvancedSearchController extends Controller {
     "searchGroupsTemplate",
     "selectValueTemplate",
     "valueTemplate",
+    "betweenValueTemplate",
   ];
   static outlets = ["list-input"];
   static values = {
@@ -179,7 +180,7 @@ export default class AdvancedSearchController extends Controller {
       return;
     }
 
-    const value = condition.querySelector(".value");
+    const value = this.#getValueInput(condition);
     const groupIndex = this.#groupElements().indexOf(group);
     const conditionIndex = this.#conditionElements(group).indexOf(condition);
     if (!value || groupIndex < 0 || conditionIndex < 0) {
@@ -215,14 +216,14 @@ export default class AdvancedSearchController extends Controller {
           operator,
         );
       } else {
-        const templateTarget = [
-          "in",
-          "not_in",
-          "text_in",
-          "text_not_in",
-        ].includes(operator)
-          ? this.listValueTemplateTarget
-          : this.valueTemplateTarget;
+        let templateTarget;
+        if (["in", "not_in", "text_in", "text_not_in"].includes(operator)) {
+          templateTarget = this.listValueTemplateTarget;
+        } else if (operator.includes("between")) {
+          templateTarget = this.betweenValueTemplateTarget;
+        } else {
+          templateTarget = this.valueTemplateTarget;
+        }
         value.outerHTML = templateTarget.innerHTML
           .replace(/GROUP_INDEX_PLACEHOLDER/g, groupIndex)
           .replace(/CONDITION_INDEX_PLACEHOLDER/g, conditionIndex);
@@ -254,7 +255,7 @@ export default class AdvancedSearchController extends Controller {
     condition.dataset.advancedSearchSelectedField = selectedField;
     this.#updateOperatorDropdown(condition, selectedField);
 
-    const value = condition.querySelector(".value");
+    const value = this.#getValueInput(condition);
     if (value) {
       this.#clearValueInputs(value);
       value.classList.add(...this.#hiddenClasses);
@@ -588,5 +589,18 @@ export default class AdvancedSearchController extends Controller {
         element.selectedIndex = -1;
       }
     });
+  }
+
+  #getValueInput(condition) {
+    const values = condition.querySelectorAll(".value");
+
+    if (values.length === 0) {
+      return null;
+      // handles removing second input for between operator
+    } else if (values.length === 2) {
+      values[1].remove();
+    }
+
+    return values[0];
   }
 }
--- a/app/javascript/controllers/advanced_search/v1_controller.js
+++ b/app/javascript/controllers/advanced_search/v1_controller.js
@@ -172,6 +172,7 @@ export default class AdvancedSearchController extends Controller {
   }
 
   handleOperatorChange(event) {
+    console.log("change");
     const operator = event.target.value;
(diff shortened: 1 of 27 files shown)


app/javascript/controllers/advanced_search/v1_controller.js
ksierks: Should this use `#getValueInput` too?

app/models/concerns/advanced_search/operators/between_operators.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+      end
+
+      def condition_numeric_between(scope, node, value)
+        casted_node = Arel::Nodes::NamedFunction.new(
+          'CAST',
+          [node.as(Arel::Nodes::SqlLiteral.new('DOUBLE PRECISION'))]
+        )
+
+        scope.where(node.matches_regexp('^-?\\d+(\\.\\d+)?$').and(
```
ksierks, line 33: I think this regex is used in other parts of advanced search, like in metadata_comparison.rb. Could this regex be extracted into a constant and reused?

app/models/concerns/advanced_search/operators/between_operators.rb
```diff
@@ -0,0 +1,56 @@
...
+
+module AdvancedSearch
+  module Operators
+    # Methods for between-based search conditions (between, date_between, numeric_between, text_between)
+    module BetweenOperators
+      extend ActiveSupport::Concern
+
+      private
+
+      def condition_between(scope, node, value, field_name)
```
ksierks, line 11: Should we validate value[0] > value[1] for the between range? I was able to search an invalid range.

ChrisHuynh333: I've added validation so the `From` value must be lower than the `To` value.
There will be awkwardness if the user compares some values like numbers to letters or numbers when using between/text_between, but after talking to Eric, we decided as long as the SQL can still query, we won't block the search via validation

331c6fa53fd0c86347e077c884a5ce6b5f0c69ca 35e774215d1c6a2ca15edee197c6c141e4a00acf

app/models/concerns/advanced_search/operators/between_operators.rb
ChrisHuynh333: I've extracted this in my other PR https://github.com/phac-nml/irida-next/pull/2104

There's also plans on centralizing a lot of the advanced search constants, so likely this will be further handled in a PR much further along

app/javascript/controllers/advanced_search/v1_controller.js
ChrisHuynh333: I'm going to leave this as is for now if that's okay with you. Although I don't think the `?.` guard will be triggered, I'll leave it in place, it won't need the second input deletion currently.
Like the comment below, the choosing of templates was going to be part of the overhaul as well so this may get refactored in a future PR

app/models/concerns/advanced_search/operators/between_operators.rb
ksierks: Thanks! I think this only make sense to do it for date_between & numeric_between.

ChrisHuynh333: I think we still want to validate the user is comparing text properly (abc to xyz).

I realized I didn't factor in upper/lower cases during validation (error if comparing a to Z), which should be fixed in ac76feeaf715217d1728d448306cb812d6dd2ead c2ba893b8f3126b44d326cc2d25b74a808247212

app/javascript/controllers/advanced_search/v1_controller.js
ChrisHuynh333: I realized this actually may not get refactored, so I've updated to use `getValueInput`, and then added the guard within `getValueInput`
051f577fd98b3ead93b3bd538426868eff7a6798

ksierks (dismissed):
Looks good to me.

app/javascript/controllers/advanced_search/v1_controller.js
```diff
@@ -589,4 +589,17 @@ export default class AdvancedSearchController extends Controller {
...
   }
+
+  #getValueInput(condition) {
+    if (!condition) return null;
+    const values = condition.querySelectorAll(".value");
+    if (values.length === 0) {
+      return null;
+      // handles removing second input for between operator
+    } else if (values.length === 2) {
+      values[1].remove();
```
ericenns, line 600: nitpick: Why is this removing the input? It may make more sense to rename this method entirely since it is not only getting the input but also removing an input.

Instead couldn't the ValueComponent be refactored to wrap the input in a wrapper div and then instead of changing outerHtml when needing to change the innerHtml. Then this method could actually return the valueWrapper without having to remove elements making it a true getter method.

app/models/advanced_search_condition.rb
```diff
@@ -11,6 +11,8 @@ class AdvancedSearchCondition
   attribute :operator
   attribute :value
 
+  attr_accessor :value
```
ericenns: nitpick(blocking): I don't believe this line is necessary as the line above already handles it.

app/validators/advanced_search/group_validator.rb
```diff
@@ -220,5 +249,23 @@ def validate_between(unique_field_condition, common_field_conditions)
     def metadata_field?(field)
       METADATA_FIELD_PATTERN.match?(field)
     end
+
+    def validate_between_values(condition) # rubocop:disable Metrics/AbcSize
```
ericenns, line 262: nitpick(blocking): Can you not add an error here when one value is correct format and the other value is not the correct format. Currently if use `date_bewtween` operator and provide a valid from value (e.g. `1979-01-01`) and then just provide " " (single space) as the to value I get error error messages on both fields when only the to value should be showing the format error.

ericenns (changes requested):

app/validators/advanced_search/group_validator.rb
ChrisHuynh333: I've stopped validation if a field is blank:
image

The validator will still go through the format validator. Let me know if you want me to skip over that too 
cff35b8c9a8046e260c8b07f72c38f689eb868a5

app/models/advanced_search_condition.rb
ChrisHuynh333: Removed cff35b8c9a8046e260c8b07f72c38f689eb868a5

app/javascript/controllers/advanced_search/v1_controller.js
ChrisHuynh333: I've renamed this to `#resetAndGetValueInput`, and only call it in a couple places where the value does need resetting.
The removal is to remove the second input of the between operator.
I realized this caused a bug when going from between operator to exist operator, since the current logic doesn't have exist value overwrite the current value input field (eg: exist value would submit with `from value` from the pre-existing between operator)., so I've refactored `blank` and `exist` operator change to now share logic and overwrite the exist operator, and then hide the input.

I played around with wrapping the value input with a wrapper div, but that had a bunch of unexpected bugs, even unrelated to the above. So I left it without the wrapper. If you want though, I can deep dive and try to fix everything to work with the wrapper.

8b3be736cc23bb467ff3fbb7acb47501dfe72b29
90523270780ff7994f03e74362fed95a928360b3
4ed50f0ca693c1d23753e69cd4b001b406861093

app/validators/advanced_search/group_validator.rb
ericenns: Yeah we could skip the format validator if the value is blank.

ChrisHuynh333: Added in 5a8d0e51b03ff57e10bcb34762272af5c831f8b5

ericenns (approved):
Looks good to me!

PR #2122, opened 2026-07-27
feat: ResetSamplesCountJob

ericenns:

Added in `ResetSamplesCountJob` which recalculates Project and Group `samples_count` columns. Optionally an array of `root_group_ids` can be passed which limit the joib to only process Projects and Groups that are descendants of the passed in `root_group_ids`, otherwise all Projects and Groups are processed. This job is idempotent and can be resumed from interruption without reprocessing already processed Projects and Groups.

1. Ensure services are running (e.g. `devenv up`)
2. Run targetted tests with `bin/rails test test/jobs/reset_samples_count_job_test.rb`
3. Expect all tests to pass
4. Optionally run broader test suite
5. Optionally run job from rails console

## PR acceptance checklist

### DIFF
--- a/app/jobs/reset_samples_count_job.rb
+++ b/app/jobs/reset_samples_count_job.rb
@@ -61,23 +61,14 @@ class ResetSamplesCountJob < ApplicationJob
   def scoped_projects(root_groups)
     return Project.all if root_groups.nil?
 
-    descendant_project_namespace_ids = descendant_namespace_ids(root_groups,
-                                                                Namespaces::ProjectNamespace.sti_name)
+    group_ids = scoped_groups(root_groups).select(:id)
 
-    Project.where(namespace_id: descendant_project_namespace_ids)
+    Project.joins(:namespace).where(namespace: { parent_id: group_ids })
   end
 
   def scoped_groups(root_groups)
     return Group.all if root_groups.nil?
 
-    descendant_group_ids = descendant_namespace_ids(root_groups, Group.sti_name)
-
-    Group.where(id: descendant_group_ids)
-  end
-
-  def descendant_namespace_ids(root_groups, type)
-    root_groups.find_each.flat_map do |root_group|
-      root_group.self_and_descendants_of_type(type).pluck(:id)
-    end.uniq
+    root_groups.self_and_descendants
   end
 end
--- a/app/jobs/reset_samples_count_job.rb
+++ b/app/jobs/reset_samples_count_job.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+# Recalculates persisted sample counters for projects and groups.
+#
+# Order is important:
+# 1) reset project counters from actual sample rows
+# 2) recalculate group counters from descendant projects
+class ResetSamplesCountJob < ApplicationJob
+  include ActiveJob::Continuable
+
+  queue_as :default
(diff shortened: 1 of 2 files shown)


app/jobs/reset_samples_count_job.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    step :reset_project_counts_step, start: 0
+    step :reset_group_counts_step, start: 0
+    step :log_completion
+  end
+
+  private
+
+  def reset_project_counts_step(step)
+    start_id = scoped_projects(@root_groups).order(:id).offset(step.cursor).first&.id
```
ksierks, line 31: Does it make sense to refactor, so the expensive call to `scoped_projects` is only called once?

app/jobs/reset_samples_count_job.rb
```diff
@@ -0,0 +1,83 @@
...
+    return if start_id.nil?
+
+    scoped_projects(@root_groups).find_each(start: start_id) do |project|
+      Project.reset_counters(project.id, :samples)
+      step.advance!
+    end
+  end
+
+  def reset_group_counts_step(step)
+    start_id = scoped_groups(@root_groups).order(:id).offset(step.cursor).first&.id
```
ksierks, line 41: Does it make sense to refactor, so the expensive call to `scoped_groups` is only called once?

app/jobs/reset_samples_count_job.rb
ericenns: Copilot had suggested the same thing to me, but on investigation it really is not the case that it is expensive.

e.g. When not passing `root_group_ids`

Image

As you can see, one query to get the first id and then a query to get the first 1000 projects starting from the first id.

e.g. When passing `root_group_ids`

Image

As you can see, this one has a few more queries that are replicated but they aren't that expensive.

But with some refactoring I have reduced the queries when passing `root_group_ids` in 984bd0c2b7536aa824cfbcfd5e465308c1b94b56

e.g.

Image

app/jobs/reset_samples_count_job.rb
ericenns: See reply [here](https://github.com/phac-nml/irida-next/pull/2122/changes#r3659263494)

ksierks (approved):

test/jobs/reset_samples_count_job_test.rb
```diff
@@ -0,0 +1,240 @@
...
+    end
+
+    first_group.reload
+
+    # First group should have been reset to correct count
+    assert_equal first_group_expected_sample_count, first_group.samples_count,
+                 'First group should have been reset during first execution'
+
+    # Now manually set the first group back to an incorrect count to verify it won't be reset again
+    first_group.update_columns(samples_count: 777) # rubocop:disable Rails/SkipsModelValidations
```
JeffreyThiessen, line 165: This is just a comment not a request for change. 

I feel like there should be a way to test this functionality without creating an invalid database state within the test. This test does check what we want but maybe there is a better way to test it like mocking either the group list or the Project calls. It's probably fine as is but something to think about on how we write test cases for jobs that update the database like this.

JeffreyThiessen (approved):
Code looks good and tests cover steps correctly to verify idempotency. 👍

PR #2123, opened 2026-07-27
Fix: Attachment Dialog Listing

ChrisHuynh333:
This PR is for [STRY0021058](https://publichealthprod.service-now.com/now/nav/ui/classic/params/target/rm_story.do%3Fsys_id%3D606dcd313b8e0b105c163064c3e45afa%26sysparm_view%3Dscrum%26sysparm_domain%3Dnull%26sysparm_domain_scope%3Dnull) and fixes an issue where if more than 20 sample attachments were selected for deletion, the deletion dialog only listed the 20 current shown on the page.
Now, all selected attachments are shown in the dialog listing.

The story only discusses the `Delete Files` dialog, but the issue was also in the `Concatenate Files` dialog, so that has been fixed as well.

Notes:
- PE files are separated in the dialogs as separate line items, and count as 2 attachments.
- `form--json-submission` logic used in other mass selection actions was not implemented here. This is because handling PE files as a stringified array of 2 values requires additional/differing logic.

Delete Files:

[Screencast from 2026-08-14 14-56-07.webm](https://github.com/user-attachments/assets/164f6401-8fe4-45e2-be47-a22ddfb8718c)

Concatenate Files:

[Screencast from 2026-08-14 15-25-54.webm](https://github.com/user-attachments/assets/ffac30c6-bcac-4af7-a4c6-a513ec3ca7dd)

Delete Files:
1. Navigate to sample files page, and upload many files (over 50), including some PE files.
2. Select the files for deletion, and verify the `infinite-scroll` works and scrolling to the bottom of the listing initiates a fetch for the remaining files to be listed.
3. Also verify the attachment count in the dialog is correct (PE files counting as 2).
4. Delete the files and verify deletion still behaves as expected. 

Concatenate Files:
1. Similarly test the concatenate files dialog attachments listing like above and verify the `infinite-scroll` works.
2. Test erroring the dialog with improper submission(s) (wrong file types, no basename), and verify the dialog still re-renders with the error and the attachment listing re-fetches
3. Concatenate Files successfully and verify the concatenation behaves as expected

## PR acceptance checklist


app/controllers/concerns/list_actions.rb
```diff
@@ -4,12 +4,24 @@
 module ListActions
   extend ActiveSupport::Concern
 
-  def list
+  def list # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
joshsadam: Issue: The new `attachment` branch looks up records by client-supplied IDs with no `authorize!` and no scope to `@sample`. This endpoint returns filenames and PUIDs, and other actions on `Projects::Samples::AttachmentsController` authorize explicitly.

Because `list` is shared with sample/workflow-execution callers, adding `authorize!` here would change those endpoints too. Could we authorize and scope in this controller instead (for example `authorize! @project, to: :read_sample?` and `@sample.attachments.where(id: flattened_attachment_ids)`) so we only return files on this sample?

config/locales/fr.yml
```diff
@@ -1704,6 +1704,10 @@ fr:
             basename_help: Le nom de fichier de base ne peut contenir que des lettres, des chiffres, '-', '.' et '_'
             success: Les dossiers ont été concaténés avec succès.
           modal:
+            description:
+              plural: 'Les 4 fichiers suivants ont été sélectionnés pour la concaténation :'
```
joshsadam: Issue: `#replaceCountPlaceholder` in `infinite_scroll_controller.js` substitutes `COUNT_PLACEHOLDER` (same as English and the French deletion plural). This string hardcodes `4`, so the concatenate dialog will always say four files in French.

Could we use `Les COUNT_PLACEHOLDER fichiers suivants ont été sélectionnés pour la concaténation :` to match `deletions.modal.description.plural`?

app/views/projects/samples/attachments/deletions/_modal.html.erb
```diff
@@ -1,32 +1,59 @@
...
+      role="status"
+      id="spinner"
+      class="
+        absolute top-2/4 left-1/2 grid hidden h-full w-full -translate-x-1/2 -translate-y-1/2 place-items-center
+        backdrop-blur-sm
+      "
+    >
+      <div class="grid place-items-center">
+        <%= icon :faded_spinner, library: "animated" %>
+        <span class="text-black dark:text-white"><%= t(:".spinner") %>.</span>
```
joshsadam: Nit: The extra `.` after `t(:".spinner")` sits outside the locale string. Both `en.yml` and `fr.yml` already end with ellipses, so the UI will show a trailing period after the copy. Could we drop the `.` here, or move the punctuation into the locale strings if we want it?

app/controllers/attachments_controller.rb
```diff
@@ -8,8 +8,8 @@ class AttachmentsController < ApplicationController
 
   layout 'attachment'
 
-  before_action :set_attachment
-  before_action :set_context_crumbs
+  before_action :set_attachment, only: %i[show]
+  before_action :set_context_crumbs, only: %i[show]
```
joshsadam: Question: `AttachmentsController` is only routed for `show` (`config/routes/attachments.rb`). Restricting `set_attachment` / `set_context_crumbs` to `show` looks like a no-op, and the new `list` action lives on `Projects::Samples::AttachmentsController`.

Was this needed to avoid a before_action error, or can we leave these filters as they were?

app/controllers/concerns/list_actions.rb
```diff
@@ -4,12 +4,24 @@
...
+      # handles flattening attachment_ids where PE files come in as a nested array and need further parsing
+      # PE files are outputted as individual ids
+      flattened_attachment_ids = params[:attachment_ids].flat_map do |item|
+        parsed = JSON.parse(item)
+        parsed.is_a?(Array) ? parsed : item
+      rescue JSON::ParserError
+        item
+      end
+
+      @attachments = Attachment.where(id: flattened_attachment_ids)
```
joshsadam: Issue: `Attachment.where(id: …)` will N+1 when `_list_attachment` renders. Each row calls `associated_attachment` (`Attachment.find_by` per row, including unpaired files) and `attachment.file.blob.byte_size` (lazy blob load).

Could we eager-load blobs with `Attachment.with_attached_file.where(id: flattened_attachment_ids)` and, for the icon, use metadata so we skip the extra query? The pair record is unused here — we only need “is this PE?” and direction:

```erb
<% if attachment.metadata['associated_attachment_id'].present? && attachment.metadata['direction'] == 'forward' %>
  <%= icon(:arrow_right, size: :sm, class: "ml-0 mr-2") %>
<% elsif attachment.metadata['associated_attachment_id'].present? %>
  <%= icon(:arrow_left, size: :sm, class: "ml-0 mr-2") %>
<% else %>
  <%= icon(Attachment.icon, size: :sm, class: "ml-0 mr-2") %>
<% end %>
```

joshsadam (changes requested):
Found a couple things for you to look into

app/controllers/concerns/list_actions.rb
ChrisHuynh333: Let me know if this wasn't what you were thinking 37c15025762a8ffe445567b992cda93d1d25f402

config/locales/fr.yml
ChrisHuynh333: Thanks, tested with a real number for translation and missed setting it back to `COUNT_PLACEHOLDER` 37c15025762a8ffe445567b992cda93d1d25f402

app/views/projects/samples/attachments/deletions/_modal.html.erb
ChrisHuynh333: Removed 37c15025762a8ffe445567b992cda93d1d25f402

app/controllers/attachments_controller.rb
ChrisHuynh333: Removed 37c15025762a8ffe445567b992cda93d1d25f402

app/controllers/concerns/list_actions.rb
ChrisHuynh333: Fixed 37c15025762a8ffe445567b992cda93d1d25f402

app/views/projects/samples/attachments/deletions/_modal.html.erb
```diff
@@ -1,28 +1,59 @@
+<%# locals: (open:, sample:) -%>
+
 <%= render Viral::DialogComponent.new(open: open, size: :extra_large) do |dialog| %>
   <% dialog.with_header(title: t(".title")) %>
-  <%= turbo_frame_tag("deletion-alert") %>
 
-  <div class="mb-4 overflow-x-visible font-normal text-slate-500 dark:text-slate-400">
-    <p class="mb-4"><%= t(".description") %></p>
+  <div
+    data-controller="infinite-scroll"
```
joshsadam, line 7: Suggestion: Remove the now-unused `files_controller.js`.

This modal and the concatenation modal both switched to `data-controller="infinite-scroll"`, so nothing renders `data-controller="projects--samples--attachments--files"` anymore. You can confirm with:

```bash
grep -rn "samples--attachments--files" app/
```

It returns no hits, which means `app/javascript/controllers/projects/samples/attachments/files_controller.js` is dead code (it was the old approach that cloned table rows in the browser). Please delete it so we don't keep two implementations of the same feature:

```bash
git rm app/javascript/controllers/projects/samples/attachments/files_controller.js
```

Stimulus auto-registers controllers from that folder, so nothing else needs updating after the delete.

app/controllers/concerns/list_actions.rb
```diff
@@ -4,12 +4,25 @@
...
+    when 'sample'
       @samples = Sample.where(id: params[:sample_ids])
-    elsif params[:list_class] == 'workflow_execution'
+    when 'workflow_execution'
       @workflow_executions = WorkflowExecution.where(id: params[:workflow_execution_ids])
+    when 'attachment'
+      authorize! @project, to: :read_sample?
+      # handles flattening attachment_ids where PE files come in as a nested array and need further parsing
+      # PE files are outputted as individual ids
+      flattened_attachment_ids = params[:attachment_ids].flat_map do |item|
```
joshsadam: Suggestion: Avoid decoding the same paired-end (PE) encoding in two different places.

Selected PE files arrive as a flat list that mixes plain ids with stringified arrays, e.g. `["att_1", "att_2", "[\"fwd\",\"rev\"]"]`. Two separate pieces of code have to understand that exact shape:
- here, by calling `JSON.parse(item)` on each element and rescuing `JSON::ParserError`, and
- in `infinite_scroll_controller.js` (`#calculateNumSelected`), by counting `[` characters.

The risk: nothing keeps those two decoders in sync. If the format changes, one side can silently break, and `rescue JSON::ParserError` hides malformed input instead of surfacing it.

You don't need to redesign selection storage in this PR — just make the boundary explicit. Pick whichever is smaller:
1. Have the client send already-flattened `attachment_ids[]` plus a separate `pe_pair_count`, so the controller parses nothing and the JS counts nothing; or
2. Send the selection as a real JSON request body and `JSON.parse` it once.

If you keep the current shape for now, at minimum move this flattening into a small named method (e.g. `flatten_attachment_ids(params[:attachment_ids])`) with a comment describing the exact input format, so there is one documented owner of it.

app/controllers/concerns/list_actions.rb
```diff
@@ -4,12 +4,25 @@
 module ListActions
   extend ActiveSupport::Concern
 
-  def list
+  def list # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
joshsadam: `ListActions#list` is shared by samples and workflow_executions, whose branches do no authorization. The new `attachment` branch inlines `authorize! @project, to: :read_sample?`, reaches into `@sample`, and owns the id-flattening — all attachment/controller concerns living in a generic list mixin, guarded by a `rubocop:disable Metrics/AbcSize, Metrics/MethodLength`. The disable is the smell telling us the method is doing too much. Suggest extracting the attachment branch into a private `list_attachments` (or a small query/form object) in AttachmentsController and keeping the concern a thin dispatcher, so auth placement stays consistent with the other branches.

app/javascript/controllers/infinite_scroll_controller.js
```diff
@@ -36,6 +37,22 @@ export default class extends Controller {
...
+  #calculateNumSelected() {
+    let numSelected = this.allIds.length;
+
+    if (this.countPeAttachmentsValue) {
+      // pe files are saved as nested stringified 2 value arrays
+      // eg: ["att_1", "att_2", "["att_3_fwd", "att_3_rev"]"]
+      // we'll count how many opening brackets ([) exist, and add an additional count for each to represent the
+      // two PE files
+      numSelected += this.allIds.reduce((total, str) => {
+        return total + (str.match(/\[/g) || []).length;
```
joshsadam: Suggestion: This is the client-side half of the paired-end (PE) encoding I flagged on `ListActions#list`.

Counting `[` characters to work out how many PE pairs are selected is fragile: it assumes an id can never contain a `[`, and it breaks silently if the stored format changes. If we adjust the server contract so the count (or an already-clean list) comes from the backend, `#calculateNumSelected` can drop the regex entirely and just use `this.allIds.length` plus the provided pair count. Please keep this in sync with whatever we settle on for the controller.

test/integration/concerns/list_actions_test.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module Concerns
+  class ListActionsTest < ActionDispatch::IntegrationTest
+    test 'should list attachments' do
```
joshsadam, line 7: Suggestion: This test is a duplicate of the `should list attachments` test in `test/integration/attachments_preview_test.rb` (same fixtures, same POST, same single `assert_response :success`).

Two things to fix:

1. Keep only one copy. Since the `list` action lives in the `ListActions` concern, this file is the better home — please delete the duplicate from `attachments_preview_test.rb`.

2. `assert_response :success` on its own doesn't prove the behavior this PR adds (that the nested PE array is flattened and the right rows come back) — it would still pass if flattening were broken. Please assert on the rendered rows, e.g.:

```ruby
post list_namespace_project_sample_attachments_path(namespace, project, sample, format: :turbo_stream), params: {
  page: 1,
  attachment_ids: ["[#{pe_fwd_attachment.id}, #{pe_rev_attachment.id}]", non_pe_attachment.id],
  list_class: 'attachment'
}
assert_response :success
assert_match pe_fwd_attachment.puid, response.body
assert_match pe_rev_attachment.puid, response.body
assert_match non_pe_attachment.puid, response.body
```

app/views/shared/attachments/_list_attachment.html.erb
```diff
@@ -0,0 +1,36 @@
+<%# locals: (attachment:) -%>
+
+<tr class="border-b border-slate-200 py-2 text-slate-500 last:border-b-0 dark:border-slate-700 dark:text-slate-400">
+  <td class="px-3 py-2">
+    <%= attachment.puid %>
+  </td>
+
+  <td class="px-3 py-2">
+    <div class="flex items-center font-sans">
+      <% if attachment.metadata['associated_attachment_id'].present? && attachment.metadata['direction'] == 'forward' %>
```
joshsadam, line 10: Optional: This row/column markup largely repeats what `app/components/attachments/table_component.html.erb` already renders — the forward/reverse arrow icons, the format `Viral::PillComponent`, and the same set of columns. If it's straightforward, consider extracting the shared row into a single partial or subcomponent that both the table component and this listing can use, so the PE icon logic lives in one place.

Not blocking. And no need to switch the PE check to `attachment.associated_attachment` here — the current `attachment.metadata['associated_attachment_id'].present?` is actually preferable in a list because it avoids an extra DB lookup per row (`associated_attachment` runs a `find_by`).

app/views/shared/loading/_table_body.html.erb
```diff
@@ -0,0 +1,12 @@
+<%# locals: (column_count:) -%>
+
+<% 4.times do %>
+  <tr class="animate-pulse">
+    <% column_count.times do %>
+      <td class="py-2">
+        <div class="mb-2 h-4 w-1/2 rounded-sm bg-gray-200"></div>
```
joshsadam: The skeleton uses `bg-gray-200` with no `dark:` variant; the rest of the loading/table UI uses the slate scale with dark variants (see `shared/loading/samples_list_skeleton`). This will look wrong in dark mode and drifts from the design tokens. Use the slate equivalents with a dark variant.

test/system/projects/samples/attachments_test.rb
```diff
@@ -59,13 +65,18 @@ class AttachmentsTest < ApplicationSystemTestCase
...
         within('dialog[open]') do
-          assert_text 'test_file_fwd_1.fastq'
-          assert_text 'test_file_rev_1.fastq'
-          assert_text 'test_file_fwd_2.fastq'
-          assert_text 'test_file_rev_2.fastq'
-          assert_text 'test_file_fwd_3.fastq'
-          assert_text 'test_file_rev_3.fastq'
-          assert_text 'test_file_D.fastq'
+          assert_text I18n.t('projects.samples.attachments.deletions.modal.description.plural')
+                          .gsub! 'COUNT_PLACEHOLDER',
```
joshsadam: Nit: `String#gsub!` edits the string in place and returns `nil` when it finds no match. Since `I18n.t` can hand back a shared/frozen string, mutating it is risky, and this line would return `nil` (failing the assertion) if the wording ever changes so `COUNT_PLACEHOLDER` isn't present. Use the non-mutating `gsub` and fold the argument onto one call:

```ruby
assert_text I18n.t('projects.samples.attachments.deletions.modal.description.plural')
                .gsub('COUNT_PLACEHOLDER', '7')
```

app/views/projects/samples/attachments/concatenations/_modal.html.erb
```diff
@@ -1,52 +1,68 @@
-<%= render Viral::DialogComponent.new(open: open, size: :large) do |dialog| %>
+<%# locals: (open:, sample:, concatenation_params:, concatenation_form:) -%>
```
joshsadam: Nit: `concatenation_params` is never used in this template — the only `concatenation_params` is the unrelated controller method. Please remove it in the three places that mention it so the partial's declared locals match what it actually uses:

1. This file, line 1:
```erb
<%# locals: (open:, sample:, concatenation_form:) -%>
```
2. `concatenations_controller.rb` (`new` action): drop the `concatenation_params: nil,` line from the `locals:` hash.
3. `concatenations/create.turbo_stream.erb`: drop the `concatenation_params: nil,` line from the `locals:` hash.

app/views/projects/samples/attachments/deletions/_modal.html.erb
```diff
@@ -1,28 +1,59 @@
...
+                          "spinner-target": "submit",
+                        } %>
+        </div>
+      <% end %>
     </div>
-  <% end %>
+
+    <div
+      role="status"
+      id="spinner"
```
joshsadam: `id="spinner"` is a generic, non-namespaced DOM id and risks collision. Prefer a scoped id or rely solely on the spinner Stimulus target instead of a global id.

joshsadam (changes requested):

app/views/projects/samples/attachments/deletions/_modal.html.erb
ChrisHuynh333: Removed 26e706c7269905e329ff27907a68a72ae24b0a7a

app/controllers/concerns/list_actions.rb
ChrisHuynh333: `infinite_scroll_controller` will now parse out the ids into a flat array and pass the flattened array to `ListActions` where `ListActions` now doesn't need to do anything other than query for the attachments

c61ec9a8274a9ee4f865f25bea046d78a7e715a2

app/javascript/controllers/infinite_scroll_controller.js
ChrisHuynh333: Above comment

test/integration/concerns/list_actions_test.rb
ChrisHuynh333: Removed one of the tests (rebase issue), added extra asserts c61ec9a8274a9ee4f865f25bea046d78a7e715a2

app/views/shared/attachments/_list_attachment.html.erb
ChrisHuynh333: Leaving this for now, given how our table_components render each column, this would likely take a separate shared view/partial per column I think? Or an over complex way of calling each column. Seems not worth it to me. Let me know if you disagree

app/views/projects/samples/attachments/concatenations/_modal.html.erb
ChrisHuynh333: Removed c61ec9a8274a9ee4f865f25bea046d78a7e715a2

joshsadam (approved):
Looks good to me

ericenns (approved):
Looks good to me!

PR #2136, opened 2026-07-29
Prevent Sample Deletions & Transfers during Workflow Executions

ksierks:
This PR prevents the user from deleting or transferring samples that are currently used in an active workflow execution. 
Fixes STRY0020901.

https://github.com/user-attachments/assets/e9210d09-8f48-4d6c-8214-0e2576cb59b8

https://github.com/user-attachments/assets/6034a313-6ab3-4881-b2c0-1401af714b67

Enable the new feature:
1. Launch rails console and run the following:
```irb
Flipper.enable(:prevent_sample_deletions_and_transfers_with_active_workflows)
```

Test the new feature:
1. Run a workflow execution.
1. Try deleting a sample from the running workflow execution.
1. Verify an error message was displayed and the sample was not deleted.
1. Repeat all steps, but transfer the sample instead of deleting it.

## PR acceptance checklist


app/services/samples/transfer_service.rb
```diff
@@ -40,6 +40,10 @@ def execute(new_project_id, sample_ids, broadcast_target = nil)
 
       validate(sample_ids, 'transfer', new_project_id)
 
+      if Flipper.enabled?(:prevent_sample_deletions_and_transfers_with_active_workflows)
+        validate_no_active_workflow_executions(sample_ids)
```
joshsadam: Blocking: The transfer guard is only enforced in the V1 transfer path.

When `v2_sample_transfer` is enabled, the controller queues `TransferJobV2`, which uses `TransferServiceV2`. That V2 path does not call `validate_no_active_workflow_executions`.

Why it matters: with V2 enabled, a sample that is part of an active workflow can still be transferred, which breaks this PR's main safety rule.

Requested outcome: add the same active-workflow guard to V2 (or move the guard to one shared helper that both V1 and V2 call).

app/services/base_sample_service.rb
```diff
@@ -74,6 +74,28 @@ def validate(sample_ids, action_type, new_project_id = nil) # rubocop:disable Me
...
 
+  # Validate that samples do not have active workflow executions.
+  #
+  # Prevents transferring samples that are currently being used in a workflow execution.
+  # Active workflow states are: initial, prepared, submitted, running.
+  #
+  # @param sample_ids [Array<Integer>] IDs of samples to check
+  # @param action_type [String] the type of action being performed (e.g., 'transfer')
+  # @raise [BaseError] if any samples have active workflow executions
+  def validate_no_active_workflow_executions(sample_ids, action_type)
```
joshsadam: Suggestion: `BaseSampleService#validate_no_active_workflow_executions(sample_ids, action_type)` looks unused and currently unsafe if it ever runs.

Evidence:
- Current transfer and destroy flows call their own service-level methods, not this 2-argument base method.
- This method passes `sample_ids` to the translation, but the transfer/destroy translation strings expect `sample_puids`.

Why it matters: if this method is called in the future, Rails would raise `I18n::MissingInterpolationArgument`.

Requested outcome: either remove this unused method, or make it the shared implementation and update it to use `sample_puids`.

app/services/base_sample_destroy_service.rb
```diff
@@ -14,11 +16,36 @@ def initialize(namespace, user = nil, params = {})
...
   private
 
+  # Validate that samples do not have active workflow executions.
+  #
+  # Prevents transferring samples that are currently being used in a workflow execution.
+  # Active workflow states are: initial, prepared, submitted, running.
+  #
+  # @param sample_ids [Array<Integer>] IDs of samples to check
+  # @raise [DestroyError] if any samples have active workflow executions
+  def validate_no_active_workflow_executions(sample_ids)
```
joshsadam: Suggestion: The active-workflow validation logic is still duplicated in transfer and destroy services.

Both methods do the same pattern: find active workflow sample PUIDs, return early when none are found, and raise a service-specific error when matches exist.

Why it matters: duplicated safety checks can drift over time, and one path may be updated while the other is missed.

Requested outcome: extract a shared helper for the common validation/query behavior, while keeping each service's own error type/message.

app/services/base_service.rb
```diff
@@ -31,4 +33,12 @@ def update_progress_bar(current_count, total_count, broadcast_target)
   def strip_whitespaces(string)
     string.gsub(/\s+/, ' ').strip
   end
+
+  def active_workflow_execution_sample_puids(sample_ids)
```
joshsadam: Optional: `active_workflow_execution_sample_puids` is sample-specific behavior inside `BaseService`, which is inherited by all services.

Why it matters: this makes the global service base class carry domain-specific sample logic, which is harder for new contributors to reason about.

Requested outcome: move this helper to a sample-focused module or sample base service, and include it only where needed.

app/views/samples/deletions/_destroy_multiple_confirmation_dialog.html.erb
```diff
@@ -78,39 +84,39 @@
...
-            <%= form.submit t(".submit_button"),
-                          tone: :danger,
-                          emphasis: :solid,
-                          data: {
-                            "spinner-target": "submit",
-                            "turbo-submits-with": t(:".loading"),
-                          } %>
+            <div>
+              <%= form.submit t(".submit_button"),
+            class: "button text-sm px-5 py-2.5 text-white bg-red-700 border-red-800 hover:bg-red-800 dark:bg-red-600 dark:text-white dark:border-red-600 dark:hover:bg-red-700",
```
joshsadam: Suggestion: This submit button changed from the component API (`tone: :danger, emphasis: :solid`) to hard-coded red Tailwind classes.

Why it matters: component tone/emphasis keeps danger buttons consistent across the app and automatically follows future design-token updates.

Requested outcome: switch this button back to `tone: :danger, emphasis: :solid`, matching the single-delete dialog.

app/views/samples/deletions/_destroy_multiple_confirmation_dialog.html.erb
```diff
@@ -1,48 +1,56 @@
...
 
-  <div
-    data-controller="infinite-scroll"
-    data-infinite-scroll-selection-outlet="#samples-table"
-    data-infinite-scroll-paged-field-name-value="sample_ids[]"
-    data-infinite-scroll-singular-description-value="<%= t(".description.singular") %>"
-    data-infinite-scroll-plural-description-value="<%= t(".description.plural") %>"
-  >
-    <%= form_with(
+  <% if errors.present? %>
```
joshsadam: Optional: Both delete dialogs now contain the same `if errors.present? ... else ... end` wrapper and identical error-render block.

Why it matters: duplicated view logic is easier to miss during future updates and increases review noise.

Requested outcome: extract the shared error block into one small partial and render it from both dialogs.

joshsadam (changes requested):
Just a few commnets.  

Can you also update `app/views/shared/samples/_errors.html.erb`:

Could we switch this to use `Viral::AlertComponent` instead of custom red markup?

Why this helps:

**Consistency**: other error dialogs already use this component, so users get the same look and behavior everywhere.
**Accessibility**: the component already includes alert semantics for screen readers, so we do not have to manage that manually.
**Contrast safety**: the current reds are mostly okay, but some parts are close to the minimum WCAG AA threshold. The shared component uses safer defaults and reduces the risk of contrast regressions later.
This would simplify the template and make future maintenance easier.

app/services/base_sample_destroy_service.rb
JeffreyThiessen: It might make the most sense to have the BaseSampleDestroyService inherit from the BaseSampleService, and have this new validation function live in the BaseSampleService. I puzzled over this for a few minutes and I think it's the cleanest solution I can come up with.

test/system/projects/samples_test.rb
```diff
@@ -668,6 +668,71 @@ class SamplesTest < ApplicationSystemTestCase
       ### VERIFY END ###
     end
 
+    test 'prevent sample transfer during active workflow execution' do
```
JeffreyThiessen: Sample transfer related tests here (and groups/samples_test.rb) are being rewritten in https://github.com/phac-nml/irida-next/pull/2229 to be in the integrations directory instead.

When making these new tests, make sure to also handle the v1 and v2 versions of the transfer service so that all cases are covered.

JeffreyThiessen (changes requested):
Looks like a good start. I added some comments in addition to what Josh already touched on.

app/services/samples/transfer_service.rb
ksierks: This PR was up 3 weeks ago and the v2_sample_transfer was merged last week. I guess I missed it during the rebase. 70fc585da4a9d4c1e5adc71e6916e3b80e1e8b16

app/services/base_sample_service.rb
ksierks: This was left behind from a bad refactor/rebase, 70fc585da4a9d4c1e5adc71e6916e3b80e1e8b16

app/views/samples/deletions/_destroy_multiple_confirmation_dialog.html.erb
ksierks: 4120113b335a8fef2fba8b1e7b178d6c6b87db82

app/services/base_sample_destroy_service.rb
ksierks: 420ccb754b4bc3538095f4e34686b275c10fe7d7

app/services/base_service.rb
ksierks: 420ccb754b4bc3538095f4e34686b275c10fe7d7

app/views/samples/deletions/_destroy_multiple_confirmation_dialog.html.erb
ksierks: I tried refactoring the error block, but it resulted with a partial only rendering another partial. I'd like to keep this as is.

joshsadam: Thanks for taking a pass at this.

I still think we should extract the shared error block so we only maintain it in one place. Rendering a partial from another partial is okay in Rails and keeps repeated logic easier to update.

Could you create one small shared partial for the error section (for example, the if errors.present? block that renders shared/samples/errors) and render that from both delete dialogs? Please keep the current UI and behavior the same; this is only about removing duplication.

If there is a specific technical blocker, please add a short note with the exact error so we can pick a safer alternative.

ksierks: The small requested partial is here, 8388a695c951777cbce3e392db8c17bf25707da7.

test/system/projects/samples_test.rb
ksierks: Your PR is approved, just waiting for it to get merged before addressing this comment, thanks.

joshsadam (dismissed):

test/system/projects/samples_test.rb
ksierks: 5f68a17196df32d44aa3362931a58bd04afdbb3e for service tests &
a687f13dfbadfd7cc68b890eee98599972d54b62 for integration tests

JeffreyThiessen (approved):
Great work! The refactor looks clean and the tests are solid. 👍

PR #2157, opened 2026-08-07
Bug: Advanced Search - Use metadata operators for metadata enum fields

ChrisHuynh333:
This PR fixes a bug where once the metadata operator feature flags are enabled and metadata fields must use metadata operators, metadata enum fields (namely workflow name and ver) would error through the UI as they still utilized standard operators (=, !=, in, not_in)
Now, enum operators populate based on whether the selected field is standard or metadata.

Before:

[Screencast from 2026-08-07 12-12-19.webm](https://github.com/user-attachments/assets/363b5286-0772-4962-88db-a74bf440a059)

Now:

[Screencast from 2026-08-07 12-13-41.webm](https://github.com/user-attachments/assets/77b22993-e84f-4e0a-a4b9-cc2c365c63e8)

1. Enable feature flag `Flipper.enable(:advanced_search_metadata_operators)`
2. Navigate to a Workflow Executions page, and ensure you have some workflows to filter through.
3. Select a standard field that is not State, and verify the normal standard operators populate
4. First, select both `State` and `Workflow Name` fields and verify the enum operators populate (State should have `=, !=, in, not_in`, and workflow name should have `text_equals, text_not_equals, text_in, text_not_in`.)
5. Create a search for `State`  and verify the search filters as expected. Re-open the `Advanced Search` dialog, and verify the search conditions are populated as expected.
6. Repeat the above for `Workflow Name`

## PR acceptance checklist

### DIFF
--- a/app/components/advanced_search/v1/component.rb
+++ b/app/components/advanced_search/v1/component.rb
@@ -31,13 +31,17 @@ module AdvancedSearch
         enum_operators = {}
         if Flipper.enabled?(:advanced_search_metadata_operators)
           # flatten metadata operators to exclude optgroup labeling
-          metadata_operators = operation_options['metadata']
+          metadata_operators = @operations['metadata']
                                .values
                                .reduce({}, :merge)
-          enum_operators['metadata'] = metadata_operators.select { |_, value| AdvancedSearch::ENUM_OPERATOR_VALUES['metadata'].include?(value) }
+          enum_operators['metadata'] = metadata_operators.select do |_, value|
+            AdvancedSearch::ENUM_OPERATOR_VALUES['metadata'].include?(value)
+          end
         end
 
-        enum_operators['standard'] = operation_options['standard'].select { |_, value| AdvancedSearch::ENUM_OPERATOR_VALUES['standard'].include?(value) }
+        enum_operators['standard'] = @operations['standard'].select do |_, value|
+          AdvancedSearch::ENUM_OPERATOR_VALUES['standard'].include?(value)
+        end
         enum_operators
       end
 
--- a/app/components/advanced_search/v1/component.rb
+++ b/app/components/advanced_search/v1/component.rb
@@ -28,11 +28,18 @@ module AdvancedSearch
       end
 
       def enum_operation_options
-        operation_options['standard'].select { |_, value| enum_operation_values.include?(value) }
-      end
+        enum_operators = {}
+        if Flipper.enabled?(:advanced_search_metadata_operators) &&
+           Flipper.enabled?(:advanced_search_disable_standard_operators_for_metadata_in_graphql)
+          # flatten metadata operators to exclude optgroup labeling
+          metadata_operators = operation_options['metadata']
+                               .values
+                               .reduce({}, :merge)
+          enum_operators['metadata'] = metadata_operators.select { |_, value| AdvancedSearch::ENUM_OPERATOR_VALUES['metadata'].include?(value) }
+        end
 
-      def enum_operation_values
-        AdvancedSearch::ENUM_OPERATOR_VALUES
+        enum_operators['standard'] = operation_options['standard'].select { |_, value| AdvancedSearch::ENUM_OPERATOR_VALUES['standard'].include?(value) }
+        enum_operators
       end
 
       def operation_options # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
--- a/app/components/advanced_search/v1/component.rb
+++ b/app/components/advanced_search/v1/component.rb
@@ -29,8 +29,7 @@ module AdvancedSearch
 
       def enum_operation_options
         enum_operators = {}
-        if Flipper.enabled?(:advanced_search_metadata_operators) &&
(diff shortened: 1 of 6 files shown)


app/components/advanced_search/v1/condition_component.rb
```diff
@@ -62,11 +62,21 @@ def enum_fields
       end
 
       def enum_operator_options
-        @operations['standard'].select { |_, value| enum_operator_values.include?(value) }
+        if Flipper.enabled?(:advanced_search_metadata_operators) && selected_field.starts_with?('metadata')
```
joshsadam: In other places the code has `.starts_with?('metadata.')` let's keep it that way to keep consistency.  Maybe we should make a constant for this and use it in all places?

app/components/advanced_search/v1/condition_component.rb
```diff
@@ -62,11 +62,21 @@ def enum_fields
...
+          # flatten metadata operators to exclude optgroup labeling
+          metadata_operators = @operations['metadata']
```
joshsadam: nit: You already have a method `metadata_operators` above, this shadows the private method.  It works, but the  name collision is confusing.  Maybe rename to `flattened_metadata_operators` or something

app/components/advanced_search/v1/component.rb
```diff
@@ -28,11 +28,17 @@ def normalized_fields(fields:, sample_fields:, metadata_fields:)
       end
 
       def enum_operation_options
-        operation_options['standard'].select { |_, value| enum_operation_values.include?(value) }
-      end
+        enum_operators = {}
+        if Flipper.enabled?(:advanced_search_metadata_operators)
+          # flatten metadata operators to exclude optgroup labeling
+          metadata_operators = operation_options['metadata']
```
joshsadam: nit:  `enum_operation_options` calls `operation_options` twice instead of reusing the already-memoized `@operations`. Since `operation_options` rebuilds a full I18n hash each call, using `@operations` would be both clearer and cheaper.

app/components/advanced_search/v1/condition_component.html.erb
```diff
@@ -104,7 +100,6 @@
       dark:hover:text-white
     "
     aria-label="<%= t("components.advanced_search_component.v1.remove_condition_aria_label") %>"
-    title="<%= t("components.advanced_search_component.v1.remove_condition_aria_label") %>"
```
joshsadam, line 107: Was removing title intended?

joshsadam (changes requested):
This is looking good, just a couple suggestions

app/components/advanced_search/v1/condition_component.html.erb
ChrisHuynh333: Yes, linter was mad that `aria-label` and `title` coexisted, only one is needed

app/components/advanced_search/v1/condition_component.rb
ChrisHuynh333: Agreed, but leaving for now. This will be part of the big overhaul centralizing data and all the shared logic throughout the advanced search component

joshsadam: Could you update this to be `metadata.` here so it matches the rest, then do the rest of the update later?

app/components/advanced_search/v1/condition_component.html.erb
joshsadam: Smart linter ;)

app/components/advanced_search/v1/condition_component.rb
ChrisHuynh333: Updated 72f9c904a3119f92c870b596b8e57573a93e6b87

app/components/advanced_search/v1/condition_component.rb
ChrisHuynh333: Renamed in `flattened_metadata_operators` 72f9c904a3119f92c870b596b8e57573a93e6b87

app/components/advanced_search/v1/component.rb
ChrisHuynh333: Thanks for catching that, 72f9c904a3119f92c870b596b8e57573a93e6b87

app/components/advanced_search/v1/condition_component.html.erb
ericenns: Not sure I really agree with the linter on that. `title` can be seen by visual users on hover while `aria-label` is only accessible to screen readers. In this case wouldn't `title` be preferred as both sighted and non-sighted users benefit from the extra label?

Which in that case I think we should add an override for the linter so that we can have both, as the screenreader sometimes ignores `title`.

Or we switch to using tooltips for buttons which we want to provide additional content to both screenreader and non-screenreader users. But maybe that could be addressed later on.

ChrisHuynh333: Added override 40da136d498695c9b38573938f4e4c81b48e74ef

deepsidhu85 (approved):
Looks good to me and works as described!

joshsadam (approved):

PR #2163, opened 2026-08-11
Bug: Fix sample attachment selection count after deletion through row action link

ChrisHuynh333:
This PR fixes two bugs that were found when deleting attachments by the delete row action link.
1. Successful PE file deletion led to a `:multi_status` status, rather than to `:success`
2. Successful PE file deletion, when selected, did not update the selected count correctly.

Before:

[Screencast from 2026-08-11 15-04-57.webm](https://github.com/user-attachments/assets/8d7ae43d-820e-4c36-a71b-64bb62d5ead4)

Now:

[Screencast from 2026-08-11 15-05-34.webm](https://github.com/user-attachments/assets/0db463a8-71a5-4361-8339-acd2b69d36e0)

1. Navigate to sample attachments page, and upload some files, including PE files.
2. Select any number of attachments, and delete a selected PE file, and verify the selected count updates as expected
3. Repeat the above with a non-PE file, and using the `Delete Files` button, and verify the selected count updates as expected.
4. Test that selection counts still update correctly elsewhere in the application (like workflows and sample tables).

## PR acceptance checklist

### DIFF
--- a/app/javascript/controllers/selection_controller.js
+++ b/app/javascript/controllers/selection_controller.js
@@ -163,17 +163,15 @@ export default class extends Controller {
     } else {
       newStorageValue = newStorageValue.filter((value) => {
         try {
+          // parse potential stringified array for PE attachments being deleted
           const parsed = JSON.parse(value);
 
-          // If it's a nested array, remove it if ANY value matches
+          // if PE attachment, it will be a nested array value where the entire array must be removed
           if (Array.isArray(parsed)) {
             return !parsed.some((item) => values.includes(item));
           }
-
-          // Otherwise, treat it as a normal string
-          return !values.includes(value);
         } catch {
-          // Not JSON, so treat it as a normal string
+          // else remove the single attachment string value
           return !values.includes(value);
         }
       });
--- a/app/javascript/controllers/selection_controller.js
+++ b/app/javascript/controllers/selection_controller.js
@@ -162,18 +162,20 @@ export default class extends Controller {
       newStorageValue = [...new Set([...newStorageValue, ...values])];
     } else {
       newStorageValue = newStorageValue.filter((value) => {
+        // exact match (single checkbox toggle / select-page store the raw value)
+        if (values.includes(value)) return false;
+        let parsed;
         try {
-          // parse potential stringified array for PE attachments being deleted
-          const parsed = JSON.parse(value);
-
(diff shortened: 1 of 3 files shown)


app/javascript/controllers/selection_controller.js
```diff
@@ -161,9 +161,20 @@ export default class extends Controller {
       // Use a Set to deduplicate values
       newStorageValue = [...new Set([...newStorageValue, ...values])];
     } else {
-      newStorageValue = newStorageValue.filter(
-        (value) => !values.includes(value),
-      );
+      newStorageValue = newStorageValue.filter((value) => {
```
joshsadam, line 164: https://github.com/user-attachments/assets/e63861c9-43f0-45cf-bfc1-694d826d5978

Unchecking a paired-end checkbox (single toggle, or via the select-page header checkbox) no longer deselects it. In `toggle`/`togglePage`, `values` contains the raw checkbox value — the array-string `"[fwd, rev]"`. The new filter parses the stored array-string and matches individual ids (`parsed.some(item => values.includes(item))`); since values holds the whole array-string and not `fwd/rev,` nothing matches and the entry is kept. `#updateUI` then re-sets `row.checked = ids.indexOf(row.value) > -1`, so the box snaps back to checked.

Make the removal filter handle both an exact-value match and a contained-id match, and always return a boolean. Suggested:

```javascript
newStorageValue = newStorageValue.filter((value) => {
  // exact match (single checkbox toggle / select-page store the raw value)
  if (values.includes(value)) return false;
  let parsed;
  try {
    parsed = JSON.parse(value);
  } catch {
    return true; // plain id that wasn't targeted
  }
  // contained-id match (row-action delete passes a single id that may be
  // part of a stored paired-end "[fwd, rev]" value)
  if (Array.isArray(parsed)) {
    return !parsed.some((item) => values.includes(item));
  }
  return true;
});
```

joshsadam (changes requested):
Found a bug, cannot deselect a paired-end file

app/javascript/controllers/selection_controller.js
ChrisHuynh333: Thanks for catching this, updated in 227c5a237e43664ff49897c5e32aaec26ad718c6 with associated system test

deepsidhu85 (approved):
Looks good to me and working as expected!

joshsadam (approved):

PR #2183, opened 2026-08-18
Workflow Submissions and File Selector system tests refactor

ChrisHuynh333:
This PR refactors the submissions and file selector tests. A large number of tests were refactored to both integration and javascript testing.

JS testing has two lines uncovered, which require specific configuration to hit.
`line 809` - pertains to `x-irida-next-selected` for specific metadata header overrides based on pipeline configuration
`line 1045` - pertains to chunked sample fetches, requiring both front and backend communcation.
Both of these are covered in other tests.

There are still a handful of large system tests for submissions, as workflow submissions requires a lot of server and client side 'talk' throughout a submission process. Due to the large 'overhead' of setting up these tests (navigate to samples table, select samples, clicking launch workflow button, selection dialog etc.), tests have been combined to test what used to be multiple tests in a smaller number of larger tests.

Example for metadata selection testing, we had a test to:
Verify metadata populated the field
Metadata selection retained during during filtering
Metadata selection retained/submitted after workflow submission

These are now combined into a single test.

File Selector:
image

Submission:
image

Samplesheet:
image

1. Ensure that services are running (e.g. `devenv up`)
2. Run all the submission integration tests
```bash
bin/rails test test/integration/groups/group_links_test.rb test/integration/workflow_executions/submissions.rb 
```
3. In `coverage` folder launch a python server to serve coverage report (e.g. `python -mhttp.server`)
4. Open localhost:8000 in your browser
5. Click on `Controllers` tab and type `submissions` in search box and confirm percentages match the image above
6. Repeat above for file_selector

## PR acceptance checklist


test/javascript/controllers/nextflow/v2/samplesheet_controller.test.js
```diff
@@ -0,0 +1,1448 @@
...
+      code: "Enter",
+      bubbles: true,
+    }),
+  );
+
+  await new Promise((resolve) => setTimeout(resolve, 60));
+};
+
+/* eslint-disable no-useless-escape */
+function renderFixture() {
```
joshsadam: Suggestion: this spec is hard to maintain because `renderFixture` contains one very large inline HTML template.

Why it matters: small UI changes can create large, unrelated diffs, and failures become harder to debug.

Could we split this into focused specs (for example pagination/filtering, submit validation, and update-samples authorization) and move fixture setup into small builders per scenario? That keeps each test easier for junior contributors to understand and update.

test/javascript/controllers/nextflow/v2/samplesheet_controller.test.js
```diff
@@ -0,0 +1,1448 @@
...
+    expect(getPageNum()).toBeNull();
+    expect(getNextBtn()).toBeNull();
+
+    expect(clearButton).not.toHaveClass("hidden");
+
+    assertTableData(allSamples, [2, 21]);
+
+    clearButton.click();
+
+    await new Promise((resolve) => setTimeout(resolve, 60));
```
joshsadam: Blocking: this file uses fixed waits such as `await new Promise((resolve) => setTimeout(resolve, 60));`.

Why it matters: fixed sleeps are timing-dependent and can become flaky on slower CI runners.

Please switch to deterministic synchronization (for example fake timers with Vitest, or waiting for a specific DOM/state change) so assertions run only after the expected state is ready.

test/controllers/workflow_executions/submission_controller_test.rb
```diff
@@ -6,43 +6,15 @@ module WorkflowExecutions
...
-      assert_response :ok
-    end
-
-    test '@fields in create' do
       post workflow_executions_submissions_path(format: :turbo_stream,
                                                 pipeline_id: 'phac-nml/iridanextexample',
                                                 workflow_version: '1.0.2', namespace_id: @group.id)
       assert_response :ok
       assert_equal ['metadata field with spaces', 'metadatafield1', 'metadatafield2', 'unique.metadata.field'],
                    @controller.instance_eval('@fields', __FILE__, __LINE__)
```
joshsadam: Suggestion: this test reaches into controller internals with `@controller.instance_eval('@fields', __FILE__, __LINE__)`.

Why it matters: integration tests are strongest when they verify observable behavior. Internal ivar checks can break during harmless refactors and may miss user-facing regressions.

Could we assert response/rendered behavior that depends on these fields, or extract field-building into a small object and unit-test that object directly?

test/integration/workflow_executions/file_selector_test.rb
```diff
@@ -0,0 +1,366 @@
...
+        }
+      )
+
+      assert_response :success
+      attachments = [selected_attachment, attachments(:attachmentPEFWD2), attachments(:attachmentPEFWD1),
+                     attachments(:attachmentF), attachments(:attachmentE), attachments(:attachmentD)]
+      assert_select 'table' do
+        assert_select 'tbody' do
+          attachments.each do |attachment|
+            assert_select 'tr' do
```
joshsadam: Blocking: these assertions are under-scoped because `assert_select 'tr'` is broad inside the per-attachment loop.

Why it matters: filename, metadata, and checked-state assertions can pass even if they come from different rows, so row-mapping bugs may be missed.

Can we scope each check to a single row (or parse rows once and compare row tuples) so all expected values are validated on the same row?

test/integration/workflow_executions/submissions_test.rb
```diff
@@ -7,15 +7,53 @@ class SubmissionsTest < ActionDispatch::IntegrationTest
     setup do
       sign_in users(:john_doe)
       @group = groups(:group_one)
+
+      @samplesheet_properties_payload = {
```
joshsadam, line 11: Optional: there is repeated test data and ordering assertions in this file (for example `@samplesheet_properties_payload` and repeated workflow-order checks).

Why it matters: when values change, multiple test blocks must be updated, which increases maintenance cost and risk of drift.

A small payload helper plus a shared workflow-order assertion helper would keep this easier for junior contributors to maintain.

joshsadam (changes requested):

test/integration/workflow_executions/submissions_test.rb
ChrisHuynh333: I'm not sure what this is saying. By having a shared payload, if values need to be changed, they only have to be changed in the single payload.

There is nothing in regards to asserting any ordering with this payload. It's merely to pass through validation and processing in order to check an associated boolean result.

test/integration/workflow_executions/file_selector_test.rb
ChrisHuynh333: Added IDs to rows and specifying row assertions based on that 34a0ab6939f7e90a69eaf9c01dc6eb28af79ec6c

test/controllers/workflow_executions/submission_controller_test.rb
ChrisHuynh333: Verifying metadata options in samplesheet header
ce58ed358ea84d22ea8c6d105f38d56e1ae13567 814dfed76d52a83e524e6a644f319db758b9c242

test/javascript/controllers/nextflow/v2/samplesheet_controller.test.js
ChrisHuynh333: Using `vi.advanceTimersByTime` now c21853cc9f405bcf8c77801bc8b5a8e3fd8e0d8d

test/javascript/controllers/nextflow/v2/samplesheet_controller.test.js
ChrisHuynh333: I've tried breaking up the samplesheet into it's main components and calling separate renders to them. However the whole complexity of the samplesheet is it essentially requires all the parts to function, there's only a few tests that can call partial renders without breaking the component.
Regardless, breaking up the main portions of the component likely make it a little easier to read and identify these core parts within the DOM
7fd4df520cc8cb058ccf370ccc6ec452d9e4f516

test/integration/workflow_executions/submissions_test.rb
joshsadam: Thanks for the reply and for clarifying the payload part. You are right that the shared payload already avoids duplication there.

The remaining ask is specifically about the repeated workflow list ordering assertions, which still appear in multiple tests. Could you extract that repeated li ordering check into one small helper so we only maintain the expected ordering in one place?

Something like this in `submissions_test.rb` would work:

```ruby
def assert_workflow_order(expected_items)
  assert_select 'li', expected_items.length do |items|
    expected_items.each_with_index do |expected_text, index|
      assert_equal expected_text, items[index].text.squish
    end
  end
end
```

Then each test can call:

```ruby
assert_workflow_order([
  'phac-nml/gasclustering 0.4.2 Genomic Address Service Clustering Workflow',
  'phac-nml/iridanextexample 1.0.2 IRIDA Next Example Pipeline',
  'phac-nml/iridanextexample 1.0.1 IRIDA Next Example Pipeline',
  'phac-nml/snvphylnfc 2.4.0 SNVPhyl nf-core pipeline'
])
```

ChrisHuynh333: Helpers added bd52b09af45716863b98adc345169502ddcbbc71

ericenns (approved):
Looks good to me!

PR #2192, opened 2026-08-21
Refactor: Server side pipeline selection button disabling

ChrisHuynh333:
This PR is an aside to the submissions testing refactor,  and refactors the pipeline selection disabling, disabled message adding, and button ordering (available vs unavailable) to be handled server side rather than client side.

Stimulus is still used for submission upon button press

image

image

1. Verify pipeline selection works as expected
2. Verify the pipelines still disable if the sample count is less/more than the pipeline allows

## PR acceptance checklist


app/components/workflow_executions/workflow_selection_button_component.rb
```diff
@@ -0,0 +1,54 @@
...
+      if @min_samples_configured
+        @workflow.minimum_samples.to_i
+      else
+        0
+      end
+    end
+
+    def maximum_samples
+      if @max_samples_configured
+        @workflow.maximum_samples.to_i
```
deepsidhu85: This if/else block can probably just be replaced with `@workflow.maximum_samples.to_i` as if max_samples is not set it will return -1 by default

deepsidhu85 (changes requested):
This looks good just have one comment below

deepsidhu85 (changes requested):
One more comment, we should probably add some tests/previews for the new components

app/components/workflow_executions/workflow_selection_button_component.rb
ChrisHuynh333: Thanks, fixed in fa38223566650f89c38cc151a2d8c80e1e286a9f. Ended up just removing `def maximum_samples` and calling directly in the code

ChrisHuynh333:
> One more comment, we should probably add some tests/previews for the new components

I've added the previews fa38223566650f89c38cc151a2d8c80e1e286a9f f7a51b40c5c439425199ac81082458fce83fbeb6

My plan was to add integration testing into the main submissions testing refactor after this PR is merged, if that's okay with you. That should also cover anything the component testing would touch, did we still want separate component testing if that's the case?

ChrisHuynh333:
@joshsadam @deepsidhu85 Refactor of  workflow selection is done and ready for re-review (along with some integration tests specific for this). Backend now handles organizing and disabling pipelines.

app/components/workflow_executions/workflow_selection_component.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      elsif max_samples_configured && @sample_count > workflow.maximum_samples.to_i
+        I18n.t('shared.workflow_executions.sample_limits.max_samples_exceeded',
+               max_samples: workflow.maximum_samples)
+      else
+        ''
+      end
+    end
+
+    def minimum_samples(min_samples_configured, workflow)
```
joshsadam: Suggestion: the `else 0` branch is unreachable and this `.to_i` is redundant.

`minimum_samples` is only evaluated after `min_samples_configured &&` short-circuits (line 31), so the parameter is always `true` and `else 0` can never run. `Irida::Pipeline#minimum_samples` (pipeline.rb:127) already returns `.to_i`, so the helper adds nothing. The same redundant `.to_i` appears on `workflow.maximum_samples.to_i` at line 35 (`maximum_samples` also coerces). Consider deleting the helper and inlining `workflow.minimum_samples` / `workflow.maximum_samples`.

app/components/workflow_executions/workflow_selection_component.rb
```diff
@@ -0,0 +1,51 @@
...
+
+        if message.blank?
+          organized[:available] << workflow
+        else
+          organized[:unavailable] << { workflow:, message: }
+        end
+      end
+    end
+
+    def sample_limit_message(workflow)
```
joshsadam: Suggestion: move this limit check onto the value object that owns the limits.

`sample_limit_message` encodes "is this sample count valid for this pipeline?" by reaching into four `Irida::Pipeline` predicates. That rule belongs on `Irida::Pipeline` (already unit-tested in `test/lib/irida/pipeline_test.rb`), e.g. `#sample_limit_message(sample_count)` returning `nil` when valid. The component then reduces to a pure sort:

```ruby
message = workflow.sample_limit_message(@sample_count)
```

This makes the rule unit-testable without rendering and keeps the component presentational.

app/components/workflow_executions/workflow_selection_component.html.erb
```diff
@@ -0,0 +1,38 @@
+<%= render Viral::BaseComponent.new(tag: 'div', data: { controller: "workflow-selection", "workflow-selection-selection-outlet": "#samples-table", "workflow-selection-field-name-value": "sample_count", "workflow-selection-sample-count-value": @sample_count }) do %>
```
joshsadam: Suggestion: render `sample_count` as a hidden field instead of round-tripping it through Stimulus.

The component already knows `@sample_count`, but it's exposed as `workflow-selection-sample-count-value` only so `#toJson` can re-inject it via `normalizeParams(params, this.fieldNameValue, this.sampleCountValue, 0)` (`workflow_selection_controller.js:87`). Rendering it directly:

```erb
<%= form.hidden_field :sample_count, value: @sample_count %>
```

lets `#toJson` collapse to `formDataToJsonParams(formData)`, and both the `fieldName` and `sampleCount` Stimulus values can be removed — one server-rendered input replacing a value plus a JS normalization step.

app/javascript/controllers/launch_workflow_controller.js
```diff
@@ -0,0 +1,17 @@
+import { Controller } from "@hotwired/stimulus";
+import { createHiddenInput } from "utilities/form";
+
+export default class extends Controller {
+  static outlets = ["selection"];
+
+  appendSelectionCount() {
+    const form = this.element.closest("form");
+
+    form.appendChild(
```
joshsadam, line 12: Nit: `appendSelectionCount` appends a new `sample_count` input on every click.

`form.appendChild(createHiddenInput("sample_count", …))` runs per click, but the form lives in the persistent page header, so repeated clicks accumulate duplicate `sample_count` inputs. Rails' last-wins keeps submissions correct, but the leftover inputs are avoidable — update or clear the existing input before appending.

app/components/workflow_executions/workflow_selection_component.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      if min_samples_configured && @sample_count < minimum_samples(min_samples_configured, workflow)
+        I18n.t('shared.workflow_executions.sample_limits.min_samples_required',
+               min_samples: workflow.minimum_samples)
+
+      elsif max_samples_configured && @sample_count > workflow.maximum_samples.to_i
+        I18n.t('shared.workflow_executions.sample_limits.max_samples_exceeded',
+               max_samples: workflow.maximum_samples)
+      else
+        ''
```
joshsadam: Nit: return `nil` rather than `''` for "no message".

Every caller checks `blank?`/`present?`, so `nil` reads more naturally as "no message" and matches idiomatic presence checks. (Folds into the `Irida::Pipeline` move suggested above.)

test/components/previews/workflow_selection_component_preview/default.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= render WorkflowExecutions::WorkflowSelectionComponent.new(workflows:
```
joshsadam: Optional: de-duplicate the preview setup.

The three preview templates each inline a near-identical 20-line `Irida::Pipeline.new(...)` triple while the `preview.rb` methods are empty stubs. Building the workflow list once in `preview.rb` (a shared private helper varying only settings/`sample_count`) and passing it via locals removes the duplication and gives the stub methods a purpose.

joshsadam (changes requested):

app/components/workflow_executions/workflow_selection_component.rb
ChrisHuynh333: Removed 78a86eb53ffa9dc2d616c1a4ac317ccdd0cfc022

app/components/workflow_executions/workflow_selection_component.rb
ChrisHuynh333: Moved `sample_limit_message` to `lib/irida/pipeline` and added associated tests 78a86eb53ffa9dc2d616c1a4ac317ccdd0cfc022

app/components/workflow_executions/workflow_selection_component.html.erb
ChrisHuynh333: Added as hidden form field 78a86eb53ffa9dc2d616c1a4ac317ccdd0cfc022

app/javascript/controllers/launch_workflow_controller.js
ChrisHuynh333: Clear input if it exists 78a86eb53ffa9dc2d616c1a4ac317ccdd0cfc022

app/components/workflow_executions/workflow_selection_component.rb
ChrisHuynh333: Return nil 78a86eb53ffa9dc2d616c1a4ac317ccdd0cfc022

test/components/previews/workflow_selection_component_preview/default.html.erb
ChrisHuynh333: Refactored to call helper function to create pipelines 78a86eb53ffa9dc2d616c1a4ac317ccdd0cfc022

app/javascript/controllers/launch_workflow_controller.js
```diff
@@ -0,0 +1,20 @@
...
+
+    form.querySelector('input[name="sample_count"]')?.remove();
+
+    form.appendChild(
+      createHiddenInput(
+        "sample_count",
+        this.selectionOutlet.getOrCreateStoredItems().length,
+      ),
+    );
+    console.log(form);
```
joshsadam: Leftover debug statement — please remove this `console.log(form)` before merge.

ChrisHuynh333: 89562d454c73c5d4082d740d7536eb05c5e9964a

joshsadam (approved):

deepsidhu85 (approved):
Looks good to me!

PR #2217, opened 2026-08-31
Datepicker: testing refactor

ChrisHuynh333:
This PR migrates `datepicker` testing to be purely done through `vitest/JS`, and removes the majority of `component/system` tests.
Some cleanup was also done within the controllers to handle edge cases (likely to not be needed in real-world use) or now unused code (datepicker no longer handles form submission)

image

image

1. Run `pnpm install --frozen-lockfile`.
2. Run `pnpm run test:js:coverage` — verify `combobox_datepicker/v1/input_controller and calendar_controller` and verify both controllers have 100% coverage.

## PR acceptance checklist


test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
```diff
@@ -0,0 +1,1773 @@
+import { Application } from "@hotwired/stimulus";
```
joshsadam: Suggestion: Could we split this spec into a few behavior-focused files and move shared setup/assertions into helpers? At 1773 lines, it is becoming harder to review safely and easier for unrelated changes to introduce regressions.

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
```diff
@@ -0,0 +1,1773 @@
...
+import { beforeEach, afterEach, describe, expect, it, vi } from "vitest";
+import FloatingDropdown from "../../../../../app/javascript/utilities/floating_dropdown.js";
+import InputController from "../../../../../app/javascript/controllers/combobox_datepicker/v1/input_controller.js";
+import CalendarController from "../../../../../app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js";
+
+/* eslint-disable no-useless-escape */
+function renderBaseFixture(locale = "en") {
```
joshsadam: Issue: Could we reduce `renderBaseFixture` to the minimum DOM contract the controllers need? Right now it mirrors full component markup, which can drift from ERB/i18n wiring and allow JS tests to pass while template wiring regresses. Keeping a small system smoke suite for real template wiring would help cover that integration path.

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
```diff
@@ -0,0 +1,1773 @@
...
+
+      await vi.runOnlyPendingTimersAsync();
+      expect(document.activeElement).toBe(getDateNode("2027-05-07"));
+
+      keypressOnDateNode("2027-05-07", "PageUp", true);
+      await vi.runOnlyPendingTimersAsync();
+      expect(document.activeElement).toBe(getDateNode("2026-05-07"));
+    });
+
+    it("Shift Page Up and Down navigation", async () => {
```
joshsadam: Issue: It looks like we have the same `Shift Page Up and Down navigation` scenario twice (around lines 1034 and 1053). Could we dedupe one and use `it.each(...)` for nearby navigation variants so we keep confidence while reducing maintenance overhead?

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
```diff
@@ -0,0 +1,1773 @@
...
+
+      inputControllerInstance(application).disconnect();
+    });
+
+    it("logs an error when the calendar cannot be found", async () => {
+      const consoleErrorSpy = vi
+        .spyOn(console, "error")
+        .mockImplementation(() => {});
+
+      vi.spyOn(document, "getElementById").mockImplementation((id) => {
```
joshsadam: Issue: Could we restore this `document.getElementById` spy in the same test (or in a consistent `afterEach`)? I can see it being created here, but not a nearby restore, which risks mock leakage and order-dependent failures across the suite.

joshsadam (changes requested):

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
ChrisHuynh333: Yes the DOM is large, but essentially everything is required for testing and functionality. I took out the calendar dots icon in the input and some button text.
Breaking up the component and trying to call multiple renders caused issues as well, so I'm going to leave that as well
5f195e63674b478fa08034ad3b1f24ead3d5cab2

Let me know if there's something specific you had in mind

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
ChrisHuynh333: Removed duplicated test 5f195e63674b478fa08034ad3b1f24ead3d5cab2

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
ChrisHuynh333: Restored mocks 5f195e63674b478fa08034ad3b1f24ead3d5cab2

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
joshsadam: Thanks for explaining the rendering issue. Could we move the fixture, setup, and shared assertions into helpers, then split the tests into input, navigation, and date-boundary files? Each test could still render the complete fixture once.

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
joshsadam: Thanks for trimming the fixture. Could we also remove the remaining decorative SVG markup and retain a small system smoke test that renders the real component, opens the calendar, selects a date, and checks the input value? The handwritten JS fixture cannot catch changes that break the ERB wiring.

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
joshsadam: Thanks for removing the duplicate. Could the simple navigation variants also use `it.each(...)`? The leap-year and month-boundary sequences can stay as separate tests.

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
ChrisHuynh333: re-added basic component system test ea158cf0987d5d7d4bbf4ad983c41b0fd723cd2c

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
ChrisHuynh333: `it.each` doesn't work well, as it resets the fixture when sometimes I need to start a sequence on a previous/next month, built upon from the previous step(s).

Maybe I had set it up incorrectly but regardless, I've instead just extracted out the steps, similar to `it.each` but within a single test, and the iterate that way. It will have similar readability.

fe4169d46e7b70b5cbac99560bed46fd04190e4a

test/javascript/controllers/combobox_datepicker/v1/combobox_datepicker.test.js
ChrisHuynh333: Broke out tests and helpers into multiple files fe4169d46e7b70b5cbac99560bed46fd04190e4a. `basic_component.test.js` is a bit of a catch all, but I think breaking that one out further is likely excessive

joshsadam (dismissed):
Looks good to me

test/javascript/controllers/combobox_datepicker/v1/keyboard_navigation.test.js
```diff
@@ -0,0 +1,340 @@
...
+    it("Arrow key navigation", async () => {
+      vi.setSystemTime(new Date("2026-05-07T09:00:00-05:00"));
+      renderBaseFixture();
+      application = await startController();
+
+      openCalendarByInputArrow();
+
+      await vi.runOnlyPendingTimersAsync();
+
+      await expectKeyboardNavigation([
```
ericenns, line 48: nitpick(blocking): Can you add more keyboard navigation cases here so that lines 649 and 697 are covered.

e.g. left, right, up, down which causes month changes.

Maybe that could be in a new test case, title month bounds keyboard navigation.

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js
ericenns: nitpick(blocking): I noticed this line is not covered in the tests and appears like it will always evaluate as true. So in that case I would remove the statment.

ericenns (changes requested):
I looked into the uncovered lines and found a few which could be covered by test cases:

app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js

* Line 583: To test this, use js to focus a date with aria-disabled and then trigger Space or Enter key which will case the return to fire as it is disabled
* Line 556: To test this get an instance of the controller (done in other tests) and then call this method with a synthetic event that uses a non unhandled key which should just return nothing. (Also appears line the if statement on line 557 can be removed as we have no event handlers for Tab key so we always prevent default).
* Line 497: To test this manually trigger a change event in javascript on the element without changing the value
* Line 417: Unsure on how to test this any ideas @joshsadam 
* Line 222: Maybe going from September 2026 to October 2026 would trigger this? I think having a test case where firstCurrentIndex evaluates to -1 would be usefull just so someone looking at the code later can understand why this was needed

joshsadam (commented):
## Coverage follow-up: pushing the datepicker JS toward 100%

I ran V8 coverage on the JS files in this PR. `utils.js` is already fully covered, and `input_controller.js` / `calendar_controller.js` are at 100% statements/functions/lines — but there were a number of uncovered **branches**. Almost all of them are reachable and just need a few more targeted cases.

Below is a description of the additional Vitest cases I'd like to see (built on the existing `test_helpers` + `expectKeyboardNavigation` harness). I've deliberately left out the full code so we can agree on the scenarios first.

### `keyboard_navigation.test.js`
- `ArrowLeft` from the **1st of the month** should cross into the previous month.
- `ArrowDown` from a date in the **last week** should cross into the next month.
- An **unhandled key** (e.g. `"a"`) on a focused date cell should be a no-op — covers the `if (handler)` guard in `navigateCalendar`.

### `base_component.test.js`
- Clicking a **disabled date** should not set the input value or close the calendar — covers the `aria-disabled` guard in `selectDate`.
- Changing the year input to the **same year** should not reload the calendar — covers the `targetYear !== selectedYear` guard in `changeYear`.
- **Tab-index fallback**: when `minDate` is only visible as an out-of-month cell (e.g. system date `2026-05-15`, `minDate` `2026-05-31`, then navigate to June) and no selected/today date is in view, the 1st of the month should become the tabbable date — covers the final `else if` in `#setTabIndex`.

### `lifecycle_and_edge_cases.test.js`
- Re-running `idempotentConnect` must **not append a second calendar** — covers the `if (this.#calendar) return` guard in `#addCalendarTemplate`.
- `handleCalendarFocus` should be a **no-op when the calendar's parent is not a `<dialog>`**.
- In the dialog scroll test, add a focused element whose rect is **already within bounds** so the `scrollBy` path is skipped — covers the in-view branch.
- `hideCalendar()` should be a **safe no-op after `disconnect()`** (when `#floatingDropdown` is `null`).
- Global keydown should **ignore non-Escape keys** and a `Tab` originating from a **non-boundary element**.

### Result
With these additions:
- `utils.js` — 100% across the board
- `input_controller.js` — **100% branches**
- `calendar_controller.js` — 100% statements/functions/lines, **98.5% branches**

### Remaining uncovered branches (unreachable defensive code)

These 3 branches in `app/javascript/controllers/combobox_datepicker/v1/calendar_controller.js` cannot be exercised through the public API. I'd suggest either removing the dead guard or annotating with `/* v8 ignore next */` so these files can join the 100% coverage ratchet.

1. **L222 — `#fillCalendarWithDates`**
   ```js
   firstCurrentIdx === -1 ? -1 : dates.indexOf(1, firstCurrentIdx + 1);
   ```
   The `=== -1` branch is unreachable. `#fillCalendarWithDates` is private and is always called with a `dates` array that spans a full month, so it always contains the 1st — `dates.indexOf(1)` can never return `-1`.

2. **L557 — `navigateCalendar`**
   ```js
   if (event.key !== "Tab") event.preventDefault();
   ```
   The `key === "Tab"` branch is unreachable. `#getKeyboardHandler` has no `"Tab"` entry, so a `Tab` key never returns a handler and never reaches this line. The guard is vestigial — it could simply be an unconditional `event.preventDefault()`.

3. **L605 — `#escapeCalendar`**
   ```js
   if (this.calendarTarget.contains(event.target)) {
   ```
   The "not contained" branch is unreachable. `#escapeCalendar` is only wired to the `keydown` action on date cells, which are always inside `calendarTarget`, so `contains(event.target)` is always `true`.

ChrisHuynh333:
Thanks for the replies @joshsadam @ericenns 

I've upped test coverage thanks to the comments above, and also removed conditionals that were unreachable.

image

image

For the remaining uncovered lines, I already have tests that enter these branches/conditionals, so I'm not sure what else can be done?

joshsadam:
Hey @ChrisHuynh333 — nice work getting this so close! The good news is we **can** hit a clean 100% (statements, branches, functions, and lines) without touching the production code. There are exactly three uncovered *branches* left, and each just needs one small test. The reason they still show as uncovered is subtle: for an `if`/`else if`, V8 wants to see **both** the "condition true" **and** the "condition false" outcome. Your existing tests already exercise the *true* side of each of these — we're only missing the *false* side.

Here's exactly what to add. Everything uses the existing `test_helpers.js` harness, so there's nothing new to import beyond helpers you already use.

### 1. `calendar_controller.js` — line 416 (the `#setTabIndex` fallback)

```js
} else if (verifyDateIsInMonth(minDate)) {
  minDate.tabIndex = 0;
  return;
}
```

**Why it's uncovered:** Your tests always reach this line with a `minDate` that *is* in the visible month (so `verifyDateIsInMonth(minDate)` returns `true`). The `false` path — where the min-date cell is only showing as an out-of-month "spillover" cell, so we fall all the way through to `getFirstOfMonthNode(...)` — is never hit.

**How to cover it:** Pick a system date + `minDate` where, after navigating a month forward, the min date appears only as an out-of-month cell and neither today nor a selected date is visible. `2026-05-15` with a `minDate` of `2026-05-31` works, because in the June grid `2026-05-31` renders as an `outOfMonth` cell (see `getJune2026Dates()`), and today (May 15) isn't on the June calendar at all. Add this to the `describe("datepicker with min date", ...)` block in `base_component.test.js`:

```js
it("tab index falls back to the 1st of the month when minDate is out of the displayed month", async () => {
  vi.setSystemTime(new Date("2026-05-15T09:00:00-05:00"));
  renderBaseFixture();
  renderMinDate("2026-05-31");
  application = await startController();

  openCalendarByInputArrow();
  await vi.runOnlyPendingTimersAsync();

  // On May, the min date (May 31) is in-month and is the tabbable date.
  expect(getDateNode("2026-05-31").tabIndex).toBe(0);

  // Navigate to June, where May 31 becomes an out-of-month spillover cell and
  // neither today (May 15) nor a selected date is visible.
  getForwardButton().click();
  await vi.runOnlyPendingTimersAsync();

  expect(getMonthSelect().value).toBe("June");
  // #setTabIndex falls through to the 1st of the month.
  expect(getDateNode("2026-06-01").tabIndex).toBe(0);
  expect(getDateNode("2026-05-31").tabIndex).toBe(-1);
});
```

### 2. `input_controller.js` — line 238 (`hideCalendar` guard)

```js
hideCalendar() {
  try {
    if (this.#floatingDropdown) {   // <-- false branch never taken
      this.#floatingDropdown.hide();
    }
  } catch (error) {
    this.#handleError(error, "hideDropdown");
  }
}
```

**Why it's uncovered:** Every test calls `hideCalendar()` while the dropdown exists, so `this.#floatingDropdown` is always truthy. The guard's `false` path only happens after `disconnect()`, which sets `this.#floatingDropdown = null`.

**How to cover it:** Call `disconnect()` first, then `hideCalendar()` and assert it's a safe no-op. Add this to the `describe("error and catch handling", ...)` block in `lifecycle_and_edge_cases.test.js`:

```js
it("hideCalendar is a safe no-op after disconnect", async () => {
  renderBaseFixture();
  application = await startController();
  await vi.runOnlyPendingTimersAsync();

  const controller = inputControllerInstance(application);
  controller.disconnect();

  // #floatingDropdown is now null; hideCalendar should not throw.
  expect(() => controller.hideCalendar()).not.toThrow();
});
```

### 3. `input_controller.js` — line 281 (`handleGlobalKeydown` Tab handling)

```js
if (event.key === "Tab") {
  if (
    event.target === ...getLastFocusableElement() &&
    !event.shiftKey
  ) {
    ...
  } else if (event.target === this.datepickerInputTarget) {   // <-- false branch never taken
    this.hideCalendar();
  }
  return;
}
```

**Why it's uncovered:** Your tests cover a Tab from the last focusable element (first `if`) and a Tab from the input itself (`else if` = true). What's missing is a Tab from an element that is **neither** the last focusable element **nor** the input — i.e., a "middle" element — which should make the `else if` evaluate to `false` and do nothing.

**How to cover it:** The forward button is a good "non-boundary" element (`getFirstFocusableElement()` returns the back button and `getLastFocusableElement()` returns the clear button, so the forward button is neither, and it's not the input). Dispatch a non-shift Tab from it and assert the calendar stays open. Add this next to the existing `"contained tab logic on calendar"` test in `base_component.test.js`:

```js
it("global keydown ignores Tab from a non-boundary element", async () => {
  vi.setSystemTime(new Date("2026-05-07T09:00:00-05:00"));
  renderBaseFixture();
  application = await startController();
  const calendar = document.getElementById("test_id-calendar");

  openCalendarByInputArrow();
  expect(calendar.hidden).toBe(false);

  // The forward button is neither the first, last, nor the input element, so a
  // (non-shift) Tab from it should be a no-op and leave the calendar open.
  getForwardButton().dispatchEvent(
    new KeyboardEvent("keydown", { key: "Tab", bubbles: true }),
  );

  expect(calendar.hidden).toBe(false);
});
```

### Result

With these three cases added, `pnpm run test:js:coverage` gives us:

| File | % Stmts | % Branch | % Funcs | % Lines |
| --- | --- | --- | --- | --- |
| `calendar_controller.js` | 100 | 100 | 100 | 100 |
| `input_controller.js` | 100 | 100 | 100 | 100 |

(All 427 JS tests still pass.) No production-code changes are needed — these were all reachable branches, just missing their "false" case. Once these are in, both controllers are eligible to join the 100% coverage ratchet.

ChrisHuynh333:
> Hey @ChrisHuynh333 — nice work getting this so close! The good news is we **can** hit a clean 100% (statements, branches, functions, and lines) without touching the production code. There are exactly three uncovered _branches_ left, and each just needs one small test. The reason they still show as uncovered is subtle: for an `if`/`else if`, V8 wants to see **both** the "condition true" **and** the "condition false" outcome. Your existing tests already exercise the _true_ side of each of these — we're only missing the _false_ side.
> 
> Here's exactly what to add. Everything uses the existing `test_helpers.js` harness, so there's nothing new to import beyond helpers you already use.
> 
> ### 1. `calendar_controller.js` — line 416 (the `#setTabIndex` fallback)
> ```js
> } else if (verifyDateIsInMonth(minDate)) {
>   minDate.tabIndex = 0;
>   return;
> }
> ```
> 
> **Why it's uncovered:** Your tests always reach this line with a `minDate` that _is_ in the visible month (so `verifyDateIsInMonth(minDate)` returns `true`). The `false` path — where the min-date cell is only showing as an out-of-month "spillover" cell, so we fall all the way through to `getFirstOfMonthNode(...)` — is never hit.
> 
> **How to cover it:** Pick a system date + `minDate` where, after navigating a month forward, the min date appears only as an out-of-month cell and neither today nor a selected date is visible. `2026-05-15` with a `minDate` of `2026-05-31` works, because in the June grid `2026-05-31` renders as an `outOfMonth` cell (see `getJune2026Dates()`), and today (May 15) isn't on the June calendar at all. Add this to the `describe("datepicker with min date", ...)` block in `base_component.test.js`:
> 
> ```js
> it("tab index falls back to the 1st of the month when minDate is out of the displayed month", async () => {
>   vi.setSystemTime(new Date("2026-05-15T09:00:00-05:00"));
>   renderBaseFixture();
>   renderMinDate("2026-05-31");
>   application = await startController();
> 
>   openCalendarByInputArrow();
>   await vi.runOnlyPendingTimersAsync();
> 
>   // On May, the min date (May 31) is in-month and is the tabbable date.
>   expect(getDateNode("2026-05-31").tabIndex).toBe(0);
> 
>   // Navigate to June, where May 31 becomes an out-of-month spillover cell and
>   // neither today (May 15) nor a selected date is visible.
>   getForwardButton().click();
>   await vi.runOnlyPendingTimersAsync();
> 
>   expect(getMonthSelect().value).toBe("June");
>   // #setTabIndex falls through to the 1st of the month.
>   expect(getDateNode("2026-06-01").tabIndex).toBe(0);
>   expect(getDateNode("2026-05-31").tabIndex).toBe(-1);
> });
> ```
> 
> ### 2. `input_controller.js` — line 238 (`hideCalendar` guard)
> ```js
> hideCalendar() {
>   try {
>     if (this.#floatingDropdown) {   // <-- false branch never taken
>       this.#floatingDropdown.hide();
>     }
>   } catch (error) {
>     this.#handleError(error, "hideDropdown");
>   }
> }
> ```
> 
> **Why it's uncovered:** Every test calls `hideCalendar()` while the dropdown exists, so `this.#floatingDropdown` is always truthy. The guard's `false` path only happens after `disconnect()`, which sets `this.#floatingDropdown = null`.
> 
> **How to cover it:** Call `disconnect()` first, then `hideCalendar()` and assert it's a safe no-op. Add this to the `describe("error and catch handling", ...)` block in `lifecycle_and_edge_cases.test.js`:
> 
> ```js
> it("hideCalendar is a safe no-op after disconnect", async () => {
>   renderBaseFixture();
>   application = await startController();
>   await vi.runOnlyPendingTimersAsync();
> 
>   const controller = inputControllerInstance(application);
>   controller.disconnect();
> 
>   // #floatingDropdown is now null; hideCalendar should not throw.
>   expect(() => controller.hideCalendar()).not.toThrow();
> });
> ```
> 
> ### 3. `input_controller.js` — line 281 (`handleGlobalKeydown` Tab handling)
> ```js
> if (event.key === "Tab") {
>   if (
>     event.target === ...getLastFocusableElement() &&
>     !event.shiftKey
>   ) {
>     ...
>   } else if (event.target === this.datepickerInputTarget) {   // <-- false branch never taken
>     this.hideCalendar();
>   }
>   return;
> }
> ```
> 
> **Why it's uncovered:** Your tests cover a Tab from the last focusable element (first `if`) and a Tab from the input itself (`else if` = true). What's missing is a Tab from an element that is **neither** the last focusable element **nor** the input — i.e., a "middle" element — which should make the `else if` evaluate to `false` and do nothing.
> 
> **How to cover it:** The forward button is a good "non-boundary" element (`getFirstFocusableElement()` returns the back button and `getLastFocusableElement()` returns the clear button, so the forward button is neither, and it's not the input). Dispatch a non-shift Tab from it and assert the calendar stays open. Add this next to the existing `"contained tab logic on calendar"` test in `base_component.test.js`:
> 
> ```js
> it("global keydown ignores Tab from a non-boundary element", async () => {
>   vi.setSystemTime(new Date("2026-05-07T09:00:00-05:00"));
>   renderBaseFixture();
>   application = await startController();
>   const calendar = document.getElementById("test_id-calendar");
> 
>   openCalendarByInputArrow();
>   expect(calendar.hidden).toBe(false);
> 
>   // The forward button is neither the first, last, nor the input element, so a
>   // (non-shift) Tab from it should be a no-op and leave the calendar open.
>   getForwardButton().dispatchEvent(
>     new KeyboardEvent("keydown", { key: "Tab", bubbles: true }),
>   );
> 
>   expect(calendar.hidden).toBe(false);
> });
> ```
> 
> ### Result
> With these three cases added, `pnpm run test:js:coverage` gives us:
> 
> File	% Stmts	% Branch	% Funcs	% Lines
> `calendar_controller.js`	100	100	100	100
> `input_controller.js`	100	100	100	100
> (All 427 JS tests still pass.) No production-code changes are needed — these were all reachable branches, just missing their "false" case. Once these are in, both controllers are eligible to join the 100% coverage ratchet.

Thanks for the clarification and need to test the false branches. I've included the code above into the respective test files and everything is 100%.

b028d75de264f35b8b615f6699b119ac46f74c15

ericenns (approved):
Looks good to me!

PR #2229, opened 2026-09-02
chore: Refactor transfer samples tests

JeffreyThiessen:

This PR moves controller and system tests for sample transfers to integration tests.

image

1. Ensure that services are running (e.g. `devenv up`)
2. Run all the group_links integration tests
```bash
bin/rails test test/integration/projects/samples/transfer_test.rb test/integration/groups/samples/transfer_test.rb
```
3. In `coverage` folder launch a python server to serve coverage report (e.g. `python -mhttp.server`)
4. Open localhost:8000 in your browser
5. type `transfers` in search box and confirm percentages match the image above

## PR acceptance checklist


test/integration/projects/samples/transfer_test.rb
```diff
@@ -0,0 +1,437 @@
...
+                          -> { @project2.samples.count } => 1 do
+          perform_enqueued_jobs only: [::Samples::TransferJob]
+        end
+
+        assert_not @project.samples.exists?(@sample1.id)
+        assert_samples_page(@project2, 21)
+        assert_select 'tbody#samples-table-body input[name="sample_ids[]"][checked]', count: 0
+      end
+
+      test 'transfer samples v2' do
```
joshsadam: **[major] The V1 and V2 tests are almost identical copies.** Each of these tests (`should enqueue ...` for group and project, `should not transfer ... session storage cleared`, `... same name in destination project`, and `updating sample selection`) has a `v2` twin. The only difference is turning on `Flipper.enable(:v2_sample_transfer)` and using the V2 job class — every assertion is the same.

You could write each test once and loop over both jobs:

```ruby
[[::Samples::TransferJob, false], [::Samples::TransferJobV2, true]].each do |job_class, v2|
  # enable the flag when v2 is true, and disable it again afterward
end
```

The helpers `assert_transfer_enqueued(job_class, ...)` and `assert_no_transfer_progress_or_selection(job_class)` already take the job class, so this is an easy change. You'd have less duplicated code, and V1 and V2 couldn't accidentally drift apart. (The `dialog close button hidden` tests are a bit different — see that comment.)

test/integration/projects/samples/transfer_test.rb
```diff
@@ -0,0 +1,437 @@
...
+        end
+
+        assert_empty @project.samples.where(id: [@sample1.id, @sample2.id, @sample30.id])
+        assert_samples_page(@project2, 23)
+        [@sample1, @sample2, @sample30].each do |sample|
+          assert_select "tbody#samples-table-body tr##{dom_id(sample)}", count: 1
+        end
+      end
+
+      test 'dialog close button hidden during transfer samples' do
```
joshsadam: **[major] There are 4 near-copies of the `dialog close button hidden` test** — one plain, one with `v2_select2`, one with `v2_sample_transfer`, and one with both flags on. They're mostly the same code with different flags turned on.

One thing to watch out for: they aren't *only* different by flags. The two `v2_sample_transfer` tests skip the `[role="progressbar"]` check that the other two keep. If that skip is intentional, you could combine all four into one helper that takes which flags to enable and whether to expect the progressbar. Then adding a flag is a one-line change instead of a brand new ~15-line test.

test/integration/projects/samples/transfer_test.rb
```diff
@@ -0,0 +1,437 @@
...
+        assert_empty @project.samples.where(id: [@sample1.id, @sample2.id, @sample30.id])
+        assert_samples_page(@project2, 23)
+        [@sample1, @sample2, @sample30].each do |sample|
+          assert_select "tbody#samples-table-body tr##{dom_id(sample)}", count: 1
+        end
+      ensure
+        Flipper.disable(:v2_sample_transfer)
+      end
+
+      test 'dialog close button hidden during transfer samples v2' do
```
joshsadam: **[minor] The V2 tests quietly skip a check.** The non-V2 close-button tests check that `[role="progressbar"]` shows up, but the `v2` versions leave that check out. If the progressbar really behaves differently when V2 is on, add a check for what V2 *should* show instead. If it was dropped by accident, add it back. Right now it's hard to tell which one it is.

test/integration/projects/samples/transfer_test.rb
```diff
@@ -0,0 +1,437 @@
...
+            sample_ids.each do |sample_id|
+              sample = Sample.find(sample_id)
+              assert_select 'p', text: /#{Regexp.escape(sample.name)}/
+              assert_select 'p', text: /#{Regexp.escape(sample.puid)}/
+            end
+          end
+        end
+      end
+
+      def assert_no_transfer_progress_or_selection(job_class) # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
```
joshsadam: **[minor]** `assert_no_transfer_progress_or_selection` is the only helper that needs `# rubocop:disable Metrics/AbcSize,Metrics/MethodLength`. That's usually a sign it's doing too much — here it both waits for the completion broadcast and checks what's inside it. If you split those into two smaller methods, you can drop the rubocop disable and it'll read more clearly.

test/integration/projects/samples/transfer_test.rb
```diff
@@ -0,0 +1,437 @@
...
+      end
+
+      test 'should get new for group if owner' do
+        get new_samples_transfer_path(namespace_id: @namespace.id, format: :turbo_stream)
+        assert_response :success
+      end
+
+      test 'should get new for group if maintainer' do
+        user = users(:joan_doe)
+        login_as user
```
joshsadam: **[minor] Two different login helpers are used.** `setup` logs in with `sign_in @user`, but some tests switch to `login_as user`. This is the only file in `test/integration` that uses `login_as` — the other 11 integration files all use `sign_in`. Changing these to `sign_in` would keep it consistent with the rest of the tests.

test/integration/projects/samples/transfer_test.rb
JeffreyThiessen: Great idea, updated in https://github.com/phac-nml/irida-next/pull/2229/commits/3f3dc8e4b5cb66d0c87b30316d230fd803ffcd88

test/integration/projects/samples/transfer_test.rb
JeffreyThiessen: The progressbar assert being missing was a mistake, good catch. Fixed with the refactors in https://github.com/phac-nml/irida-next/pull/2229/commits/3f3dc8e4b5cb66d0c87b30316d230fd803ffcd88

test/integration/projects/samples/transfer_test.rb
JeffreyThiessen: The progressbar assert being missing was a mistake, good catch. Fixed with the refactors in https://github.com/phac-nml/irida-next/commit/3f3dc8e4b5cb66d0c87b30316d230fd803ffcd88

test/integration/projects/samples/transfer_test.rb
JeffreyThiessen: refactored

test/integration/projects/samples/transfer_test.rb
JeffreyThiessen: refactored in https://github.com/phac-nml/irida-next/pull/2229/commits/3f3dc8e4b5cb66d0c87b30316d230fd803ffcd88

JeffreyThiessen:
Tests have been refactored for the groups side now too, this is ready for review again.

joshsadam (dismissed):

test/integration/groups/samples/transfer_test.rb
```diff
@@ -0,0 +1,297 @@
...
+          assert_select "tbody#samples-table-body tr##{dom_id(@sample29)}", count: 1
+          assert_project_samples_page(@project4, 26, sort: 'puid asc')
+          assert_select "tbody#samples-table-body tr##{dom_id(@sample1)}", count: 1
+          assert_select "tbody#samples-table-body tr##{dom_id(@sample2)}", count: 1
+        ensure
+          Flipper.disable(:v2_sample_transfer) if v2
+        end
+      end
+
+      test 'sample transfer project listing should be empty for maintainer if no other projects in hierarchy' do
```
ericenns, line 177: **nitpick (blocking):** I think your missing an assert here, specifically one that checks that the select2 input has no selectable options.

Like what you have in `test 'sample transfer project listing should be empty for maintainer if no other projects in hierarchy'`from the projects test

ericenns (changes requested):
Looking really good, just one test needs an update to add a missing assert

test/integration/groups/samples/transfer_test.rb
JeffreyThiessen: Good catch, looks like i used the wrong assert for that one while copy pasting the asserts. https://github.com/phac-nml/irida-next/pull/2229/commits/0756ff20e74e216f1d8f927362be1ec7416078f5

ericenns (approved):
Looks good to me!

PR #2269, opened 2026-09-11
Harden JavaScript test isolation and enforce coverage in CI

joshsadam:
## What does this PR do and why?

Make the JavaScript test setup reliable before expanding coverage for #2233. Controller suites now remove their fixtures and await Stimulus disconnect callbacks before stopping the application. Unexpected Stimulus lifecycle/action errors fail the test, and final browser-state cleanup still runs after failed teardown.

Restore browser mocks between tests, clear both storage areas, and fix two existing order dependencies in samplesheet and workflow-selection tests. CI now runs coverage and enforces the existing per-file thresholds; empty test discovery fails. Add helper regression tests and a short JavaScript testing guide.

The localStorage fallback matches the setup change in #2265. This branch starts from main and does not require the other coverage PRs. No production JavaScript changes.

## Screenshots or screen recordings

Not applicable: test infrastructure and documentation only.

## How to set up and validate locally

1. Install the locked dependencies with `pnpm install --frozen-lockfile`.
2. Run `pnpm run test:js:coverage` — 309 tests across 19 files pass; all configured coverage thresholds pass.
3. Run `pnpm exec vitest run --sequence.shuffle --sequence.seed=2233` — the same 309 tests pass in shuffled order.
4. Read `docs/testing/javascript.md` for controller lifecycle, mock, coverage and browser-test guidance.

Changed JavaScript files pass ESLint and Prettier. An additional temporary failure probe confirmed that a captured controller error fails teardown while the following test receives real timers, restored globals and empty storage. Overall application coverage remains about 31%; this PR enforces the existing per-file targets rather than claiming global 100% coverage.

## PR acceptance checklist

- [x] I have evaluated the [PR acceptance checklist](https://phac-nml.github.io/irida-next/docs/development/development_processes/code_review#acceptance-checklist) for this PR.

Issue #2233, opened 2026-09-03
test(js): drive all JavaScript files to 100% coverage


joshsadam:
## Goal

Drive per-file JavaScript coverage (statements / branches / functions / lines) to **100%** for
every `app/javascript/**/*.js` file, enforced via the existing Vitest "ratchet" allowlist in
`vitest.config.js`, then flip to a single global 100% threshold and drop the per-file allowlist.

This is a tracking issue for the incremental effort. Work ships as many small, subsystem-grouped
PRs; each PR brings a file to 100% and adds it to the `test.coverage.thresholds` allowlist.

## Conventions

- **Per-file recipe**
  1. Find gaps: `npx vitest run --coverage --coverage.include='<src>' --coverage.reporter=text <test>`
  2. Add/extend a mirrored test under `test/javascript/...` (utilities → assert pure functions;
     controllers → DOM fixture + `Application.start()` Stimulus harness, see
     `selection_controller.test.js` / `treegrid_controller.test.js` / `combobox/v1_controller.test.js`).
  3. Annotate genuinely-unreachable defensive guards with `/* v8 ignore next */`.
  4. Add the file to `test.coverage.thresholds` in `vitest.config.js` (all four = 100).
  5. `pnpm run test:js:coverage` green; lint the changed test.
- **Excluded from measurement** (bootstrapping / worker entry modules, run on import or in a worker
  context, not meaningfully unit-testable) — configured via `coverage.exclude`:
  `application.js`, `active_admin_navigation.js`, `controllers/index.js`,
  `controllers/application.js`, `controllers/combobox_datepicker/constants.js`, `workers/**`.
- **PR #2228 coupling**: PR #2228 extends `selection_controller.test.js`. Do `selection_controller.js`
  and `table_selection_controller.js` **last** to avoid merge conflicts.

## Current status

7 files at 100% and locked into the ratchet:
`two_lists_selection_controller.js` (v1), `workflow_selection_controller.js`, `treegrid_controller.js`,
`utilities/collection.js`, `utilities/live_region.js`, `utilities/message_formatter.js`,
`utilities/word_connector.js`.

Partially covered (tests started, not yet at 100%): `utilities/form.js` (94.64/90/100/94.54 — buggy
legacy Rack-derived `normalizeParams` has unreachable compound-condition branches; needs
`/* v8 ignore */` on dead operands), `controllers/combobox/utils.js` (93/78),
[...]

PR #2272, opened 2026-09-11
tests(js): Cover clipboard controls and fix feedback lifecycle and token masking

joshsadam:

Add behavior coverage for clipboard, copy and token controls, and fix the failures exposed by those tests:

- Cancel feedback timers on disconnect and restore reusable elements. Repeated copies restart the feedback period.
- Ignore clipboard writes that complete after an earlier connection has ended.
- Copy the button value when a descendant is clicked, and destroy tooltip resources when their target is removed.
- Show token-copy success only after the write completes, handle rejected or unavailable clipboard writes, preserve the full masking length, and remask revealed tokens on reconnect.

All three controllers now enforce 100% statements, branches, functions and lines. Regression tests were run against the original implementation before fixes. Part of #2233; based directly on main after #2269.

No visual redesign. These changes correct transient feedback and token state during repeated actions and reconnects. Validation is automated DOM testing with mocked clipboard/tooltip boundaries; no browser screenshots were captured.

1. Run `pnpm install --frozen-lockfile`.
2. Run `pnpm run test:js:coverage` — 335 tests pass with shuffled order, including 26 new tests. All three selected controllers meet every 100% coverage threshold.
3. Inspect the mirrored clipboard, copy and token controller tests for rejection, timer and disconnect/reconnect cases.

Changed JavaScript passes ESLint and Prettier. Independent code review found no blockers. No Rails or real-browser tests were run for this branch.

## PR acceptance checklist


deepsidhu85 (approved):
Looks good to me!

PR #2290, opened 2026-09-15
chore: Remove @ variables from locals definition in partials

ericenns:

This PR updates all partials (except samples destroy as that is handled in a separate PR) to remove default @ variables within `<%# locals: (variable: @default_variable)` in all partials, as the intention with locals definitions is that they are explicitly passed to the partials. With this PR all locals are now explicitly passed and do not rely on @ variables.

1. Verify that tests pass and all partials that used @ variables as defaults have them explicitly passed.

## PR acceptance checklist


app/views/groups/members/_create_modal.html.erb
```diff
@@ -8,7 +8,14 @@
...
   <% if @new_member&.invalid? %>
-    <%= render partial: "form" %>
+    <%= render partial: "form",
+      locals: {
+        access_levels: access_levels,
+        available_users: available_users,
+        namespace: namespace,
+        new_member: new_member,
+        tab: tab,
+      } %>
```
joshsadam: **Blocking:** This wrapper now has two sources of truth for the same data.

`_form` is strict locals and needs `access_levels`, `available_users`, `namespace`, `new_member`, and `tab`. This parent still decides with `@new_member` (a controller instance variable) and then passes a *different* `new_member` local into the form:

```erb
<% if @new_member&.invalid? %>   <%# instance variable from the controller %>
  <%= render partial: "form",
    locals: {
      new_member: new_member,    <%# local passed by the caller, or NameError %>
      tab: tab,
      ...
    } %>
<% else %>
  ... new_group_member_path(tab: @tab)  <%# instance variable again %>
```

That only works on index because `@new_member` is nil there, so we take the `else` branch and never look up `access_levels` / `available_users`. Index still renders this as:

```erb
<%= render partial: "create_modal", locals: { open: false, tab: @tab } %>
```

If `@new_member` is ever invalid on that path, this template will raise `NameError` for locals that were never passed.

Could we give this partial one contract?

```erb
<%# locals: (open:, access_levels:, available_users:, namespace:, new_member:, tab:) -%>
<% if new_member&.invalid? %>
  ...
<% else %>
  ... new_group_member_path(tab: tab)
```

Then pass that same hash from every caller (index HTML, index turbo stream, create turbo stream, and the group-links index stream). On the lazy-load paths `new_member` and `available_users` can be `nil`; we only need real values when showing the invalid form.

The projects sibling (`app/views/projects/members/_create_modal.html.erb`) has the same split.

app/views/projects/members/_create_modal.html.erb
```diff
@@ -8,7 +8,14 @@
...
   <% if @new_member&.invalid? %>
-    <%= render partial: "form" %>
+    <%= render partial: "form",
+      locals: {
+        access_levels: access_levels,
+        available_users: available_users,
+        namespace: namespace,
+        new_member: new_member,
+        tab: tab,
+      } %>
```
joshsadam: **Blocking:** Same split-brain as the groups create modal.

The condition still uses `@new_member`, the form now requires locals, and the lazy-load URL still uses instance variables:

```erb
<% if @new_member&.invalid? %>
  <%= render partial: "form", locals: { namespace: namespace, tab: tab, ... } %>
<% else %>
  <%= turbo_frame_tag "new_member_dialog_content",
        src: new_namespace_project_member_path(@namespace.parent, @namespace.project, tab: @tab) %>
```

`namespace` / `tab` are already being passed on the create stream, so the `@namespace` / `@tab` reads are a second channel. Index still only passes `{ open: false, tab: @tab }`.

Could we add a `<%# locals: %>` comment and use only those locals (including `namespace.parent` / `namespace.project` / `tab` in the turbo-frame URL), with every caller sending the same hash?

app/views/projects/_transfer.html.erb
```diff
@@ -32,7 +32,10 @@
         </div>
       </div>
 
-      <%= render "projects/transfer_form", project: @project %>
+      <%= render "projects/transfer_form",
+        authorized_namespaces: authorized_namespaces,
+        project: project,
+        transfer_form: transfer_form %>
```
joshsadam: **Blocking:** This parent was only half-converted.

The nested form is now strict, so we correctly pass locals here:

```erb
<%= render "projects/transfer_form",
  authorized_namespaces: authorized_namespaces,
  project: project,
  transfer_form: transfer_form %>
```

But this wrapper has **no** `<%# locals: %>` comment, and the warning copy a few lines above still reads the controller ivar:

```erb
project_name: @project.human_name
project: @project.name
```

The groups sibling in this PR shows the finished version:

```erb
<%# locals: (group:, authorized_namespaces:, transfer_form:) -%>
...
group_name: group.human_name
```

Why this is easy to break later: adding `<%# locals: (project:, ...) %>` to this file will hide `@project`. The warning text would then be `nil` or raise, while the form below still works because it uses the `project` local.

Could we give `_transfer.html.erb` the same locals comment as groups and replace `@project` with `project` in this template?

app/views/groups/group_links/_invite_group_modal.html.erb
```diff
@@ -8,7 +10,7 @@
   <% dialog.with_header(title: I18n.t("groups.group_links.new.title")) %>
 
   <% if defined?(new_group_link) && new_group_link&.invalid? %>
-    <%= render partial: "form", locals: { new_group_link: new_group_link } %>
+    <%= render partial: "form", locals: { new_group_link: new_group_link, access_levels: access_levels, namespace: namespace, namespace_linkable_groups: namespace_linkable_groups, tab: tab } %>
```
joshsadam: **Suggestion:** `defined?(new_group_link)` is leftover from before this partial declared locals.

Line 1 now requires `new_group_link:`:

```erb
<%# locals: (open:, new_group_link:, access_levels:, namespace:, namespace_linkable_groups:, tab:) -%>
```

With strict locals, that variable is always defined (it may still be `nil`). `defined?(new_group_link)` will always be true, so it no longer guards a missing local. Callers that forget it will get a Rails missing-local error instead of taking the `else` branch.

Could we collapse this to the actual condition?

```erb
<% if new_group_link&.invalid? %>
```

app/views/projects/group_links/_invite_group_modal.html.erb
```diff
@@ -8,7 +8,14 @@
...
   <% if defined?(new_group_link) && new_group_link&.invalid? %>
-    <%= render partial: "form", locals: { new_group_link: new_group_link } %>
+    <%= render partial: "form",
+      locals: {
+        access_levels: access_levels,
+        namespace: namespace,
+        namespace_linkable_groups: namespace_linkable_groups,
+        new_group_link: new_group_link,
+        tab: tab,
+      } %>
```
joshsadam: **Blocking:** Groups got a locals contract for this wrapper; projects did not.

We now forward `namespace` into `_form`, but this file still has no `<%# locals: %>` comment, still uses `defined?(new_group_link)`, and the lazy-load URL still reads `@namespace`:

```erb
<% if defined?(new_group_link) && new_group_link&.invalid? %>
  <%= render partial: "form", locals: { namespace: namespace, ... } %>
<% else %>
  <%= turbo_frame_tag "new_group_link_dialog_content",
        src: new_namespace_project_group_link_path(@namespace.parent, @namespace.project, tab: tab) %>
```

So `tab` comes from a local, while `namespace` comes from an instance variable on the other branch. The groups version of this modal already uses only locals for the turbo-frame path.

Could we add the same `<%# locals: %>` comment as groups, switch the `if` to `new_group_link&.invalid?`, and use `namespace.parent` / `namespace.project` in the `src`?

app/views/projects/samples/metadata/_new_metadata_modal.html.erb
```diff
@@ -1,4 +1,4 @@
 <%= render Viral::DialogComponent.new(open: open) do |dialog| %>
   <% dialog.with_header(title: t(".title")) %>
-  <%= render partial: "projects/samples/metadata/form" %>
+  <%= render partial: "projects/samples/metadata/form", locals: { sample: @sample } %>
```
joshsadam: **Blocking:** `_form` is now strict (`<%# locals: (sample:) -%>`), but this wrapper still pulls `sample` from the controller:

```erb
<%= render partial: "projects/samples/metadata/form", locals: { sample: @sample } %>
```

`MetadataController#new` still only passes `open: true`. That works only because this modal has **no** locals comment, so `@sample` is still visible here.

The attachment modal in this PR is the complete version of the same pattern. It forwards a local, and the controller passes it:

```erb
<%# _new_attachment_modal.html.erb %>
<%= render partial: "projects/samples/attachments/form",
  locals: { attachment: attachment, sample: sample } %>
```

```ruby
# AttachmentsController#new
locals: { open: true, attachment: ..., sample: @sample }
```

Could we do the same here: give this modal a locals comment, take `sample:` as a local, and pass `sample: @sample` from `MetadataController#new` and `samples/metadata/bulk_create.turbo_stream.erb`?

app/views/profiles/passwords/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%# locals: (minimum_password_length: @minimum_password_length, user: @user) -%>
+<%# locals: (minimum_password_length:, user: @user) -%>
```
joshsadam: Could we change `user: @user` to `user:` here? `edit.html.erb` already passes it explicitly, and this is one of the instance-variable defaults this PR is removing.

app/views/groups/bots/_bot_modal.html.erb
```diff
@@ -1,4 +1,4 @@
-<%# locals: (open:, access_levels: @access_levels, bot_account: @bot_account) -%>
+<%# locals: (open:, access_levels: , bot_account: ) -%>
```
joshsadam: **Nit:** This looks like the old defaults were deleted in place:

```erb
<%# locals: (open:, access_levels: , bot_account: ) -%>
```

The projects sibling is the intended spelling:

```erb
<%# locals: (open:, access_levels:, bot_account:) -%>
```

In Ruby both are required keyword args, so this is not a nil default. Matching the projects signature just makes that obvious.

joshsadam (changes requested):

app/views/profiles/passwords/_form.html.erb
ericenns: Removed in https://github.com/phac-nml/irida-next/pull/2290/commits/a2fbde30c31114deed534c76c06a1703a43b5516

joshsadam (approved):

